PR #34, opened 2018-02-08
removing lantern dependency

timkpaine:
- usable from any comm-connection assuming it implements the correct connection string/api
- comm://\<jupyter session id\>/\<applicationname\>/\<channel id\>


texodus:
looks good!

PR #41, opened 2018-02-12
Support ns and us time resolutions

nmichaud:


texodus:
Looks good!  Could use a test though.

PR #52, opened 2018-02-20
change mime renderer type to not interfere with saving notebook

timkpaine:
- needs to be `application/psp+json`
- making attributes start with _


texodus:
Looks good!

PR #99, opened 2018-04-26
Upgrade to Hypergrid 2.1.8

joneit:


joneit:
@nmichaud I'd like to go over the setPSP method with you to better understand it purpose. Can any of this be moved to the data model?

PR #118, opened 2018-05-18
Adding template polyfill for browsers that do not support the template tag

neilslinger:
Adding template polyfill for browsers that do not support the template tag


texodus:
Thanks for the fix!

PR #121, opened 2018-05-24
Adding dark mode via theme

timkpaine:
@texodus only issue is this doesn't theme hypergrid


timkpaine:
tmp

texodus:
Looks good, thanks!  Some basic Hypergrid theming will be in 0.1.14.

timkpaine:
line highlight is still in light mode, makes it so you can't see the date

PR #124, opened 2018-05-29
Ignore boost in perspective cpp

LukeSheard:
Small DX improvement if you have to copy boost headers into the project.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -77,6 +77,7 @@ flycheck_*.el
 *.out
 *.app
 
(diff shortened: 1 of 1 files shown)


texodus:
Thanks!  Looks good.  Worth noting:  there is now a `PSP_DOCKER` environment variable for portable Emscripten, and should generally be the preferred way to build Perspective unless you are experimenting with compiler versions or need ultimate compile speed.

PR #135, opened 2018-06-12
Support for labeled drop boxes in perspective viewer

msturdikova:
Added support for visible labels when columns are dropped into viewer configuration widgets, such as row_pivots, column_pivots, sort & filters.


texodus:
I've made some additions to this and re-submitted in #137 - closing

PR #149, opened 2018-07-09
Add to_csv to view object

segarman:
WIP, will write tests once I can get a local build of perspective


packages/perspective/src/js/perspective.js
```diff
@@ -541,12 +518,6 @@ view.prototype.to_json = async function(options) {
 
     let data;
 
-    if (options.format && options.format === "table") {
```
segarman: removed this since it will break anyways, prev line 558 was always calling data.push which will fail if data was an object

texodus:
I've rebased and merged these in #150 - thanks!

PR #156, opened 2018-07-14
Split Build Steps into individual tasks

LukeSheard:
This splits out all the long commands into steps with `npm-run-all`. Also adds a step which clears the build directory before a build run.


texodus:
Looks great, Thanks!

PR #157, opened 2018-07-16
Add typing for perspective

segarman:


packages/perspective/index.d.ts
```diff
@@ -0,0 +1,149 @@
...
+    export enum TypeNames {
+        STRING = 'string',
+        FLOAT = 'float',
+        INTEGER = 'integer',
+        BOOLEAN = 'boolean',
+        DATE = 'date'
+    }
+
+    export type ValuesByType = {
+        [ key in TypeNames ]: Array<string>
```
LukeSheard: For primitives it's better to use string[] for arrays.

packages/perspective/index.d.ts
```diff
@@ -0,0 +1,149 @@
...
+
+    export enum SortOrders {
+        ASC = 'asc',
+        ASC_ABS = 'asc abs',
+        DESC = 'desc',
+        DESC_ABS = 'desc abs',
+        NONE = 'none',
+    }
+
+    enum NUMBER_AGGREGATES {
```
LukeSheard: These might potentially be better as Enums rather than strings?

packages/perspective/index.d.ts
segarman: https://palantir.github.io/tslint/rules/array-type/
we use "generic" on our project, hence Array<T> over T[]
but happy to change if you want me to

packages/perspective/index.d.ts
segarman: they are enums, or you mean to create a global enum with all of them and NUMBER, STRING, BOOLEAN to be just a subset of them?

packages/perspective/index.d.ts
LukeSheard: Sure, since there's no tslint setup of similar and they're equivalent let's just go with this for now. We can always change later.

packages/perspective/index.d.ts
LukeSheard: I was thinking that maybe we can take the aggregate keywords and Enum those. But I couldn't find a nice way to make them all align up. Since these are private I'll take another look later and we can hoist them into private enums later.

segarman: thanks

timkpaine:
given that the jupyterlab sub package is already in typescript, typing should also be added there (instead of the current <any> casts).

timkpaine:
shouldn't the jupyterlab package use this

PR #161, opened 2018-07-18
Forgo publishing cpp source code

LukeSheard:
- Skip publishing the cpp source code for perspective.
- Slightly improve the logging and build time for perspective core module


texodus:
Looks good, thanks!

Issue #160, opened 2018-07-18
Grid highlight in dark mode not visible (both regular and jupyterlab extension)


timkpaine:
![tmp](https://user-images.githubusercontent.com/3105306/42885194-538d5d2a-8a6e-11e8-838f-989e700719db.gif)

PR #168, opened 2018-07-23
fix for #160

timkpaine:
- [ ] does not use fake class, just uses the (currently unused) td. 
- [x] need to do the fix on jlab plugin too


timkpaine:
original:
tmp1


this pr:
tmp2

texodus:
Looks good, thanks!

Issue #35, opened 2018-02-08
Investigate automatic wasm copying for jupyterlab webpack


timkpaine:
see [this](https://github.com/jupyterlab/jupyterlab/pull/3743) and [this](https://github.com/jupyterlab/jupyterlab/blob/master/jupyterlab/staging/webpack.config.js#L116-L118)

"Basically, if you import a file that is prefixed by JUPYTERLAB_(RAW|URL|FILE)_LOADER_, it will use that particular webpack loader. So if in your build step, you copy the relevant files to your build directory and prefix them with JUPYTERLAB_FILE_LOADER_, they can be bundled up with webpack using its file loader"

texodus:
Thanks, a fix for this has been merged and will be in the 0.1.18 release.

The Jupyter Lab plugin should no longer require additional assets.

PR #170, opened 2018-07-25
Jupyter embedded mode

texodus:
Builds a single fat build of Perspective viewer, plugins, worker and WebAssembly payload, for easier embedding in Jupyter-lab.


texodus:
fixed #35

PR #179, opened 2018-08-04
jupyterlab build complaining

timkpaine:
@texodus can you take a look at this, getting the following error:

screen shot 2018-08-03 at 11 16 08 pm


texodus:
While this patch doesn't break anything, the error message indicates something wrong with your build environment - perpsective.parallel.js should be resolved via `babel-loader` which is already configured to output ES5.  You can try (from project root):

```bash
lerna run clean
lerna clean
lerna bootstrap --hoist
npm run build
cd packages/perspective-jupyterlab
jupyter labextension install .
```

If you're still experiencing issues, feel free to re-open with some more detail about your environment.

PR #186, opened 2018-08-09
Add tests and fix for regressed column update.

RohanPadmanabhan:
Added test case for construction of a table and updating a table in column oriented form (as described in docs). Add fix for issue of list of names and types being appended to even if they are already populated.


texodus:
Looks good, thanks!  This is a needed fix.

PR #189, opened 2018-08-13
Fixing filters when setting multiple filters per column

msturdikova:
Fixing issue when the filters attribute of the perspective-viewer gets incorrectly set when multiple filters are set per column.


texodus:
Looks good, thanks!

PR #196, opened 2018-08-16
Drop python

deepankarsharma:
Drop vestigial pieces of code that are not used anymore. In particular drop things that refer to Python.


timkpaine:
@deepankarsharma you missed a few references.

PR #199, opened 2018-08-19
[WIP] Initial ideas around a React Viewer

LukeSheard:
This is an initial cut around exporting a React component which can be used in applications. I'm yet to put together a working example though. 

## Outstanding Questions

- *What kind of API do we want to export in React?* 

- *How do we expect users to consume `perspective-viewer`?* In trying to setup an example I found that there are imports of css automatically. 

- Tied into above - *is there a way we can export a perspective bundle doesn't doesn't need static assets in relative directories?* This would make the react component more consumable as users would only have to add `wasm-loader` to their webpack config.


texodus:
Thanks for the PR!

In answer to outstanding questions:

* _What kind of API do we want to export in React?_    Not clear we do (see below).  

* _How do we expect users to consume perspective-viewer?_   Granted, examples and documentation need be greatly expanded but some details on why something (e.g. inline styles for `<perspective-viewer>`) is _unexpected_ would help, in framework-agnostic terms.

* _Is there a way we can export a perspective bundle doesn't doesn't need static assets in relative directories?_  There is an asset-less build of `<perspective-viewer>` and all plugins in the `perspective-jupyterlab` module, as it makes installation in Jupyter quite a bit easier.  However, we do not bundle this be default because it is huge (~5mb), and because WASM cannot be inlined in a plain text JS file without encoding it as a string, impacting load time and overall bundle size (this is what `wasm-loader` and `arraybuffer-loader` do).

We are of course interested in improving the developer experience for all Javascript frameworks, but I think at this point this PR is premature (as stated in the title), so I'm going to close it for now.   A few things to please address before you re-open: 

* Splitting the commits into separate PRs for e.g. Travis CI caching, Karma, React, Typescript will greatly expedite them getting merged.

* I think the overall motivation for these changes needs to be clarified:

    - Exporting a `react` specific component in the first place.  `perspective-viewer` uses a web component interface specifically to avoid needing framework-specific bindings, and I'd prefer not maintain them given the breadth of domain expertise necessary.  Is it possible to make `perspective-viewer` more `react` friendly without a new interface?

    - Utilizing `karma` - if there's a justification for using this over `jest`, let's detail it and implement the change across all modules with `node` tests.

    - An entirely new NPM module - can this live in perspective-viewer?

* Assuming a `react` component is warranted, perhaps inheriting from the existing Web Components implementation is not ideal - it currently inlines the `webcomponents.js` polyfill for example.  Can we re-factor `PerspectiveViewer` class in the `perspective-viewer` module to be shared between these implementations without Web Components involved?

* Assuming `typescript` is warranted, I'd prefer we at least share the same `tsconfig.json` settings from `perspective-jupyterlab` so we at least are limited to just 1 dialect of `babel` and 1 dialect of `typescript`.

* I would think `typescript` type definitions would be generally useful outside of a `react` component.

* Please follow the extent conventions for package layout (`src/{lang}`, `test/{lang}`).

* Examples and documentation (as you note).  Even if we ultimately decide against e.g. a `react` specific component, `react` specific examples and documentation are always welcome additions!

PR #204, opened 2018-08-22
Fixed typescript/webpack file resolution error

texodus:


timkpaine:
see how easy that was?

PR #205, opened 2018-08-23
Column oriented partial updates

RohanPadmanabhan:
**Column oriented partial updates:**
Thanks to the great work in #192, partial updates work in column form using `undefined` to show that a value has not changed. For example

```
// a is pKey
{ a : [1, 2, 3], 
  b : ['a', 'b', 'c'],
  c : [true, false, true] }
```
can be update with
```
{ a : [1, 2],
  b : ['d', undefined],
  c : [undefined, true] }
```
to give
```
{ a : [1, 2, 3], 
  b : ['d', 'b', 'c'],
  c : [true, true, true] }
```

This commit extends this idea so that if any columns are not at all, they do not need to be specified. For instance if your table has 100 columns but only one column changes you only need to provide that column and the primary key. This means that instead of passing a delta

```
{ a : [1, 2, 3],
  b : ['d', 'e', 'f'],
  c : [undefined, undefined, undefined],
  ...
  z : [undefined, undefined, undefined] }
```
you only need to pass the delta

```
{ a : [1, 2, 3],
  b : ['d', 'e', 'f'] }
```

**Keys in different order**
This commit also fixes and adds a test for an issue I introduced when fixing the regressed column updates. If the keys (column name) of the update were in a different order to what was originally defined, it would cause an error. Please see test for more details.


RohanPadmanabhan:
Not sure how this change broke the multiple removes test I had written earlier as that test is not column oriented. Any help greatly appreciated

texodus:
Thanks for the PR and detailed report.  I've rebased this branch on master (removing the merge commits), and added an additional fix in #208 - so closing this one.

PR #207, opened 2018-08-25
Correcting sort values in documentation and examples 

msturdikova:


texodus:
Thanks for the PR! 

You are correct that the documentation for the `sort` attribute is in error, and the fix is appreciated. 
 However, the patch itself modifies the wrong files - the modified files e.g. [perspsective-viewer/README.md](https://raw.githubusercontent.com/jpmorganchase/perspective/master/packages/perspective-viewer/README.md) are auto-generated from the JSDoc inlined into the source itself, as per the note on the first line of each of these files:

    

To update the documentation for a module, you must update the [JSDoc in the source itself](https://github.com/jpmorganchase/perspective/blob/master/packages/perspective-viewer/src/js/view.js#L1014), then run `lerna run docs` from project root (or `npm run docs` from the package in question).  Otherwise - these change swill be lost on the next release or docs update.

PR #211, opened 2018-08-26
First & Last by Index Aggregates

neilslinger:
Separating LAST and LAST_VALUE aggregates in JS mapping
Renaming first/last to be first/last by index for clarity
Provide sensible default implementation sorting by pkey
Added tests


texodus:
Looks good, thanks!

PR #212, opened 2018-08-26
WIP: python binding/integration

timkpaine:
- [ ] transfer pypi
- [ ] copyrights
- [ ] sub module?


timkpaine:
ping @texodus and @deepankarsharma for thoughts on how to best do integrations. Also, we can provide a standalone C++ build (just omit the part of the python build that links in binding), but we really need the C++ tests from the athena build.

texodus:
Thanks for the PR!

This looks great but needs some work and planning before we can consider merge (or where even to merge it).  I've cleaned it up a bit on the [python](https://github.com/jpmorganchase/perspective/tree/python) branch for discussion & review.

I concur with @deepankarsharma on all points above, especially the question of licensing which is a critical consideration.

PR #218, opened 2018-08-29
Minor cleanup in gnode_state.cpp

deepankarsharma:
Fuse the two loops into one.


texodus:
Thanks for the PR!

Increasing the CPU utilization limit of Docker is a good idea, but while we're messing around with this let's just make it completely configurable and extend it to `emmake -j%` also.  I've cherry-picked the C++ fix for this into #220, and will add PSP_CPU_COUNT as an environment flag in #221 - so closing this PR.

PR #225, opened 2018-09-01
Cpp tests

deepankarsharma:
This changeset adds build for TBB and GTest and uses those to start the long journey of having C++ tests available.


texodus:
This looks great, thanks for the PR!

PR #229, opened 2018-09-05
Add Travis node_modules caching

LukeSheard:
Should improve travis build times slightly.


texodus:
Thanks for the PR!

While I'm happy to merge this, it's worth noting that we actually [used Travis caching before for node_modules (and docker images) and ended up removing it].(https://github.com/jpmorganchase/perspective/commit/9f69ff58c50664f7ddf60355b98eb49a8a633b95)

When we had it, it seemed the caching did not actually speed up the build at all (and notice the travis build times from this PR are inline with previous builds).  However, they did cause infrequent issues with cached images being out of date, or corrupted images being cached.  I'll keep an eye on this, and if it is still an issue we'll need to back this out.

PR #230, opened 2018-09-07
Added Jupyter API for incremental updates

texodus:
Works with timkpaine/perspective-python#32

### DIFF
--- a/packages/perspective-jupyterlab/src/ts/utils.ts
+++ b/packages/perspective-jupyterlab/src/ts/utils.ts
@@ -7,7 +7,6 @@
  *
  */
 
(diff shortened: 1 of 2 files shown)


timkpaine:
I forgot to update the perspective-python to stop using the docker build required for the cpp extensions, so I just merged the corresponding changes and am testing now.

timkpaine:
should we add a `index` attribute while we're at it?

timkpaine:
https://github.com/timkpaine/perspective-python/issues/33

timkpaine:
async python: https://github.com/timkpaine/perspective-python/blob/master/perspective/helpers.py

PR #231, opened 2018-09-07
1D charts generated from columnar data

sc1f:
- Add axis generator and data parser for columnar data
- Rename old row-based methods for backward compatibility

*Merge only after `to_column` correctly strips total rows.* I included a `.filter()` inside `ColumnarIterator` that can be removed in order to verify that total rows are stripped.


.travis.yml
```diff
@@ -29,7 +29,6 @@ install:
   - npm install
 
 script:
-  - npm run lint
```
texodus: This seems like it was picked up from a merge - should be cleaned up please!

sc1f:
FIXME: charts should not flash on re-render.

PR #232, opened 2018-09-08
Add eslint

LukeSheard:
I found that eslint was already a dependency in some of the perspective packages, along with the eslintrc. I hoisted it to become a build dependency, enabled it in travis and fixed all the errors that it found that I could identify as being fixable.

There are some outstanding comments below where the code confused me.


packages/perspective-viewer/src/js/row.js
```diff
@@ -171,6 +176,8 @@ class Row extends HTMLElement {
         let agg_dropdown = this.querySelector('#column_aggregate');
         let aggregate = this.getAttribute('aggregate');
         if (agg_dropdown.value !== aggregate) {
+            // TODO: Should type be removed here? What is it doing
+            // eslint-disable-next-line no-undef
             agg_dropdown.value = aggregate || perspective.AGGREGATE_DEFAULTS[type];
```
LukeSheard: `type` here seems completely undefined, so I assume that `perspective.AGGREGATE_DEFAULTS[type]` will also return `undefined` in all cases. If so can we just remove the right hand side of the or?

packages/perspective/src/js/perspective.js
```diff
@@ -333,6 +331,8 @@ class ArrowColumnLoader extends TypeVisitor {
         return this;
     }
     // visitNull(type/*: Arrow.type.Null*/) {}
+    // TODO: Should this variable be an argument? 
+    // eslint-disable-next-line no-unused-vars
     visitBool(type/*: Arrow.type.Bool*/) {
```
LukeSheard: There are many of these below, are the `type` arguments really needed?

packages/perspective/src/js/perspective.parallel.js
```diff
@@ -185,6 +185,8 @@ class WebSocketWorker extends worker {
 export default {
     worker: function (url) {
         if (window.location.href.indexOf(__SCRIPT_PATH__.host()) === -1 && detectIE()) {
+            // TODO: Should this be here? Where is it defined?
+            // eslint-disable-next-line no-undef
             return perspective;
```
LukeSheard: Where does `perspective` come from in this branch of code? Seems to be completely undefined in IE if I read this correctly.

packages/perspective-viewer/src/js/row.js
texodus: Yes, aggregate is always defined now - this looks like a legitimate catch by eslint!

packages/perspective/src/js/perspective.js
texodus: No, likely not.  `@apache-arrow/es5` defines `TypeVisitor` which uses defines this variable in Typescript - but these are certainly vestigial wrt this module.

packages/perspective/src/js/perspective.parallel.js
texodus: This is a nasty edge case, specifically regarding when IE tries to load perspective as a WebWorker, but as a cross-origin request.  This is not supported by IE at all as far as I can tell, so we instead download the inline version of the asmjs build of perspective, by [inlining the script tag into `<head>` via Javascript](https://github.com/jpmorganchase/perspective/blob/master/packages/perspective/src/js/perspective.parallel.js#L25).  This binds the `perspective` symbol globally, and a reference to this symbol is returned here when an attempt to instantiate a worker is made.

LukeSheard: Ah okay, that makes sense. In that case it should be enough to change this to `window.perspective` to indicate that we're using a global variable. I'll update the PR with that change.

packages/perspective-viewer/src/js/row.js
LukeSheard: +1 for tooling! I'll remove the right hand side and remove the comment.

texodus:
Thanks for the PR!  This is a nice improvement to the overall code quality.

I've attempted to address the issues cited in review, rebased the branch on master, removed the merge commit and squashed 644b36d.  Updated PR is at #236, to which I will add you as a reviewer.

PR #233, opened 2018-09-11
Null fix

texodus:
Fixed 'null' updates and initial loads.


deepankarsharma:
> See change to gnode_state.cpp lines 298-304 above: being able to express that is what the rest of the delta is about.

Thanks for the clarification, it was very helpful.

texodus:
Thanks @sc1f and @neilslinger for the commits, as well as @deepankarsharma and @sl236 for reviewing!  

I suspect there are more issues lurking with `null` support, but this is a substantial improvement already.

PR #236, opened 2018-09-11
Add ESLint

texodus:
Address the issues cited in review of #232 , rebased the branch on master, removed the merge commit and squashed 644b36d.


packages/perspective/src/js/perspective.parallel.js
```diff
@@ -185,6 +185,8 @@ class WebSocketWorker extends worker {
 export default {
     worker: function (url) {
         if (window.location.href.indexOf(__SCRIPT_PATH__.host()) === -1 && detectIE()) {
+            // TODO: Should this be here? Where is it defined?
+            // eslint-disable-next-line no-undef
             return perspective;
```
LukeSheard: Shouldn't this be `window.perspective`? Also it might be wise to add a comment here explaining how this comes to be for future reference.

texodus:
Right on, thanks!

PR #224, opened 2018-09-01
Added clang-format file

deepankarsharma:
Added a clang format file to ensure uniform formatting of C++ code.

PR #242, opened 2018-09-18
Add eslint-prettier

LukeSheard:
With similar intentions to #224 this adds `eslint-config-prettier`and `eslint-plugin-prettier` to the eslint config, and adds a pre-commit hook which automatically formats all the javascript files for you. 

* 4a534b7 adds the configuration to check for style errors
* 33592ba adds the pre-commit hook
* babb867 is an autoformat of the codebase to make sure this PR passes CI.


texodus:
This is great, thanks for the PR!

As Perspective is a multi-language project (C++, Javascript and soon Python, plus non-languages like HTML and LESS), I believe it is important to impose at least some basic consistency across the codebase proper where applicable.  We'll make a similar request to #224 when I have time to review the `clang-format` options in more detail, but for now I believe these overrides to `prettier` defaults are justified:

* [`bracketSpacing: false`](https://prettier.io/docs/en/options.html#bracket-spacing) IMO the default is not consistent with the rest of the project - it is e.g. not the default for `clang-format` braced initializers, which they justify under `Cpp11BracedListStyle` [here](https://clang.llvm.org/docs/ClangFormatStyleOptions.html).
* [`tabWidth: 4`](https://prettier.io/docs/en/options.html#tabs) Again, simply for consistency with #224 and the extent LESS, HTML, JSON, etc formatting.
* Worth also mentioning that we should probably have consistent `printWidth`, as I (and I assume any other developer) will not be switching IDEs when implementing changes across several Perspective languages.  I do tend to agree however that #224's `printWidth` is a bit on the conservative side, so this is probably fine at 200 for now.

Personally, I am disinterested in the details of code formatting outside of consistency - assuming you concur, would you mind amending this commit by adding the above 2 overrides to the `prettier` configuration in [4a534b7](https://github.com/jpmorganchase/perspective/commit/4a534b7d5aa39728f175d774900e9ccea7eb0f79), and redoing [babb867](https://github.com/jpmorganchase/perspective/commit/babb8673a626b2e324a073e3d898173196448e11) to respect these changes?

LukeSheard:
Very reasonable request, I'll update the options now and re-run babb867.

PR #246, opened 2018-09-24
Column resize fix refactor

texodus:
Fixed issue with minimum widths of column selector when in collapsed mode, as well as autofix tasks for LESS and HTML to complement the JS and C++ ones added earlier.


package.json
```diff
@@ -30,7 +31,7 @@
 		"postinstall": "lerna bootstrap --hoist",
 		"precommit": "npm run lint:autofix && git add . -u",
 		"lint": "eslint packages/*/src/**/*.js packages/*/test/**/*.js",
-		"lint:autofix": "npm run lint -- --fix && lerna run lint:autofix:clang",
+		"lint:autofix": "npm run lint -- --fix && lerna run lint:autofix:clang && css-beautify packages/**/src/less/*.less -r && html-beautify packages/**/src/html/*.html -r",
```
LukeSheard: Can we split these into parallel tasks using `npm-run-all` ??

texodus: Can and should and done.

PR #249, opened 2018-09-28
Refactor examples

texodus:
Deprecate `perspective-examples` module in favor of a top-level `examples/` directory.


examples/git_history/index.html
```diff
@@ -14,20 +14,28 @@
 
     <meta name="viewport" content="width=device-width, initial-scale=1, maximum-scale=1, minimum-scale=1, user-scalable=no">
 
-    <script src="mobile_fix.js"></script>
     <script src="perspective.view.js"></script>
```
LukeSheard: I think all of these can be refactored to be from `node_modules/@jpmorganchase/perspective`.

packages/perspective-viewer-highcharts/src/config/highcharts.plugin.esm.config.js
```diff
@@ -3,8 +3,9 @@ const common = require("@jpmorganchase/perspective/src/config/common.config.js")
 
 module.exports = Object.assign({}, common(), {
     entry: "./src/js/highcharts.js",
+    externals: [/^[a-z0-9\@].*$/],
     output: {
-        filename: "highcharts.plugin.js",
+        filename: "highcharts.plugin.esm.js",
         library: "perspective-view-highcharts",
         libraryTarget: "umd",
```
LukeSheard: Can this be changed to be `commonjs2` and the name to `cjs` since this is not `esm` target.

packages/perspective-viewer-hypergrid/src/config/hypergrid.plugin.esm.config.js
```diff
@@ -3,8 +3,9 @@ const common = require("@jpmorganchase/perspective/src/config/common.config.js")
 
 module.exports = Object.assign({}, common(), {
     entry: "./src/js/hypergrid.js",
+    externals: [/^[a-z0-9\@].*$/],
     output: {
-        filename: "hypergrid.plugin.js",
```
LukeSheard: Same naming conflicts with the highcharts plugin.

packages/perspective/src/js/perspective.node.js
```diff
@@ -30,46 +30,78 @@ delete module.exports["worker"];
 
 let CLIENT_ID_GEN = 0;
 
+const DEFAULT_ASSETS = [
```
LukeSheard: I think these URLs should be `require.resolve` so that the WebSocket host can be in a directory that is not the root of the package, e.g. I have a structure like:

```
/
   package.json
   src/
       server.js
```

for my server.

examples/git_history/index.html
texodus: This is specific to formatting `<perspective-viewer>` to not scroll the viewport on mobile because these demos are full-page - I believe this is specific only to these examples' usage of `<perspective-viewer>`.

However, we don't host these sites anymore so no need for a mobile_fix at all - will delete this.

packages/perspective/src/js/perspective.node.js
texodus: Good call.

examples/git_history/index.html
texodus: Actually scratch that comment - this commit is outdated, this was already deleted.

packages/perspective/src/js/perspective.node.js
LukeSheard: @texodus Also might be an idea to try catch the plugins which are not part of `@jpmorganchase/perspective`.

texodus: Turns out this was tricky to do, as none of these are deps of @jpmorganchase/perspective itself - ended up using resolve with a modified `paths` to point to `process.cwd()`.  This will work as long as all `@jpmorganchase/*`packages are hoisted in your project ... unclear of another way to do this unfortunately.

texodus:
Changes applied, thanks!

packages/perspective/src/js/perspective.node.js
```diff
@@ -14,11 +14,17 @@ const perspective = require("./perspective.js");
...
 const http = require("http");
 const WebSocket = require("ws");
+const process = require("process");
 
 const path = require("path");
 
 const load_perspective = require("../../obj/psp.sync.js").load_perspective;
 
+// eslint-disable-next-line no-undef
+const RESOLVER = typeof __non_webpack_require__ !== "undefined" ? __non_webpack_require__.resolve : module.require.resolve;
```
LukeSheard: What's the difference between this and just using `require.resolve`?

PR #255, opened 2018-10-02
Add 'not in' filter operation

nmichaud:
For columns with lots of possible string values, 'not in' can be more efficient than an 'in' filter with all the other values.


texodus:

PR #258, opened 2018-10-04
Fix bug with pkey column update with missing agg value

deepankarsharma:
Fix bug reported by @nmichaud where there was an issue with updates specifying the pkey column where value for agg column is missing. Added a test for this as well.

### DIFF
--- a/packages/perspective/test/js/pivots.js
+++ b/packages/perspective/test/js/pivots.js
@@ -207,6 +207,42 @@ module.exports = perspective => {
             expect(answer).toEqual(result2);
         });
 
+        it("['x'] test update pkey column", async function() {
+            const schema = {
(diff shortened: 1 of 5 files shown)


packages/perspective/test/js/pivots.js
```diff
@@ -207,6 +207,42 @@ module.exports = perspective => {
...
+
+            table.update(rec1);
+
+            let view = table.view({
+                row_pivot: ["id"],
+                aggregate: [{op: "sum", column: "pos"}]
+            });
+
+            let result1 = await view.to_json();
+            console.log(result1);
```
texodus: Watch out for spurious console statements :)

texodus:
Thanks for the PR!

PR #259, opened 2018-10-04
Fixed bug which caused computed-columns creation to break updates in …

texodus:
…perspective-viewer.

### DIFF
--- a/packages/perspective/src/js/perspective.js
+++ b/packages/perspective/src/js/perspective.js
@@ -1306,12 +1306,19 @@ module.exports = function(Module) {
         let pdata;
         let cols = this._columns();
         let schema = this.gnode.get_tblschema();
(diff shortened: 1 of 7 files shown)


packages/perspective/src/js/perspective.js
```diff
@@ -1314,6 +1315,12 @@ module.exports = function(Module) {
             pdata = parse_data(data, cols, types);
         }
 
+        for (let i = 0; i < names.size(); i++) {
+            if (cols.indexOf(names.get(i)) === -1) {
+                pdata.types.splice(i, 1);
+            }
+        }
```
nmichaud: I'm not sure this works the way you intend. As you splice values from the types array, the index value derived from the name array will no longer be valid.

PR #266, opened 2018-10-05
Move C++ to top level directory

LukeSheard:
This PR moves the C++ out of the perspective package and into a generic `/src` directory in the repo. This provides the opportunity to make the build process for emscripten and any other bindings not coupled to the Lerna lifecycle of perspective package.


texodus:
I've restructured a bit of the file layout and cleaned up a few things, as well as rebased the branch on master to resolve the conflicts.  I'm also pretty sure this PR breaks the nascent C++ tests, but this is not your responsibility to fix, considering they are not run as part of the build or Travis.

Thanks for the PR - not only is this a substantial in code quality, but it makes the build stdout prettier, which I'm a huge fan of 👍 💯 🥇

PR #273, opened 2018-10-11
Expose view columns as TypedArrays

sc1f:
Create C++/JS API to return view columns as a `TypedArray` for faster consumption and transferable usage.


texodus:

PR #274, opened 2018-10-11
Computed columns broken with delta updates on non-dependent columns

nmichaud:


texodus:
Thanks for the high quality PR!

PR #275, opened 2018-10-11
Cleanup cpp api

nmichaud:
Also fix a slight memory leak.


texodus:
Thanks for the PR!  

I think its about time we discussed how we want update semantics to work in general, and it would be nice to see whatever we decide to be well asserted with tests & benchmarks as well.  For now, this is a strict improvement on the haphazard delta implementation we use ...

PR #277, opened 2018-10-12
Expansion depth should be stable across updates

nmichaud:
When a context has the expansion depth set, new nodes that tick in aren't automatically expanded to that depth. This PR fixes that.


texodus:

packages/perspective/src/js/perspective.js
```diff
@@ -797,37 +797,21 @@ module.exports = function(Module) {
     };
 
     /**
-     * Expand the tree down to `depth`.
+     * Set expansion `depth` pf the pivot tree.
      *
      */
-    view.prototype.expand_to_depth = async function(depth) {
+    view.prototype.set_depth = async function(depth) {
```
LukeSheard: Given that this is a breaking change would it not be better to deprecate the old functions? (Unless they’ve not been released from master yet?)

PR #278, opened 2018-10-13
Remove subpath imports from external packages

LukeSheard:
Since perspective.parallel.js is the entrypoint for `@jpmorganchase/perspective` this import is a no-op but means that all the files are processed again.

### DIFF
--- a/packages/perspective/src/js/perspective.js
+++ b/packages/perspective/src/js/perspective.js
@@ -7,7 +7,7 @@
  *
  */
 
-import {AGGREGATE_DEFAULTS, FILTER_DEFAULTS, SORT_ORDERS, TYPE_AGGREGATES, TYPE_FILTERS, COLUMN_SEPARATOR_STRING} from "./defaults.js";
+import {AGGREGATE_DEFAULTS, FILTER_DEFAULTS, SORT_ORDERS, TYPE_AGGREGATES, TYPE_FILTERS, COLUMN_SEPARATOR_STRING} from "../../../../src/js/constants";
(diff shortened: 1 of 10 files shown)


packages/perspective/src/js/perspective.js
```diff
@@ -7,7 +7,7 @@
  *
  */
 
-import {AGGREGATE_DEFAULTS, FILTER_DEFAULTS, SORT_ORDERS, TYPE_AGGREGATES, TYPE_FILTERS, COLUMN_SEPARATOR_STRING} from "./defaults.js";
+import {AGGREGATE_DEFAULTS, FILTER_DEFAULTS, SORT_ORDERS, TYPE_AGGREGATES, TYPE_FILTERS, COLUMN_SEPARATOR_STRING} from "../../../../src/js/constants";
```
LukeSheard: I don't necessarily like this, but it works in this instance..

texodus:
Thanks for the PR!

I think this PR needs some clarification over the intended fix.  I am specifically not clear exactly that "all the files are processed again" - if this is in reference to a sub-import causing the entire `@jpmorganchase/perspective` module to be compiled twice (?), the [docs on Webpack Module Path imports](https://webpack.js.org/concepts/module-resolution/#module-paths) suggest that this is not the case, rather that just the one file `defaults.js` is re-processed.

Regardless, the symbols from `defaults.js` are clearly intentionally public - seems a much better solution to this is to make them explicit exports of `@jpmorganchase/perspective`, as [some of them already are](https://github.com/jpmorganchase/perspective/blob/master/packages/perspective/src/js/perspective.parallel.js#L187)

Issue #257, opened 2018-10-03
Computed Column UX bugs


texodus:
* Column drop target flickers when a column is dragged out of the computed column selector
* Selected columns are cleared when the computation type is changed
* Duplicate named columns silently fail
* Error messages are nice - but IMO we should replace these with just disabling the `Save` button if the inputs are invalid
* Dark theme has some missing colors

sc1f:
Fixed dark theme and removed error messages.

texodus:
Merged

PR #283, opened 2018-10-16
Fix computed column UX issues

sc1f:
#257 all issues have been fixed.

### DIFF
--- a/packages/perspective-viewer/src/js/utils.js
+++ b/packages/perspective-viewer/src/js/utils.js
@@ -7,6 +7,33 @@
  *
  */
 
+/******************************************************************************
+ *
+ * Drag & Drop Utils
+ *
+ */
+
+global.dragEnter = function dragEnter(ev) {
+    ev.stopPropagation();
+    ev.preventDefault();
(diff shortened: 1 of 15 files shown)


packages/perspective-viewer/src/js/utils.js
```diff
@@ -7,6 +7,33 @@
  *
  */
 
+/******************************************************************************
+ *
+ * Drag & Drop Utils
+ *
+ */
+
+global.dragEnter = function dragEnter(ev) {
```
LukeSheard: While we're here can remove these global event handlers which pollute the global namesapce and change them to be event Listeners on the elements when they're created in the element?

sc1f: The whole drag-and-drop implementation definitely needs a rework - I'm for refactoring the implementation but I'll defer to @texodus on whether it should be part of this PR or a different one.

texodus: This should be fixed but I agree this PR is not about cleaning up drag & drop generally - I've added a task to the project tracker to tackle this in a future sprint.

PR #285, opened 2018-10-21
Get depth API

nmichaud:
Forgot to include this in #277

### DIFF
--- a/src/cpp/context_one.cpp
+++ b/src/cpp/context_one.cpp
@@ -239,6 +239,11 @@ t_ctx1::set_depth(t_depth depth) {
     m_depth_set = true;
 }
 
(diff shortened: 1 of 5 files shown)


texodus:
Thanks for the PR!

I think this patch needs a bit more work before we can merge it.  We've updated our [contributing.md](https://github.com/jpmorganchase/perspective/blob/master/CONTRIBUTING.md) to make our PR guidelines more explicit.  Please take a look - specifically for this change, I think we'd need

* An explanation of what this API does and why it needs to be written in C++;  as it stands this information can be much more efficiently calculated from the `view()`s config object.
* A Javascript API
* Tests and JSDoc

PR #310, opened 2018-11-14
Jupyterlab plugin: Cleanup and computed columns

timkpaine:
- Removing stale dependencies
- removing comm feature (no longer needed)
- adding computed columns
- adding embed hook, will require much more work so separating from this 
- bump copyright year


texodus:

PR #312, opened 2018-11-14
docs: update installation guide

ColinEberhardt:
It looks like you need to load perspective.js in order to instantiate the 'engine':

```
const worker = perspective.worker();
```

### DIFF
--- a/docs/installation.md
+++ b/docs/installation.md
@@ -47,6 +47,7 @@ library can be used directly from
 CDN by simply adding these scripts to your `.html`'s `<head>` section:
 
 ```html
+<script src="https://unpkg.com/@jpmorganchase/perspective/build/perspective.js"></script>
 <script src="https://unpkg.com/@jpmorganchase/perspective-viewer/build/perspective.view.js"></script>
 <script src="https://unpkg.com/@jpmorganchase/perspective-viewer-hypergrid/build/hypergrid.plugin.js"></script>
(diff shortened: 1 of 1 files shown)


texodus:
Thanks for the PR!  Much appreciated.

To clarify, `perspective.js` is the API for the WebWorker.  This is already a dependency of `perspective.view.js`, but this bundle does not expose it as a global variable;  rather, it will internally create it's own WebWorker as needed to manage any data passed to it, which you can access [via the `<perspective-viewer>`'s `worker` property](https://jpmorganchase.github.io/perspective/docs/perspective-viewer.html#worker) like so

```javascript
let elem = document.getElementById('my_viewer');
let table = elem.worker.table([{x:1, y:2}]);
elem.load(table);
```

Thus, `perspective.js` is only needed if you A) want to create additional WebWorkers or B) just want to use the Perspective engine library directly.  We could alternatively just export the `perspective` symbol globally from `perspective.view.js` - feedback welcome here.  Either way, this part of the documentation needs some substantial fleshing out!

ColinEberhardt:
Thanks for merging the PR.

> Either way, this part of the documentation needs some substantial fleshing out!

Your API documentation is really quite good. What you are lacking is any 'getting started' type of documentation. These are typically much more tutorial-based.

As an example, the following blog post, which introduces d3fc (our charting component library), has been very popular:

https://blog.scottlogic.com/2018/09/21/d3-financial-chart.html

PR #315, opened 2018-11-15
Fix weighted mean null/missing value handling

neilslinger:
Weighted mean aggregation calculation now correctly handles null/missing values


texodus:
Thanks for the PR and test!

PR #320, opened 2018-11-20
Use Yarn workspaces as a drop-in for Lerna

LukeSheard:
After some experimentation I found that using [Yarn Workspaces](https://yarnpkg.com/lang/en/docs/workspaces/) does the same job as using `lerna bootstrap --hoist` without the overhead that it introduces. I found that on my Mac that for a fresh clone of the repository this reduces the initial `yarn` (previously `npm install`) from over `2 minutes` to `13 seconds`. 

Currently this fails the build because `npm-run-all` identifies the project as a yarn project and `yarn` isn't available in the `Dockerfile` for `perspective/emsdk`. I've updated the `Dockerfile` and validated this works locally, but would need to be published as `perspective/emsdk:latest` for this CI to work. I can provide a standalone PR for the `Dockerfile` if we want to validate that change though?


texodus:
Thanks for the PR!

This is a great addition.  I've altered this branch's history prior to merge as it was not compiling:

* removed the merge commit and rebased on master [contributing.md].(https://github.com/jpmorganchase/perspective/blob/master/CONTRIBUTING.md)
* Published an updated `perspective/emsdk` image to Dockerhub with yarn and emscripten 1.38.20
* Fixed incorrect `yarn.lock` versions.
* Rewritten the test harness to handle yarn's new node_modules layout.

PR #322, opened 2018-11-21
Move Apache Arrow into Perspective Code

LukeSheard:
Currently it's not possible to import `@jpmorganchase/perspective` without yourself installing `@apache-arrow/es5-esm` as it's not listed as a dependency. 

We know that the currently published module depends on a git repo, rather than an NPM package. I've added a script which copies the content of the package into the perspective package, updated the imports to use those local files, and added the required dependencies from arrow.


LukeSheard:
@texodus Any idea why the PR build would be failing for this? Seems like the docker image for testing isn't available. 

```
Unable to find image 'perspective/puppeteer:latest' locally
docker: Error response from daemon: unauthorized: authentication required.
See 'docker run --help'.
```

texodus:
Thanks for the PR!

> Currently it's not possible to import @jpmorganchase/perspective without yourself installing @apache-arrow/es5-esm as it's not listed as a dependency.

This is not the case - the dependency is inlined in the node build and WebWorker, and is never required as a dependency.  Ergo, I don't think this patch is necessary - please let me know if you believe this to be closed in error though!

Issue #318, opened 2018-11-19
PerspectivePlugin lack dependencies


dknfeiov:
node@8.11.2
perspective@0.2.7

Hi ,  i use perspective by cdn before , everything is ok , but now i want to use webpack with the plugin PerspectivePlugin supported
![image](https://user-images.githubusercontent.com/8175516/48683864-18751100-ebea-11e8-8112-f705e0f130a8.png)
and import plugins for recommend
![image](https://user-images.githubusercontent.com/8175516/48683913-4c503680-ebea-11e8-8d50-46b2a06c481a.png)
some errors happend
![image](https://user-images.githubusercontent.com/8175516/48683986-9a653a00-ebea-11e8-8d69-0a0b57dc0040.png)
It is look like some dependence haven't list  in the package.json 
i view the "@jpmorganchase/perspective-viewer" package, the webpack-plugin.js use some loader and other dependences, but the package.json don't list
![image](https://user-images.githubusercontent.com/8175516/48684140-49a21100-ebeb-11e8-9c6e-8e121c471904.png)
![image](https://user-images.githubusercontent.com/8175516/48684159-5d4d7780-ebeb-11e8-9e02-86cb664e0650.png)

is the reason caused the problem ?

LukeSheard:
@dknfeiov I noticed this as well over the weekend. I've started work trying to fix the underlying issues here: https://github.com/jpmorganchase/perspective/tree/webpack-plugin

texodus:
Merged & release in 0.2.9

PR #327, opened 2018-11-26
UX Cleanup

texodus:
* Render size warnings now show the set size in the warning message.
* The `preload` script attribute works on a library-level singleton.


texodus:
#318

PR #329, opened 2018-11-28
Refactor parse_data

sc1f:
- Introduce `DataParser` class as handler for data conversion
- Break up and remove old implementation of `parse_data`


texodus:
Nice work, this refactor looks great!

PR #335, opened 2018-12-04
Build into dist, plays nicer with ipywidgets

timkpaine:
`ipywidgets` embedding expects the output to be dist:
https://github.com/jupyter-widgets/ipywidgets/blob/master/packages/html-manager/src/libembed-amd.ts#L46


texodus:
Looks good, thanks!

PR #336, opened 2018-12-07
Port data_types and get_data_type

sc1f:
- Port `data_types` and `get_data_type`
- Remove vestigial code in DataParser
- Add test (skipped) for nasty edge case of upgrading int to float columns


texodus:
Thanks for the PR, looks good!

PR #346, opened 2018-12-19
making cpp a first class citizen again

timkpaine:
- move CMakeLists.txt to top level where it belongs :-)
- adding separate commands to build and test cpp code (@LukeSheard  can you take a look that my hooks in the package.json's are sensible)
- adding support  for TBB and Boost discovery outside of emscripten
- adding in @nmichaud and @deepankarsharma gtest hook


packages/perspective/package.json
```diff
@@ -24,13 +24,17 @@
         "bench": "npm-run-all bench:build bench:run",
         "prebuild": "mkdir -p build && CLICOLOR_FORCE=1 yarn compile",
         "precompile": "mkdir -p obj && mkdir -p ../../obj/",
-        "compile": "cd ../../obj/ && emcmake cmake ../scripts/ && emmake make -j${PSP_CPU_COUNT-8}",
+        "compile": "cd ../../obj/ && emcmake cmake ../ && emmake make -j${PSP_CPU_COUNT-8}",
```
LukeSheard: IMO this stuff should all just be in the TLD for the project.

texodus: This build is a dependency of `@jpmorganchase/perspective`, so as a developer, I'd like it only run when I build this package, not e.g. when I build `@jpmorganchase/perspective-jupyterlab`.

texodus:
Thanks for the PR!

PR #352, opened 2018-12-26
Unified data access interface

sc1f:
- Move data ingestion and metadata inference into C++
- Create a unified API for accessing data regardless of underlying format (columnar vs. row-based)
- Ported inference functions return native C++ types


packages/perspective/src/js/perspective.js
```diff
@@ -1572,35 +1561,24 @@ export default function(Module) {
...
-                pdata = parser.parse(__MODULE__, data);
-
-                if (pdata.row_count > CHUNKED_THRESHOLD) {
-                    let new_pdata = [];
-                    while (pdata.cdata[0].length > 0) {
-                        const chunk = pdata.cdata.map(x => x.splice(0, CHUNKED_THRESHOLD));
-                        new_pdata.push(Object.assign({}, pdata, chunk));
-                    }
-                    pdata = new_pdata;
-                } else {
```
texodus: This is a bit tricky - unsure if we will need this again in the future.  It was oroginalyl added to prevent `vecFromJSArray()` calls in C++ from blowing up the wasm heap, but data accessor no longer calls this per column.

sc1f: We do call `vecFromJSArray()` for each load, update, and delete *operation* instead of per-column, so there is still the danger of blowing up the WASM heap even with this change. 

I was working on a version of the accessor that did not rely on the JS side passing in column names and data types, as we already have that information stored on the C++ side (thus avoiding the `vecFromJSArray()` casting), but I ran into some non-deterministic `memory access out of bounds` issues. 

Theoretically it's possible to access the already-created `gnode`  in `make_table`, grab the schema from there, and then grab the names and types. Actually doing it, however, creates a bunch of errors that I haven't been able to nail down yet.

texodus:
Thanks!  This is a great addition, very nice work!

PR #354, opened 2018-12-31
Test quality-of-life updates

sc1f:
- Add the following flags to the test script:
  - `--bail`: quits after the first test suite failure
  - `--debug`: preserves all console statements
- Make order of `expect().toEqual()` in Perspective.js tests consistent.


texodus:
Thanks for the PR!

The mixed-up `expect()` order was especially annoying - very nice to see this finally fixed!

PR #356, opened 2019-01-02
Python Bindings

timkpaine:
- implemented with boost::python
- minimal for now until API fixes can occur
- added python test suite
- hooked CPP only and CPP+Python builds to travis matrix

- Unrelated -> CMakeLists enhancement to not redownload and build third party if already present


timkpaine:
@texodus this is good to go, I can wait and refactor after your #355. I might need to tweak the caching to make the build faster

texodus:

PR #362, opened 2019-01-05
speed up cpp and python builds

timkpaine:


timkpaine:
back to <11min from mid 30s

texodus:
Thanks for the PR!

I've addressed this in a different way in #363 - separating teh docker builds from the build commands themselves, and publishing the images to the [perspective dockerhub](https://hub.docker.com/u/perspective) account instead.  If there are other changes you think need to be merged from this branch, let me know.

PR #365, opened 2019-01-06
Test and lint python

timkpaine:
- remove python-specific makefile for scripting
- remove docs in python folder
- hook in test
- hook in lint
- add publish_python


texodus:

PR #367, opened 2019-01-06
Promote column

texodus:
An optimization to type inference to prevent reading all numeric JSON data twice, in order to discriminate which numeric type to apply to the column.


src/cpp/column.cpp
```diff
@@ -911,6 +911,13 @@ t_column::unintern_c(t_uindex idx) const {
     return m_vocab->unintern_c(idx);
 }
 
+std::string
+t_column::str() const {
```
sc1f: I added `str()` on columns for debugging, but we can take it out before merging. Not sure what use it would have in production.

texodus: Ok - every byte counts!

Issue #368, opened 2019-01-07
Infer Schema functionality breaks some examples


matt-hooper:
It appears that the infer schema functionality changed between 0.2.9 and 0.2.11. This has broken some json based examples.

For example, the cryptocurrency example does not work using 0.2.10 (and above) of perspective.
The example reads from a json feed and auto-generates the table schema. 

The generated schema is different between the two versions (even though the data from the feed is identical):

|Version 0.2.9|Version 0.2.10|
|-----------|--------------|
|client_oid: "string"|-| 
|order_id: "string"|order_id: "string"|
|order_type: "string"|-|
|price: "float"|price: "float"|
|product_id: "string"|product_id: "string"|
|-|remaining_size: "float"|
|sequence: "float"|sequence: "float"|
|side: "string"|side: "string"|
|size: "float"|-|
|time: "datetime"|time: "datetime"|
|type: "string"|type: "string"|

The renaming of the *size* field to *remaining_size* breaks the example:
![image](https://user-images.githubusercontent.com/33057010/50763849-e9cdfb00-1268-11e9-8308-27d65ca9d2f4.png)

I have also seen a similar issue with a different json based data source.
See attached example forecast_min.zip
[forecast_min.zip](https://github.com/jpmorganchase/perspective/files/2732506/forecast_min.zip)

This example reads data from a json file. 
When using 0.2.9 of perspective, the types of the data are correctly inferred:

![image](https://user-images.githubusercontent.com/33057010/50765340-85fa0100-126d-11e9-8483-b30d737a573c.png)

When 0.2.11 is used, the types are all read as string values. None of the number types are inferred:

![image](https://user-images.githubusercontent.com/33057010/50765429-c5285200-126d-11e9-9037-815438552d73.png)

Is this new functionality intended?
Is it possible to force perspective to work in the old way and auto infer the number values (without manually passing a pre-defined schema)?

sc1f:
Definitely a bug inside the type parser—currently working on a fix. Thanks for the report!

sc1f:
@matt-hooper is the crypto example you mention the one on [bl.ocks.org](https://bl.ocks.org/ColinEberhardt/6e287f871410ecd970b038343b166514)? I have the issue fixed for your forecast data, but I want to see if it fixes the issue on the crypto example as well. Thanks!

matt-hooper:
[...]

PR #370, opened 2019-01-08
Correctly type parse numbers wrapped in strings

sc1f:
- Fix issue #368 in data parser
- Use `std::min` instead of embind to JS Math module


texodus:
Thanks for the PR!  Looks good

PR #373, opened 2019-01-09
Fixes local cpp build -> need to cd into build dir (not needed on docker, working dir is set)

timkpaine:


texodus:
Thanks for the PR!  This was a silly omission on my part ...

PR #374, opened 2019-01-10
remove boost::format

timkpaine:
for both size reasons, and because it causes issues on the (future) arrow build.


texodus:
Thanks for the PR!

Issue #349, opened 2018-12-19
UI filters should give user feedback when they are valid


texodus:

Ro4052:
I've been having a look at this issue but wasn't sure of the best way to approach the problem. My current approach is a function on the Perspective table to check whether a filter is valid, which can be called from the viewer. Invalid filters can then be highlighted and not passed with the other filters into the new view. One area to investigate is the best way to highlight a filter, I'm currently just using a dashed red line.

![invalid_filter](https://user-images.githubusercontent.com/29425751/50779920-8f00c780-1299-11e9-8d5a-944251a3ccb1.PNG)

I'm not sure if this is the most appropriate approach but hopefully it’s a start for addressing this issue.

texodus:
This looks rad!  Pretty much precisely what I had in mind - to give the user some feedback whether the filter they've entered has affected the data set.  I think we would pretty happily accept a PR as is.

Re: styling - we've been applying Material guidelines as convenient, so if you are looking for inspiration the [Google Material Guidelines](https://material.io/design/components/text-fields.html#anatomy) are pretty much indisputable (for the material.css themes at least) - these do a red underline, red overlabel and red warning icon right floated.  I think at this point in development though, these should be viewed as inspiration rather than devotion - what you have looks good to me and is certainly a clear functional improvement over what's in 0.2.11

Ro4052:
I've had a little look at the material guidelines.

![invalid_filter_material](https://user-images.githubusercontent.com/29425751/50836807-05610080-1352-11e9-8745-1ed76d903fb2.PNG)

Would this be more suitable? This approach ended up being a little more involved but is perhaps more appropriate?

texodus:
Looks *awesome*

Ro4052:
Thanks for the feedback. It's still a bit of a work in progress and needs some cleaning up. I'm still also looking at ways to improve the styling.

texodus:
Thanks for the report, and fix!  This has been released in 0.2.12

PR #375, opened 2019-01-10
Invalid filters

Ro4052:
#349 UI filters should give user feedback when they are valid

* Exposed new table method for determining if a filter is valid
* Viewer validates filters before passing to table
* Invalid filters operands are underlined in red with a red exclamation mark next to them
* Filtering by an empty string is now invalid (this seems more appropriate for the current viewer UI)
* Added a screenshot test for the new invalid filter highlighting
* Added test for filtering by empty string
* Added tests for newly exposed is_valid_filter method

invalid_filter_exclamation


texodus:
Thanks for the PR!  This is a great addition - User experience is awesome, and the tests in particular are greatly appreciated!

I think as a future enhancement we may want to reconsider [this API](https://github.com/jpmorganchase/perspective/pull/375/files#diff-05ea229e1b4c3f912beed3a82b83861cR168), in favor of general validation across APIs which take configuration objects as parameters - see also [my comments on API differences between viewer and library](https://github.com/jpmorganchase/perspective/issues/314).  I think this will help keep the API surface small, predictable, as well as portable and consistent between host languages.

As this is currently undocumented, IMO it is fine as is for now.

PR #378, opened 2019-01-12
Add pre-compiled JavaScript artifacts for Perspective package

LukeSheard:
This PR does two things to enhance and improve the development workflow for apps running on the webpack plugin provided by `@jpmorganchase/perspective`:

1. Split the plugin itself into a separate package - this reduces overhead for installation when end users are only the UMD bundles. 
2. The plugin now uses pre-compiled JS assets (via `@babel/cli`). This reduces the overhead on webpack for end users since it doesn't need to run babel itself on every build for the package anymore. 
3. Webpack plugin now takes options for the various loaders needed during compilation so an end user can choose where the worker code will be written to (useful for caching). 
4. Perspective Viewer + Plugins now exports a `CJS` module with it's dependencies externalised.


LukeSheard:
@texodus any ideas why the `/pr` build failed for a screenshot change? Not sure how this change could affect that one individually. 

https://travis-ci.org/jpmorganchase/perspective/jobs/478847521#L870

packages/perspective-webpack-plugin/index.js
```diff
@@ -0,0 +1 @@
+module.exports = require("@jpmorganchase/perspective/webpack-plugin");
```
ColinEberhardt: When I saw this PR I was expecting to see `webpack-plugin` and its associated loaders moved into the `perspective-webpack-plugin` project. Is there any reason why that is not the case?

packages/perspective-jupyterlab/src/config/plugin.config.js
```diff
@@ -25,11 +25,45 @@ module.exports = {
     plugins: [new webpack.ContextReplacementPlugin(/moment[\/\\]locale$/, /(en|es|fr)$/), new PerspectivePlugin()],
     module: {
         rules: [
+            {
```
texodus: We should not remove the `less` loaders here.  Note there is no `style-loader` - these are not CSS asset files, they are [inline styles for the shadow DOM](https://developer.mozilla.org/en-US/docs/Web/Web_Components/Using_shadow_DOM#Attaching_the_shadow_DOM_to_the_shadow_root).  They cannot by inlined as `<style>` tags in `<head>`, and need to be present as JS strings during component registration, so a standard webpack config which also uses `style-loader` will cause these imports to fail cryptically.

packages/perspective-jupyterlab/src/config/plugin.config.js
```diff
@@ -15,7 +15,7 @@ module.exports = {
     entry: "./src/ts/index.ts",
     resolveLoader: {
         alias: {
```
texodus: The inclination to resist scope creep is *much* appreciated - but if we're going to move these files at all, maybe we just move them to where they belong or leave them in `src/`?

packages/perspective-jupyterlab/src/config/plugin.config.js
```diff
@@ -25,11 +25,45 @@ module.exports = {
...
+                        options: {
+                            level: 2
+                        }
+                    },
+                    {
+                        loader: "less-loader"
+                    }
+                ]
+            },
+            {
```
texodus: Likewise, removing this will not mix well with traditional use of `html-loader` - these are for strings in the shadow DOM.

packages/perspective/webpack-plugin.js
```diff
@@ -7,84 +7,61 @@
...
 
-const BABEL_CONFIG = require("./babel.config.js");
+const include = [__dirname];
 
 class PerspectiveWebpackPlugin {
     constructor(options = {}) {
         this.options = options;
     }
 
     apply(compiler) {
```
texodus: This refactored plugin does not work - it fails to build the [webpack example project in `examples/`](https://github.com/jpmorganchase/perspective/tree/master/examples/webpack).

packages/perspective-jupyterlab/src/config/plugin.config.js
```diff
@@ -25,11 +25,45 @@ module.exports = {
...
+                test: /\.(html)$/,
+                loader: "html-loader"
+            },
+            {
+                test: /\.(arrow)$/,
+                loader: "arraybuffer-loader"
+            },
+            {
```
texodus: The Jupyterlab plugin does not even use babel - it is written in Typescript.  Is it not a red flag that this refactor requires so much extra webpack configuration for e.g. this package?

packages/perspective/webpack-plugin.js
```diff
@@ -7,84 +7,61 @@
...
 
-        rules.push({
-            test: /\.js$/,
-            include: load_path,
-            exclude: /node_modules[/\\](?!\@jpmorganchase)|psp\.(asmjs|async|sync)\.js|perspective\.(asmjs|wasm)\.worker\.js/,
-            loader: "babel-loader",
-            options: BABEL_CONFIG
-        });
-
-        rules.push({
```
texodus: I think removing this breaks the `"inline"` plugin flag, which needs to work in concert with this specific loader, as in the Jupyter plugin, but there are other reasons to use it - e.g. when we have webpack 4 support where [wasm loading from emscripten modules is completely ****ing broken](https://github.com/webpack/webpack/issues/7352).

packages/perspective-jupyterlab/src/config/plugin.config.js
LukeSheard: > We should not remove the less loaders here. 

Is this a typo? Can you clarify what you think the code should be here is?

packages/perspective-jupyterlab/src/config/plugin.config.js
LukeSheard: Ideally I'd like to be in a place we don't publish `/src` at the end of this PR, so I'm open to either 
1. Leave them where they are currently
2. Move them into `webpack-plugin` package 

Considering we're chatting about this, maybe it's best to move them into the `webpack-plugin` package?

packages/perspective/webpack-plugin.js
LukeSheard: Fixed in latest comment.

packages/perspective/webpack-plugin.js
```diff
@@ -7,84 +7,61 @@
...
-            test: /psp\.(sync|async)\.wasm\.js$/,
-            include: load_path,
-            use: {loader: WASM_LOADER_PATH, options: {name: "[name]"}}
-        });
```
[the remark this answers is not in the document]
LukeSheard: Good spot, we can come back to webpack 4 in a later PR.

packages/perspective-jupyterlab/src/config/plugin.config.js
LukeSheard: I think this is no longer needed with my changes to `perspective-viewer` and the plugins to publish JS files only.

LukeSheard: I was correct, these are no longer needed in this package so I've removed them.

packages/perspective-jupyterlab/src/config/plugin.config.js
LukeSheard: Yup. Removed.

packages/perspective-jupyterlab/src/config/plugin.config.js
LukeSheard: In doing some more work I've decided to just go with option 2 here.

LukeSheard:
@texodus Updated the PR with all the comments we've discussed in person and on here.

LukeSheard:
To reduce the scope creep of this PR I've raised #390, #391 and #392 and #393 which fix a lot of the initial issues I had with getting the `webpack-plugin` to work as expected. 

To avoid conflicts once those are merged I can raise the final `webpack-plugin` PR.

PR #381, opened 2019-01-14
Interoperability between CMAKE_BUILD_TYPE and PSP_DEBUG

timkpaine:
`CMAKE_BUILD_TYPE` is the standard, and is configured in `DEBUG` or `RELEASE`. We use `PSP_DEBUG` in the C++ code to trigger debug statements, and rely on it in `cmake` to build with debug symbols and without optimizations. This PR lets `PSP_DEBUG` be set from either the `PSP_DEBUG` environment variable, or from the `CMAKE_BUILD_TYPE` flag for cmake.


texodus:
Thanks for the PR - this is a great improvement to the build process!

PR #382, opened 2019-01-14
[WIP] Move webpack configuration to project root

ColinEberhardt:
As per the [discussion in this issue](https://github.com/jpmorganchase/perspective/issues/376#issuecomment-453911712) the focus workspaces features doesn't work at the moment because webpack configuration is part of the `@jpmorganchane\perspective` package.

I think this configuration should be moved to the root of the project.

This PR illustrates the change for perspective-viewer only (hence it is a Work In Progress). I've simply taken the various webpack files and pasted them into `scripts\webpack`

With these changes, the following now works:

```
$ cd packages/perspective-viewer
$ yarn install --focus
$ yarn run build
```

If you're happy with this as an approach, I'll clean up and roll it out across the project.


LukeSheard:
@ColinEberhardt The webpack configs need to remain part of a package that's exported so that users can consume the API through webpack. Take a look at #378 which has a base for breaking these things out into a shared directory, we can make a private development package which is symlinked in but not published.

ColinEberhardt:
> @ColinEberhardt The webpack configs need to remain part of a package that's exported so that users can consume the API through webpack. 

I must admit, I'm not entirely sure what you mean! The APIs the Perspective currently exposes are the `view` and `table` via `perspective`, and the various 'viewer' WebComponents - are these the APIs you'd expect people to consume via Webpack?

texodus:
@LukeSheard is correct - Perspective requires a [plugin to be used via webpack](https://github.com/jpmorganchase/perspective/blob/master/packages/perspective/webpack-plugin.js) due to a few features which are tricky to configure correctly without carnal knowledge of the build, namely:
* Feature detecting WASM and loading a separate WebWorker at runtime precludes use of `worker-loader`
* CORS restrictions on WebWorkers means we need to XHR the source when a cross-origin load is detected (again at runtime), and the Worker needs to be loaded [via Blob](https://developer.mozilla.org/en-US/docs/Web/API/Web_Workers_API/Using_web_workers#Embedded_workers).
* As a speed optimization - the workers take *ages* to compile, and can't be extended in user code, so these are pre-compiled and copied to your webpack output path.
* asm.js in particular cannot be modified by babel/uglify/etc source-to-source compilers, as none are aware of `"almost asmjs"` or `"use asmjs"` directives.
* Shadow DOM requires some non-traditional usage of CSS/HTML assets that conflict with standard webpack configs, yielding confusing errors.

The implementation is based off [Microsoft's Monaco webpack plugin](https://github.com/Microsoft/monaco-editor-webpack-plugin), which has complications.  All this aside though - I am unclear why the webpack configuration is the crux of your issues with `--focus`.  It seems this is more an issue of the way the plugin resolves paths than the specific path of the config itself.

As I don't believe this approach is going to work, I'm going to close this PR.  If you feel this is in error, feel free to re-open or come over to the gitter channel where we can discuss.

PR #387, opened 2019-01-18
Empty filters

Ro4052:
#349 #375 Prior to my pull request for invalid filters, I've had a slight re-think of my implementation. My implementation resulted in empty strings being invalid filter operands. Although this makes sense for the viewer component, it doesn't seem to be appropriate for those using Perspective without the viewer. Therefore, I've pulled the check for empty strings into the viewer and updated the tests as necessary.


texodus:
Thanks for the PR!

While this is most certainly more correct, the issue is a bit complex - users _should_ be able to do anything that developers can do wrt data transforms, eg filter on datasets with empty string values - and `null` values for that matter, which perspective uses for both invalid aggregates as well as missing values.  Like I said in the previous PR, I think this functionality ultimately needs a bit of a re-think, both in terms of its API and user experience - in the interim, any imporvement is warranted!

PR #389, opened 2019-01-19
[WIP] Merging JS and Python binding code - Part 2

timkpaine:


texodus:
Thanks for the PR!  This is a great step towards unifying these APIs - I have some anxiety about exposing the C++ types directly, and this helps alot as @sc1f continues to remove this functionality on the JS side, making it easy to extend it on the Python side.

PR #390, opened 2019-01-20
Remove cross /src imports in packages

LukeSheard:
Remove the cross imported files from `/src` directories in perspective packages. Useful so that we can build self contained bundles without needing to reach into packages. 

1. Added a comment for usages of `COLUMN_SEPARATOR_STRING` and replaced with constant. 
2. Replaced `detectIE` with external `detectie` package.


texodus:
Thanks for the PR! Tested in IE and works great - I added a small cleanup to remove the commented out code and update `yarn.lock`

PR #391, opened 2019-01-21
Output CJS module bundles for perspective-viewer packages

LukeSheard:
Output a `slim` version of the perspective viewer packages which is targeted when consuming the package from a module bundler (e.g. webpack, rollup, etc.)

1. Reduce the scope creep of #378 
2. unminified, precompiled assets, will be quicker for Webpack to parse and compile than the UMD builds. 
3. It will prevent duplication of dependencies of `perspective-*` packages when built together - leading to smaller bundle sizes for consumers.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -121,7 +121,7 @@ website/static/css/material.dark.css
 
 # CPP Compile
 /src/include/boost
+cjs
 obj
-packages/perspective/es5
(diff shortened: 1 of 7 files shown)


texodus:
Thanks for the PR!  Discussed offline, this is good to merge plus some future additions

PR #392, opened 2019-01-21
Output CJS files perspective package

LukeSheard:
Converts the internals of the `webpack-plugin` to use `precompiled` CJS assets generated by Babel. 

Has the same benefits as #391.


texodus:

PR #395, opened 2019-01-23
Benchmark limit + minor changes to DataAccessor

sc1f:
1. Benchmark script now has a `--limit <NUMBER>` flag, where `<NUMBER>` is the number of previous releases that will be run. For example, `--limit 1` will bench only the master, while `--limit 2` benches master and the preceding release.

2. Minor changes to the C++ side of DataAccessor to fit `const&` semantics for function parameters that are not mutated.

### DIFF
--- a/docs/development.md
+++ b/docs/development.md
@@ -152,9 +152,14 @@ with your changes to preserve them for future comparison.
 yarn bench
 ```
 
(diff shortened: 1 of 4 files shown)


texodus:

PR #396, opened 2019-01-23
C++ Date Validation

sc1f:
1. Add a date parser on the C++ side that correctly validates date/datetime strings using the `iomanip` header. 
  - *The parser is not called by any running code, nor is it included in `CMakeLists.txt`.* 
  - The header does not include support for milliseconds and timezones, which is fine for string validation but will be problematic for parsing.
2. Added tests for the date parser, and moved computed column tests into their own `describe` block.
3. Fixed the use of now-undefined `PSP_ABORT` in `PSP_VERBOSE_ASSERT2`.


texodus:
Thanks for the PR!

PR #398, opened 2019-01-25
Split webpack plugin into a separate

LukeSheard:
_There's a few too many commits in this PR - will squash this weekend._ 

What does this do?
- Moves the loader code into a separate package so `@jpmorganchase/perspective` does not have a dependency on build tools. 
- Moves all build tools into `devDependencies`
- Moves loaders into their appropriate module webpack config for better clearer declaration over which packages need which modules.


packages/perspective/src/js/perspective.wasm.js
```diff
@@ -7,15 +7,26 @@
...
+            },
+            print(message) {
+                console.log(message);
+            },
+            locateFile(path) {
+                if (path.endsWith(".wasm")) {
+                    return wasm_path;
+                }
+                return path;
+            }
```
texodus: This breaks asm.js - does not inline the `mem` file

packages/perspective/src/js/perspective.asmjs.js
```diff
@@ -7,7 +7,7 @@
  *
  */
 
-require("babel-polyfill");
+require("@babel/polyfill");
```
texodus: This causes conflict warning with `useBuiltins`

packages/perspective/src/js/perspective.parallel.js
```diff
@@ -14,7 +14,7 @@ import {worker} from "./api.js";
 import asmjs_worker from "./perspective.asmjs.js";
 import wasm_worker from "./perspective.wasm.js";
 
-import wasm from "./psp.async.wasm.js";
+import wasm_path from "./psp.async.wasm.js";
```
texodus: Because this was also added to `perspective.wasm.js` above, this asset is inlined twice in the jupyterlab plugin

packages/perspective-jupyterlab/src/config/plugin.config.js
```diff
@@ -15,7 +15,7 @@ module.exports = {
     entry: "./src/ts/index.ts",
     resolveLoader: {
         alias: {
-            "file-worker-loader": "@jpmorganchase/perspective/src/loader/file_worker_loader.js"
+            "file-worker-loader": "@jpmorganchase/perspective-webpack-plugin/lib/file-worker-loader.js"
```
texodus: Please keep these packages consistent `lib` vs `src` - there's no reason to re-culture every package we add.

packages/perspective-jupyterlab/package.json
```diff
@@ -38,6 +38,8 @@
         "@phosphor/widgets": "^1.5.0"
     },
     "devDependencies": {
+        "babel-plugin-transform-custom-element-classes": "^0.1.0",
```
texodus: Same - why is this here?

packages/perspective-jupyterlab/babel.config.js
```diff
@@ -0,0 +1,22 @@
...
+        ["@babel/plugin-proposal-decorators", {"legacy": true}],
+        "transform-custom-element-classes",
+        [
+            "@babel/plugin-transform-for-of",
+            {
+                "loose": true
+            }
+        ]
+    ]
+}
```
texodus: This package does not use Babel - why do we need this?

packages/perspective/src/js/perspective.parallel.js
```diff
@@ -67,20 +72,31 @@ class WebWorker extends worker {
...
+            const [worker, buffer] = await Promise.all([override.worker(), override.wasm()]);
+            const msg = {
+                cmd: "init"
+            };
+            for (var key in this._worker) {
+                worker[key] = this._worker[key];
+            }
+            this._worker = worker;
+            this._worker.addEventListener("message", this._handle.bind(this));
+            this._worker.postMessage(msg, [buffer]);
```
texodus: This is [not how transferables work](https://developer.mozilla.org/en-US/docs/Web/API/Transferable) - there is even test coverage for this feature!

packages/perspective/src/js/perspective.wasm.js
```diff
@@ -7,15 +7,26 @@
  *
  */
 
-const load_perspective = require("../../obj/psp.async.js").load_perspective;
-const perspective = require("./perspective.js").default;
+import {load_perspective} from "../../obj/psp.async.js";
+import wasm_path from "./psp.async.wasm.js";
```
texodus: See above - these are stub files that the loader rewrites, they cannot be moved

texodus:
There are multiple issues with this patch
* It [doesn't compile](https://travis-ci.org/jpmorganchase/perspective/builds/484528093?utm_source=github_status&utm_medium=notification)
* Adding the missing dependencies - the test suite fails.
* Transferable support is broken in wasm loading, and detection no longer works
* IE/asmjs loading is broken
* cross-origin CDN builds no longer work (paths resolve to local domain).
* Jupyterlab plugin no longer builds within Jupyter - and the asset has doubled in size due to asset inlining.
* Please see also the [contributing guidelines](https://github.com/jpmorganchase/perspective/blob/master/CONTRIBUTING.md), specifically the parts about single commits, merge commits, reorganizing packages, "revert" commits, tests and docs.

More importantly, this does not resolve the central circular-dependency issue with moving the plugin out of the perspective package.  I don't think further commits to this branch are going to make it mergeable, so I'm going to close this PR for now.  As a next step, may I recommend focusing on just 1 change at a time, such as moving the existing loader to a new package, and validating that the change both compiles and tests clean?

PR #401, opened 2019-01-29
adding iex streaming example to readme

timkpaine:
Adds an example based on IEX's new IEX Cloud streaming api to readme. Utilizes blocks and server-side-events.


texodus:
This is cool, thanks for the PR!

PR #403, opened 2019-01-31
Add publicPath support to webpack-plugin 

LukeSheard:
Support webpack's `publicPath` option and add long term caching options into the webpack plugin. 

Changes: 
- 42dec1d - update webpack example to be a cross origin asset application
- 84d8536 - renames the loader filenames for clarity of their real intention
- fac6af1 - fixes the bug with locating the worker and wasm assets paths when loaded with a `publicPath`
- 2e1adfd - adds support for long term caching into the output of the perspective loaders. 
- bc3585c - updates the webpack example to utilize the long term caching webpack options


packages/perspective-webpack-plugin/src/js/psp-worker-loader.js
```diff
@@ -55,17 +54,15 @@ exports.default = function loader(content) {
                 .replace(/(cjs\/js)/, "build");
         }
         content = fs.readFileSync(inputPath).toString();
-        if (!options.inline) {
```
texodus: This causes duplicate hashed webworker assets to be output in the Jupyterlab plugin on every build

screen shot 2019-02-01 at 10 08 20 pm

texodus:
Thanks for the PR!

Thoroughly tested, fixed one minor regression, but everything else looks good!

PR #413, opened 2019-02-04
Port View to C++

sc1f:
As outlined in `API.md`, we are porting functionality previously limited to `perspective.js` into C++, the largest chunks of which involve the `view` and `table` classes. 

This PR begins the porting process for view.

Changes:

- Add `View.h`, which defines and implements a templated class of View that is dependent on the class of the underlying context.
- Inside `View.h`, methods from the JS version of view are ported and implemented, including `schema`, `_column_names`, and various methods related to pivot table operations.
- Add `translation.js`, which begins to implement methods related to bridging the gap between C++ and JS, in this case vector + map extraction functions.
- Add `make_view` function inside `emscripten.cpp`, which returns a shared pointer to the newly created View.
- Add `_View` in the instantiation of the Javascript view, which constructs a View object in C++ and maintains an internal reference, allowing us to call the methods on that object. The internal `_View` is deleted when the JS view is deleted. 

In terms of design direction, `API.md` outlines the long-term outcome of the refactor. I'd love any and all comments on what's there as the design of the code in the PR.


src/include/perspective/View.h
```diff
@@ -0,0 +1,339 @@
...
+}
+
+template <typename CTX_T>
+t_index
+View<CTX_T>::expand(std::int32_t idx) {
+    return m_ctx->open(idx);
+}
+
+template<>
+t_index
```
timkpaine: inline or move to cpp

src/include/perspective/View.h
```diff
@@ -0,0 +1,339 @@
...
+}
+
+template <typename CTX_T>
+t_index
+View<CTX_T>::collapse(std::int32_t idx) {
+    return m_ctx->close(idx);
+}
+
+template<>
+t_index
```
timkpaine: inline or move to cpp

src/include/perspective/View.h
```diff
@@ -0,0 +1,339 @@
...
+        m_ctx->set_depth(depth);
+    } else {
+        std::cout << "Cannot expand past " 
+        << std::to_string(row_pivot_length) 
+        << std::endl;
+    }
+}
+
+template<>
+void
```
timkpaine: inline or move to cpp

src/include/perspective/View.h
```diff
@@ -0,0 +1,339 @@
...
+ * The schema of this View. Output and logic is as the above
+ * schema(), but this version is specialized for zero-sided
+ * contexts.
+ * 
+ * Returns
+ * -------
+ * std::map<std::string, std::string> schema of the View
+ */
+template <>
+std::map<std::string, std::string>
```
timkpaine: inline or move to cpp

src/include/perspective/View.h
```diff
@@ -0,0 +1,339 @@
...
+/**
+ * The column names of the View. Same as above but 
+ * specialized for zero-sided contexts.
+ * 
+ * Returns
+ * -------
+ * std::vector<std::string> containing all column names
+ */
+template<>
+std::vector<std::string>
```
timkpaine: inline or move to cpp

timkpaine:
I would prefer the class be lowercase too but we're straddling C++/JS/Python(soon) conventions

texodus:
Thanks for the PR!  This is a great start to making the dynamic language bindings portable.

PR #426, opened 2019-02-11
C++ subproject

texodus:
This PR moves the C++ files from the project root to the `cpp/` directory, such that the project repo may be shared by multiple C++ projects in the future.  It also moves the `clean`, `lint` and `fix` C++ tasks to node.js scripts in the `scripts/` folder, such that they may be easily parameterized by project.


timkpaine:
@texodus clang-format is not consistent, mine says this is a problem but if I fix it then the test says its a problem 
screen shot 2019-02-14 at 10 16 12 am

PR #427, opened 2019-02-11
Tidy up dependencies

LukeSheard:
Tidy up a couple of unneeded dependencies to reduce install load on users:

- migrate from `underscore` to the more common `lodash`
- Remove `less` as a dependency from `@jpmorganchase/perspective`


texodus:
Thanks for the PR!  Changes looks good

PR #428, opened 2019-02-11
Allow mac failures on travis, have travis fail fast

timkpaine:
This PR does two things:
- the Mac image on travis hangs occasionally. Since several of us are developing full time on MacBooks anyway, we can allow this one to fail. 
- Makes travis fail fast, so we dont waste resources when people need to iterate on PRs.


texodus:
Thanks for the PR!   Looks good

PR #431, opened 2019-02-13
Add `replace()` and `clear()` to table

texodus:
Adds a `replace()` and `clear()` method to the `perspective.table()` API, as well as symmetric methods to the `<perspective-viewer>` web component.


texodus:
Thanks for the PR!

PR #432, opened 2019-02-14
Remove inlined webpack loader declaration

LukeSheard:
Ensure that the webpack loader declarations are in the plugin and configurations so that custom resolution isn't affected when consuming the plugin.


texodus:
Thanks for the PR!

As this does not currently compile, I'm going to close this - please re-open when it is ready for review

PR #434, opened 2019-02-18
Make sort syntax consistent in tests

sc1f:
Previously we allowed for 2 types of sort syntax to be passed through configs:

- `[["column", "asc"]]`: a 2D array with a column name and a direction string
- `["column"]`: an array of column names

This PR deprecates the latter format in our tests and uses the clearer, first format.


texodus:
Thanks for the PR!  This was a silly leftover vestige from a by-gone era, before we had 9 different sort types - past time it was cleaned up!

PR #437, opened 2019-02-19
Purges phosphor and jlab tests 

timkpaine:
As discussed, this removes the (very minimally implemented) javascript tests from phosphor and jupyterlab sub packages to avoid any nonsense with custom jest configs.


texodus:
Thanks for the PR!

PR #445, opened 2019-02-22
Fix regression on view expand

sc1f:
- View should no longer be able to expand for indices past the number of row pivots.
- Add tests to confirm expand works as intend
- Refactor view.cpp to explicitly specialize `expand()`, `collapse()`, and `set_depth()`


texodus:
Thanks for the PR!

PR #447, opened 2019-02-23
Fix for date/datetime category axis labels

neilslinger:
Values for category axis are now strings, this causes the Date conversion to fail. Casting to Number first.


texodus:
Thanks for the PR!  This has been fixed and tested already in #448

PR #449, opened 2019-02-25
Add tests for filtering datetime columns by string

sc1f:
- Verify that we can use correctly-formed date strings to filter datetime columns.
- Add core engine + plugin tests to prevent regression.


texodus:
Thanks for the PR!  Tests are always welcome.

Issue #442, opened 2019-02-20
perspective-jupyterlab not publishing to dist package


timkpaine:
Was added here: https://github.com/jpmorganchase/perspective/commit/fe689257923a955221e5332f0130ef07251a6d88#diff-9d8f23b3f4bb47ef250d6f6a116df908

Then you re-broke it here:
https://github.com/jpmorganchase/perspective/commit/627ac4829d056d0932cae5d3586430d494484624#diff-9d8f23b3f4bb47ef250d6f6a116df908

texodus:
Thanks for the report!  This has been "fixed" and was disted in 0.2.19, then superseded by #467

PR #450, opened 2019-02-25
Moved `perspective-jupyterlab` build from `build/` to `dist/` to work…

texodus:
… around hardcoded resolution path


texodus:
Fixes #442

PR #452, opened 2019-02-25
Port view to C++

sc1f:
This PR ports the view class to C++ in order to enhance portability and allow for more flexible language bindings.

_JS Changes_
- `table.prototype.view` refactored to pass the config into C++
- Checks for `psp_okey` for column-only views now use `view.column_only`
- Test in `internal.js` has been marked skip due to change in how we handle incorrect arity aggregates - **reenable**?

_C++ changes_
- `View` class holds aggregates, pivots, filters, sorts, and a `t_config` object
- `t_config` has new constructors and members oriented towards view creation
- `make_context` is no longer exposed to emscripten
- `get_aggspecs`, `get_fterms`, `get_sortspec` now handle the parsing of config options
- added utility to_string/from_string methods for aggregate and filter enums in `base.h`

Python tests do not pass at the moment because function bindings in `binding.h` have been changed. I'm starting the PR as a place where we can review + talk about these changes in the codebase.


timkpaine:
the python binding needs a lot of work before it can be fully usable, in the mean time so long as the tests pass feel free to comment out as needed and i'll reimplement later

cpp/perspective/src/cpp/base.cpp
```diff
@@ -280,6 +312,137 @@ filter_op_to_str(t_filter_op op) {
     return "";
 }
 
+t_filter_op
+str_to_filter_op(std::string str) {
+    if (str == "<") {
```
timkpaine: don't like switch statements?

timkpaine:
@sc1f  @texodus good to go, we should consider adding some more CPP linting. I also think @sc1f should add C++ tests for these blocks (even if they're still being tested from JS), so we have a good sense for how to use them when it comes time to extend the other bindings.

cpp/perspective/src/cpp/base.cpp
sc1f: You can’t switch on a string in C++, so it was this or write another enum that gets mapped to the original enum.

timkpaine: you can do some [stuff](http://www.rioki.org/2016/03/31/cpp-switch-string.html), but I think what you've done here is more readable and fine.

texodus:
Thanks for the PR!  This looks great and really improves the JS and C++ API code.  I've added a few additional cleanup bits and gone ahead and removed the `EMSCRIPTEN_BINDINGS` code this PR deprecates - which by itself cuts almost 20k off the WASM asset.

@timkpaine Thanks for the review.  Fully agreed on all fronts - we'll definitely need to address these.

PR #458, opened 2019-02-27
Integration tests for D3FC

texodus:
This PR adds simple smoke/integration tests to `perspective-viewer-d3fc`, using the same puppeteer/record matching approach used by `perspective-viewer` and friends, details of which can be [found in the perspective documentation](https://jpmorganchase.github.io/perspective/docs/development.html#testing).  These tests are designed to prevent us from merging changes which "obviously" break the plugin, rather than provide fine-grained feedback as to what failed;  as such, they are really meant to be run mostly by travis, and before opening a PR, rather than iteratively during development.

The tests also assert that the Chrome console logs no errors while running - and one fix for tooltips was needed to achieve this, checking that the SVG tree itself is still attached to the DOM.


texodus:
Discussed offline

PR #466, opened 2019-03-04
Bug/stop console log in area chart and hide y axis

redbearsam:
Super minor fixes. But I didn't want to shoehorn them into some unrelated feature branch, so here they are.


redbearsam:
Oops, meant to hit fork!

PR #473, opened 2019-03-06
Add t_data_slice for data output

sc1f:
This PR begins to abstract the process of getting data out of perspective for use in visualizations/serialization. Right now, we create an intermediate array inside `emscripten.cpp` that contains the data extracted from the context. The end goal of the PR is to eliminate that intermediate structure—and the logic surrounding it—so as to make the data output API portable and trivial to implement in binding languages.

`t_data_slice` is a class that contains the underlying slice of data and associated metadata, which allows us to eventually remove the intermediate logic of `get_data` in `emscripten.cpp`. 

_Changes_
- Adds `t_data_slice` class
- Adds `is_column_only` to context construction
- Moves `get_column_extent` to independent class

Making the PR so we can discuss and get these changes into mainline before the branch gets too big; I've tried to benchmark these changes but the script seems to be broken on my machine and isn't benching properly.

### DIFF
--- a/cpp/perspective/CMakeLists.txt
+++ b/cpp/perspective/CMakeLists.txt
@@ -285,6 +285,8 @@ set (SOURCE_FILES
 	src/cpp/context_two.cpp
 	src/cpp/context_zero.cpp
 	src/cpp/custom_column.cpp
+	src/cpp/data_slice.cpp
+	src/cpp/date.cpp
 	src/cpp/data.cpp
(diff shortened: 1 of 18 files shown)


texodus:
Thanks for the PR!

This looks great!  I added one small fix for the windows build due to missing `std::min` on Windows.  Benchmarks show a ~10% improvement in JSON generation, which is quite a nice bonus!

PR #474, opened 2019-03-07
D3FC fixes

texodus:
This commit enabled D3FC integration tests in Travis and the default test runner, and fixes a few minor issues that the test caught -
* [Fixes](https://github.com/jpmorganchase/perspective/pull/474/files#diff-6c0e85d60c2dc287691ec61e66dad322R13) `NaN` sizing bug, from [this commit](205042d510256f771bde36744a968a6507f52919) which had modified the signature of `domain()` in `mainAxis.js`, but not for the size property of scatter charts.
* Applies `withOutOpacity()` modification to line charts, which had become hard to see with the introduction of chart opacity.


texodus:
@DevAndyLee Thanks for the review!

PR #478, opened 2019-03-07
Date filter issue

JHawk:


packages/perspective/test/js/filters.js
```diff
@@ -110,6 +110,43 @@ module.exports = perspective => {
...
                 table.delete();
             });
+
+            describe("filtering on date column", function() {
+                const schema = {
+                    w: "date"
+                };
+
+                const date_results = [
+                    {w: +new Date(1535760000000)}, // Fri Aug 31 2018 20:00:00 GMT-0400
```
JHawk: Setting the schema converts the datetime above to a date in the result set

texodus:
Thanks for the PR!

PR #479, opened 2019-03-08
d3fc plugin - Label rotation, interpolated legend, zooming, scatter chart things

redbearsam:
Scatter chart things:

- Shapes for points instead of simple circles when using splits or multi-serie.
- yScatter chart implemented.
- x/yScatter using canvas (currently in addition to x/yScatter using svg). This will likely be stripped out later as the performance on the canvas version is better, particularly evidently when zooming. For the time being however, tooltips are not functional on this version, so we're awaiting that feature's completion.

NB: The tooltip behaviour on various scatters and lines is generally unideal. We are currently working on an approach more similar to highcharts where it is not necessary to hover over the points themselves - where the tooltip appears for the closest point.


texodus:
Thanks for the PR!

Exceptional contribution as per norm!

PR #481, opened 2019-03-09
Get version from package json: fixes #423

### LINKED ISSUE #423: Pull JupyterLab widget version from package.json
https://github.com/jpmorganchase/perspective/blob/master/packages/perspective-jupyterlab/src/ts/version.ts#L11

should come from 

https://github.com/jpmorganchase/perspective/blob/master/packages/perspective-jupyterlab/package.json#L3

timkpaine:


texodus:
Thanks for the PR!

PR #484, opened 2019-03-14
D3fc plugin - "Nearby" tooltips for some charts

matt-hooper:
- Tooltips for line charts, area charts and canvas based XY scatter using nearby point logic
- Added nearby point logic to Y scatter
- Removed SVG XY scatter chart. Canvas XY scatter is now the default
- Additional refactors and bug-fixes


texodus:
Thanks for the PR!  Looks great!

PR #488, opened 2019-03-18
D3fc plugin - Basic OHLC and Candlestick charts

matt-hooper:
Added D3FC Candlestick and Open, High, Low, Close (OHLC) charts and added a new example to the Simple examples to demonstrate.

image

image


timkpaine:
I can finally kill my highcharts candlesticks branch!

texodus:
Thanks for the PR!  This has been on the top of the feature request list for a long time!

It would be great if in the future, Perspective could calculate time buckets like this so we could create these chart types from arbitrary data.

PR #517, opened 2019-04-03
Add row delta to View

sc1f:
When calling `view.on_update(callback, {mode: pkey})`, Perspective now returns a sorted array of row indices which were updated in the previous update operation.

This PR includes an addition to the benchmark suite for `on_update` calls, as well as tests for each use case of the row delta code.


texodus:
Thanks for the PR!  Great new feature, and the benchmarking is quite informative as well!

Issue #522, opened 2019-04-05
Column-only views in Hypergrid display an extra empty row


sc1f:
<img width="727" alt="Screen Shot 2019-04-04 at 4 15 40 PM" src="https://user-images.githubusercontent.com/13220267/55600531-02439980-5722-11e9-8861-ccbed8c4f9f8.png">

In column-only views, Hypergrid displays an extra empty row whereas `to_json` and highcharts displays the same view correctly.

texodus:
Thanks for the report - this is fixed and will be in the 0.2.22 release.

PR #523, opened 2019-04-05
Fix issue #522: row count is correct on column-only views

sc1f:
This PR fixes the persistent issue of an extra empty row on column-only views—the `num_rows` method is edited in the C++ to ignore the aggregate row for column-only views.

- Regression tests are added in the engine as well as in the Hypergrid plugin.


texodus:
Thanks for the PR!  Good catch and good fix - this problem has apparently been lying around since the inception of the column-only pivot feature.  

Great tests is much appreciated as well!

PR #540, opened 2019-04-11
Remove header rows for column-only views

sc1f:
Previously, the engine emitted a header row—containing all the data—for column-only views, which we sliced out in JS. This PR creates offsets inside `View` and `t_data_slice` in order to allow us to skip these rows in the engine.


texodus:

PR #544, opened 2019-04-15
Called save and restore with the correct context

Ro4052:
To match the other plugin calls, I believe these methods should also use the chart node as the context.


texodus:

Issue #510, opened 2019-04-02
feature request : add support to cross filtering


djouallah:
I have being playing with perspective only since yesterday, the ease of use and performance is amazing !! thanks for your work.

I use currently vega-lite,  which support the current features, I am not sure if perspective support that or not.

cross filtering              : is it possible to click on a chart on one view and it will filter the others view ?
complex aggregation :  for example sum(field1)/total(field1)

texodus:
Cross filtering has been supported for a few versions - there's [bl.ocks example](https://bl.ocks.org/JHawk/9b20383f042853f27c6e720baf4a19db) as well as a [local examples in `/examples`](https://github.com/jpmorganchase/perspective/blob/master/examples/simple/linked.html).  Brief documentation on how these events work [is here](https://github.com/jpmorganchase/perspective/blob/master/docs/md/usage.md#click-events).

As for complex aggregation, we only support hard-coded aggregates for now, as calculating these in Javascript code via Perspective's C++ will almost certainly be slower than just calling `Array.prototype.map()` due to the marshalling cost;  however, `sum(field1)/total(field1)` already exists as `"average"`.  If there are some other general purpose aggregates you think would be useful, feel free to open a new issue and we can add them!

djouallah:
that's amazing, thanks for the examples

fibinger:
May I know what was the reason to stop supporting it?

PR #545, opened 2019-04-15
Link click event example in README

JHawk:


texodus:
Thanks for the PR!  Looks good, and is an example for #510

PR #548, opened 2019-04-20
Remove 'aggregate' syntax in engine

sc1f:
This PR removes the old `aggregate` syntax in `perspective.js` and the C++ core, making the API symmetrical to the plugin API, which uses `aggregates` and `columns`.

- Views are constructed using the `aggregates` and `columns` syntax
- Passing in `aggregate` throws an error within `table.prototype.view`
- Weighted mean aggregate parsing is removed with the new syntax
- Change highcharts to use `aggregates` syntax


sc1f:
Update: benchmark suite still uses old “aggregate” syntax, and I want to fix that before this goes into master.

PR #552, opened 2019-04-23
Remove 'aggregate' syntax in engine

sc1f:
This PR removes the old aggregate syntax in perspective.js and the C++ core, making the API symmetrical to the plugin API, which uses aggregates and columns.

- Views are constructed using the `aggregates` and `columns` syntax
- Backwards compatibility is maintained for `aggregate` 
- Weighted mean aggregate parsing is removed with the new syntax
- Change highcharts to use aggregates syntax


texodus:
Change looks good, thanks for the PR!

PR #553, opened 2019-04-25
Add 'styleElement' method to tell perspective-viewer to restyle picking up css changes

zemeolotu:
Only implemented for hypergrid and highcharts (d3fc implementation coming soon)


texodus:
Thanks for the PR!  Changes look good!

PR #561, opened 2019-05-02
Correctly read and generate boolean values for Arrow format

sc1f:
This PR allows the engine to parse and generate booleans correctly for the Apache Arrow format.

- `to_arrow()` correctly generates booleans using a bitmap
- tests are added for boolean reading + arrow updates

Partial update tests are skipped due to a filesystem read error that originates from within the test files.


texodus:
Thanks for the PR!  This looks great - and great test coverage too!

PR #563, opened 2019-05-07
D3 Treemaps

matt-hooper:
- Implementation of d3 based Treemaps:

image

- Zoom to fixed date range
image

- Implementation of Save/Restore functionality to allow Legend size and position to be saved.


texodus:
Thanks for the PR!

PR #571, opened 2019-05-09
Refactor Javascript API +add documentation

sc1f:
This PR breaks up `API.js` into separate files inside the `API` folder, adding documentation for each part of the API as well as the host/worker architecture. The `Host` class is moved from `perspective.js` into the `API` folder, and additional documentation has been added for clarity around how the `Worker` and `Host` classes function & extend.


texodus:

PR #572, opened 2019-05-13
Feature/treemap enhancements

redbearsam:
Enhancements:

1. Legend moved to a gutter down the right-hand side so it does not overlap any charts.
2. Text can be dynamically resized to be smaller where necessary to ensure display of as many labels as reasonably feasible.
3. Text specific transitions have been introduced to prevent jarring shifts between treemap layers. Once the text location and size has been calculated the first time, the map of this is retained so the slightly slower and more compute intensive transition does not have to be used again when returning to a level.
4. Tooltips cannot overflow off the chart on the top and left edges.


texodus:

Issue #562, opened 2019-05-07
[D3FC] - Charts do not handle large numerical values on the Y-axis


matt-hooper:
### Steps to Reproduce:
 1. Include a dataset that contains very large numerical values (e.g. 100,000,000+)
 2. Select a D3FC chart (e.g. d3fc Y Line).
 3. Add the large value to the chart

### Expected Result:
The values should be displayed and visible on the Y-axis

### Actual Result:
Only the last 8 digits of the numbers are displayed:

**D3FC**
![image](https://user-images.githubusercontent.com/33057010/57293545-f0f10280-70bc-11e9-8c3a-a683fa10f271.png)

### Highcharts equivalent
In the equivalent Highcharts implementation, large numbers are formatted as 100M (for million), this prevents the labels getting too big:

**Highcharts**
![image](https://user-images.githubusercontent.com/33057010/57293647-3c0b1580-70bd-11e9-98b6-6c504fa867b2.png)

PR #573, opened 2019-05-13
Fix the format of long numbers

matt-hooper:
#562


texodus:
Thanks for the PR!

Great improvement to large range readability!

PR #577, opened 2019-05-14
Get d3fc css directly from the package

DevAndyLee:
Instead of copying it from the *head*. The css of both the "element" and "chart" components are required.

### DIFF
--- a/packages/perspective-viewer-d3fc/src/config/d3fc.watch.config.js
+++ b/packages/perspective-viewer-d3fc/src/config/d3fc.watch.config.js
@@ -4,7 +4,10 @@ const rules = []; //pluginConfig.module.rules.slice(0);
 rules.push({
     test: /\.js$/,
     exclude: /node_modules/,
-    loader: "babel-loader"
+    loader: "babel-loader",
+    options: {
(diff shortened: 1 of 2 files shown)


DevAndyLee:
I think may need to cancel this PR.

In order to get the correct css from the d3fc package, we need to reference the individual packages' es6 modules (e.g. "@d3fc/d3fc-chart/srv/css"), rather than the compiled "d3fc" package (which doesn't export the css variables).

However, because of the way the babel compile works, those references aren't run through babel, which means they remain as es6 modules. That causes the error in UglifyJs - "Unexpected character '`'".

texodus:
As you say - I'm not aware of a way to import these as they are currently disted.  Looking into how [`d3fc-scripts`](https://github.com/d3fc/d3fc/blob/master/packages/d3fc-scripts/scripts/bundle.js) works however is illuminating - looks like it uses rollup/babel to package the css assets as JS modules, which seems much better than Perspective's webpack-plugin approach (though we would still need the latter for the WebWorker & wasm assets).

DevAndyLee:
Let's leave this for now then.

PR #583, opened 2019-05-16
Fixed tooltip value when split_values and multiple main values

DevAndyLee:
We previously showing the "mainValue" number, without subtracting "baseValue".
This affects Area and Column/Bar charts when there are both multiple aggregate values, and a split-by.


texodus:
Thanks for the PR!

PR #586, opened 2019-05-20
Added "Multi Chart" that allows mix-and-match of different series types

DevAndyLee:
This PR is a PoC and depends on PR #581. I thought I'd offer it this way so you can get a look at it and provide any feedback. Once #581 has been completed, we can re-do this as a PR into master.

The "Multi Chart" behaves much like the "Y Bar Chart", "Y Line Chart", "Y Area Chart" and "Y Scatter Chart", but can use series types from any of the above.

image

Notice the line/bar chart icons underneath the two y-axes. Clicking on them will cycle through the available series types for that axis (line/bar/area/scatter). The chosen series types will be remembered in "settings" if you save and load.

This chart reproduces a lot of the features of the other 4 charts (though is not quite identical), so I'm not sure how best to introduce it. Matt suggested maybe we could get rid of the other 4 charts, and provide a mapping function to this chart. For example, if someone is restoring a saved "Y Line Chart", we could instead show a "Multi Chart" with the "line" series selected.


DevAndyLee:
Rebased on `master`

davisford:
This looks pretty slick.  Wondering if there are plans to push this through?  Seems like the PR it was waiting on has been merged.

RajaPoseidon:
this is a useful , having multi type chart . Will it be soon available ?

PR #589, opened 2019-05-24
Row deltas return arrow-serialized data

sc1f:
When calling `view.on_update` with the `modes` parameter set to `row`, the callback receives an arrow-serialized version of the data from the rows that were updated. 

- `on_update` mode parameters have been changed to `cell` and `row`
- new `get_data` methods have been added to each context allowing for data from certain row indices to be returned.
- `to_arrow` and `col_to_js_typed_array` now take `options.data_slice,` which specifies a custom `t_data_slice` object from which they will serialize.
- `fill_col_bool` now respects `undefined` and `null` values for arrow-serialized booleans.
- `t_data_slice` has a new constructor which specifies a vector of row indices
- `process_subscribe` in `host.js` now transfers arrow data and respects the arguments that are in its caller. This enables full functionality for `on_update` mode flags as well as the arrow returned by `get_row_delta`.

### DIFF
--- a/cpp/perspective/src/cpp/context_one.cpp
+++ b/cpp/perspective/src/cpp/context_one.cpp
@@ -376,17 +376,19 @@ t_ctx1::get_row_delta() {
     PSP_TRACE_SENTINEL();
     PSP_VERBOSE_ASSERT(m_init, "touching uninited object");
     t_uindex eidx = t_uindex(m_traversal->size());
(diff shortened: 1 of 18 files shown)


texodus:
Thanks for the PR!

Very cool new feature, and Initial benchmarks look awesome!  I have some fixes for issues around practically setting up this example, and an example project for this, which I'll submit in a separate PR.

PR #596, opened 2019-05-26
simplify python setup.py, add psp_test to python tests

timkpaine:
Just a few boilerplate items before python merging:
- a better python make setup for perspective/table
- output perspective test into perspective/tests folder


texodus:
Thanks for the PR!

PR #607, opened 2019-06-04
Allow transfer as columns

timkpaine:
Small enhancement to allow transferring data as column-based. 
for https://github.com/timkpaine/perspective-python/issues/52


texodus:
Thanks for the PR!

PR #608, opened 2019-06-04
Restore highchairs on phosphor/jlab until d3 is complete

timkpaine:


texodus:
`@finos/perspective-viewer-d3fc` is now the default charting plugin for `@finos/perspective-viewer` - there are no more changes coming for this that will make this any more "complete" a transition than it is currently, e.g. `d3_y_line` is never going to be "promoted" to `y_line`.  You should not generally import both `@finos/perspective-viewer-highcharts` and `@finos/perspective-viewer-d3fc` in the same app - 

* You are massively increasing your payload size for identical plugins
* You have 2 identical plugins presented to he user with slightly different names.

I can merge this for now to deal with he fact that the `perspective-python` package lags master, but having these releases coordinated and tested in the same repository is a substantially better way to handle this in the future.

timkpaine:
"`d3_y_line` is never going to be "promoted" to `y_line`"

If this is the case, then `y_line` and all others should be deleted and renamed to `highcharts_y_line` to avoid confusion. `perspective-python` is not lagging in this case, it is simple that `y_line` and others no longer exist as names when they absolutely should be (most users don't care about the underlying js library). If we instead make it `<lib>_y_line` then i can enforce defaults in python

texodus:
Even better - `@finos/perspective-viewer-highcharts` is already marked deprecated.

timkpaine:
But there are missing chart types for d3

texodus:
Only `xy_line` currently - highcharts is e.g. missing `candlestick` and `ohlc`

timkpaine:
Does it have `y_bar` I didn't see

Issue #612, opened 2019-06-08
Demos broken on safari


timkpaine:
This webpage is broken on safari:
https://perspective.finos.org

404s on https://perspective.finos.org/blob:https://perspective.finos.org/perspective.wasm.worker.js.map

texodus:
Thanks for the report - this was fixed in #632 and will be in the 0.3.2 release.

PR #632, opened 2019-06-25
Fixed safari hypergrid rendering

texodus:


texodus:
Don't know exactly that the description is apt, but certainly Safari works now #612

timkpaine:
@texodus is this just on desktop? I believe the grid was also broken in Chrome on mobile.

PR #633, opened 2019-06-25
Prepare core C++ for Python API

sc1f:
This PR begins the process of building a new Python API which lives within the core `src/cpp` directory. The main change is the `val.h`header, which creates a header-only `t_val` class that allows us to share code between the Emscripten and Python API. 

Currently, we `typedef emscripten::val t_val`, thus the actual `t_val` header will not be used until the Python API is built.

- Clean up `emscripten.cpp` and document `binding.h`
- Add `val.h` and the `t_val` header
- Clean up `python` directory


timkpaine:
Can you revert the changes to the python directory and the build and test scripts, we are going to need those later.

timkpaine:
Take a look at my node PR for a view at what the python side will look like. For now just limit your changes to the cpp/js side

sc1f:
Closing in order to better write the Python API.

PR #643, opened 2019-07-03
Add hover styles to copy, download, reset

sc1f:
Slightly improve quality of life for the toolbar UI—adds hover styles and clearer titles for buttons and transpose button.


texodus:
Thanks for the PR!  Graphical polish is always welcome.

Issue #668, opened 2019-07-22
Autocomplete error when filtering string column with empty/null values


jspillers:
### Steps to Reproduce:
 1. yarn start simple
 2. navigate to a dataset that has a string column with null/empty values
 3. filter by said string column

### Expected Result:
autocomplete and filter both work

### Actual Result:
autocomplete throws an error when it tries to use null as a completion choice

<img width="818" alt="Screen Shot 2019-07-22 at 3 17 40 PM" src="https://user-images.githubusercontent.com/2891/61659368-4fc62100-ac95-11e9-923e-702b878f4e91.png">
<img width="741" alt="Screen Shot 2019-07-22 at 3 23 40 PM" src="https://user-images.githubusercontent.com/2891/61659366-4fc62100-ac95-11e9-9f10-e4492366570b.png">
<img width="587" alt="Screen Shot 2019-07-22 at 3 18 07 PM" src="https://user-images.githubusercontent.com/2891/61659367-4fc62100-ac95-11e9-8fa8-30af19a1b901.png">

jspillers:
Fixed by #669

PR #669, opened 2019-07-23
Fix exception for filter labels

jspillers:
Fix autocomplete error when filtering string column w/null values

* Factor out a function for filtering out null values before feeding
them into the autocomplete choices
* Add unit testing config to perspective-viewer package
* Add a unit test to cover this bugfix

Fixes issue: https://github.com/finos/perspective/issues/668


texodus:

PR #675, opened 2019-07-30
Fix #616: perspective-jupyterlab versioning

### LINKED ISSUE #616: Version pulled from package.json too early, ends up 1 version behind dist
https://github.com/finos/perspective/blob/master/packages/perspective-jupyterlab/src/ts/version.ts#L9

- replace with looser versioning scheme (minor version)

sc1f:
Adds a build in the `version` script, which `lerna publish` will run _after_ updating version numbers but _before_ pushing to the remote. This should update the disted Jupyterlab with the right version of Perspective. 

See comment on [Lerna issue #1275](https://github.com/lerna/lerna/issues/1275#issuecomment-367057638) for explanation of version lifecycle hook as it relates to Lerna.

### DIFF
--- a/packages/perspective-jupyterlab/package.json
+++ b/packages/perspective-jupyterlab/package.json
@@ -26,7 +26,8 @@
         "test": "npm-run-all test:run",
         "test:run": "echo done",
         "build": "webpack --color --config src/config/plugin.config.js",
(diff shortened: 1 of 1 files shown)


texodus:

PR #676, opened 2019-08-01
Remove nan filtering, add null filtering

jspillers:
Filter by null and not null

    * Add cases/constants to handle null/not null to emscripten.cpp,
      base.cpp/.h, scalar.cpp, and python.h
    * Fix table.cpp so that null values are not skipped when filtering by null
    * Alter _is_valid_filter in emscripten.cpp with a special case for
      null/not null so it doesnt reject those filters incorrectly
    * Refactor constants.js to allow for accessing filter operators in a
      non magic string way
    * Add special condition to row.js filter update so that the operand
      field is hidden when selecting null/not null filter
    * Add special condition to perspective.js is_valid_filter so null/not
      null is always valid (operand is irrelevant in this case)
    * Add Specs

    Remove NaN filtering and related feature flags

    * Retain the "true" behavior for nan handling where applicable


texodus:

PR #684, opened 2019-08-15
Fix scatter row_pivots_values exception on click

JHawk:


texodus:
Thanks for the PR!  Looks good

PR #692, opened 2019-08-28
Calculate row offset in C++, refactor Table, remove implicit primary key mode

sc1f:
This PR refactors the workings of the `Table` class and ports over some more functionality from JS to C++. Most importantly, the `limit_index`/`offset` property is now calculated within the `Table` class, which removes another chunk of state-dependent JS code. 

- Remove `GNODE_TYPE_IMPLICIT_PKEYED` which is not being used in the production codebase, and conflicts with features already present.
- Additional documentation added in JS and C++ for the table


texodus:

PR #703, opened 2019-09-03
`perspective-viewer` cleanup

texodus:
A late-summer cleaning of the `<perspective-viewer>` web component:

* Supports missing attributes now, and auto replace empty `[]` and `{}` attributes.
* Removes default shadowing aggregates.
* Supports boolean attributes for `settings` and `updating`.
* Removed `render_time`.
* Fixes serialization of state from `save()`.

This will be a breaking API change for most developers.

### DIFF
--- a/docs/static/css/material.dark.css
+++ b/docs/static/css/material.dark.css
@@ -15,7 +15,7 @@
  *
  */
 @import url("https://fonts.googleapis.com/css?family=Material+Icons");
-@import url("https://fonts.googleapis.com/css?family=Open+Sans");
+@import url("https://fonts.googleapis.com/css?family=Open+Sans:400,700");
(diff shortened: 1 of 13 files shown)


timkpaine:
Will this fix https://github.com/timkpaine/perspective-python/issues/29?

texodus:
No, that is not fixable, `attributeChangedCallback` does not properly support boolean attributes for web components at present (as also noted in the link).

RandomFractals:
more breaking api changes? 

@texodus can you please document them in your change log this time?

It becomes harder and harder to rely on your viewer lib at this point. is that intentional?

Thanks in advance! Looking forward to your summer break cleanup release :)

P.S: I do hope this is the last set to solidify that viewer api and configs ...

PR #706, opened 2019-09-03
Test implicit index & remove extraneous example file

sc1f:
This PR adds tests in the core engine to assert that behavior around implicit `__INDEX__` columns work as expected, in that updates with the `__INDEX__` column should behave like partial updates on the row ID specified in `__INDEX__`. 

- tests generation of implicit index column in `to_format`
- tests reading of implicit index for 0, 1, and 2-sided views
- removes extraneous `schema.html` in examples folder


texodus:

PR #707, opened 2019-09-05
Refactor C++

sc1f:
- Removes unnecessary string copies in `emscripten.cpp`
- Use `std::move` when possible in constructors for `Table`, `View`, and `t_view_config`
- Clarify purpose of `filter_op` in the view config, and refactor base methods to accept "or" as a valid filter op
- clean up filter parsing in `emscripten.cpp`


texodus:

PR #723, opened 2019-09-13
Python

timkpaine:


texodus:
Tasks transferred from issues

- [ ] Pull python docs from python folders into docusaurus
- [ ] Make sure datetime localization is consistent between psp internals and python to/from string
- [ ] Implement column extraction as numpy array or arrow column
- [ ] Make sure cpp folder ends up in perspective-python.table's manifest so that build happen.
- [ ] Implement to/from numpy code in C++

Also [here](https://github.com/finos/perspective/projects/2#card-26438285)

PR #729, opened 2019-09-17
Allow partial updates on computed column source columns

sc1f:
This PR refactors logic around constructing computed columns in order to move operations to the `t_gnode`'s `_process` method. 

Instead of constructing computed columns before the table is sent to `_process()`, use the metadata of the `_process` method to allow for partial updates (implicit and explicit index) on underlying columns for computed columns, which automatically propagates changes to computed columns. 

Previous iterations would recalculate each computed column for the total number of rows on the table; this allows computed column recalculation to happen ONLY for rows that were updated within `_process()`.


texodus:

PR #731, opened 2019-09-19
fixes view->plugin rename that didnt make it from original repo

timkpaine:


texodus:
Thanks for the PR!  Looks Great!

PR #733, opened 2019-09-21
Enable update with __INDEX__ on explicit index column

sc1f:
This PR removes the `abort()` we previously added for updates using implicit index, thus updates using `__INDEX__` (which always shadows the primary key) now work on indexed tables.


texodus:
Thanks for the PR!  Looks great!

PR #748, opened 2019-09-30
add compute() to dynamically show/hide sidebar computation panel

sc1f:
Because computed columns is not yet implemented in the Python API, make sure the viewer does not render the computation sidebar when `Table.compute()` returns false.


texodus:

PR #749, opened 2019-10-02
Fix update bugs

sc1f:
- Remove errant `only` in perspective.js tests
- Make sure correct callback indices are removed in global cache
- Test remove updates


texodus:

PR #751, opened 2019-10-03
ES6 Client API

sc1f:
Rewriting client as an ES6 class to allow for easier compatibility with Typescript in `perspective-jupyterlab` widget.


texodus:
Thanks for the PR!  Looks good.

PR #759, opened 2019-10-10
Add `to_csv` to Python API

sc1f:
This PR adds and tests `to_csv` for the perspective-python API. The method itself uses `pandas.DataFrame.to_csv`, so it includes the index columns that come as part of dataframe construction.

### DIFF
--- a/python/perspective/perspective/table/view.py
+++ b/python/perspective/perspective/table/view.py
@@ -289,11 +289,18 @@ class View(object):
                 - end_col: defaults to the total columns in the view
                 - index: whether to return an implicit pkey for each row. Defaults to False
                 - leaves_only: whether to return only the data at the end of the tree. Defaults to False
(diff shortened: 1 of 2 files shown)


python/perspective/perspective/table/view.py
```diff
@@ -277,6 +277,24 @@ def to_df(self, **options):
...
+                - end_row: defaults to the number of total rows in the view
+                - start_col: defaults to 0
+                - end_col: defaults to the total columns in the view
+                - index: whether to return an implicit pkey for each row. Defaults to False
+                - leaves_only: whether to return only the data at the end of the tree. Defaults to False
+
+        Returns:
+            str : a CSV-formatted string containing the serialized data.
+        '''
+        return self.to_df(**options).to_csv(date_format='%Y/%m/%d %H:%M:%S')
```
timkpaine: can you pull `date_format` from **options to support overriding

`date_format = options.pop('date_format', '%Y...`

texodus:

PR #767, opened 2019-10-14
Use ciso8601 date parser

sc1f:
This PR adds the ciso8601 date parser as a first-try parse for string dates. Tested on a 52x10000 dataframe of string datetime values, the new parser improved dataframe loading performance by a factor of 4.


timkpaine:
Don't use this.

PR #768, opened 2019-10-14
Upgrade Arrow to 0.15.0, link python arrow from prebuilt library

sc1f:
This PR upgrades our version of arrow to 0.15.0, and restricts the custom build of arrow + flatbuffers to WASM and C++ only builds.

For Python builds, we link against `libarrow.15.dylib` which is built when you run `pip install pyarrow`. Make sure pyarrow is upgraded to the latest version.

Works on my machine :)


cmake/modules/FindPyArrow.cmake
```diff
@@ -47,6 +47,7 @@ find_path(PYTHON_PYARROW_INCLUDE_DIR arrow/python/api.h
   HINTS "${__pyarrow_path}" "${PYTHON_INCLUDE_PATH}" NO_DEFAULT_PATH)
 
 set(PYTHON_PYARROW_LIBRARY_DIR ${__pyarrow_library_dirs})
+set(PYTHON_PYARROW_SHARED_LIBRARY ${__pyarrow_library_dirs}/libarrow.15.dylib)
```
timkpaine: No good

cmake/modules/FindPyArrow.cmake
```diff
@@ -47,6 +47,7 @@ find_path(PYTHON_PYARROW_INCLUDE_DIR arrow/python/api.h
   HINTS "${__pyarrow_path}" "${PYTHON_INCLUDE_PATH}" NO_DEFAULT_PATH)
 
 set(PYTHON_PYARROW_LIBRARY_DIR ${__pyarrow_library_dirs})
+set(PYTHON_PYARROW_SHARED_LIBRARY ${__pyarrow_library_dirs}/libarrow.15${CMAKE_SHARED_LIBRARY_SUFFIX})
```
timkpaine: let me validate on windows

texodus:
Thanks for the PR!  Looks Good!

I've updated the docker images to use `pyarrow` 0.15.0, but at least for now these are not used.

PR #776, opened 2019-10-21
Multiple views on JLab plugin

sc1f:
This PR allows for multiple output views on the same Jupyter widget - all operations are shared between the widget instances, so making a change to pivots, filter etc. in one rendered instance will change the other.


texodus:
Thanks for the PR!

PR #777, opened 2019-10-21
Add exception handling, clean up PSP_COMPLAIN_AND_ABORT

sc1f:
This PR cleans up our error handling throughout the C++ and Python codebase.

- Adds `PerspectiveCppError`, an exception that will be raised whenever `psp_abort()` is called (by `PSP_COMPLAIN_AND_ABORT`, for example). This simplifies error handling in Python significantly, as all exceptions from Perspective can be caught with `except (PerspectiveError, PerspectiveCppError)`.
- `PerspectiveManager` now catches all errors and propagates them back to the client, which will return a rejected promise to the user. Previously, errors were propagated as keyboard interrupts, and an error on one client—pivoting by an unknown column, for example—would crash the server. 
- Removes unneeded outputs to `cerr` and moves the output to the `psp_abort` function.
- Removes `assets` folder from `python/node`, and adds to `.gitignore`.


timkpaine:
looks good!

texodus:
Thanks for the PR!

PR #778, opened 2019-10-21
Adjust setup.py for MacOS wheel dist

sc1f:
This PR allows us to build a portable wheel distribution for OSX:

- Our binaries depend on @rpath, which is set relative to the build location of the binary. This causes a problem when distributing to PyPi/installing from pip. In `setup.py`, we adjust the install names of the built binary to reference `@loader_path`, i.e. relative to the actual location of the binary download.
- Pin pyarrow in our requirements. This means that `TBB` is the only system requirement that needs to be installed from brew/build from source.
- Fix bumpversion config
- Fix manifest - graft source files and not tests


timkpaine:
looks good!

texodus:
Thanks for the PR!

PR #780, opened 2019-10-21
Link `lerna version` to `bumpversion`

sc1f:
This PR adds `version_python.js` to the `version` script, which will run after Lerna increases the version number but before pushing to the remote.


texodus:
Nice, thanks for the PR!

PR #784, opened 2019-10-28
Python 2/GCC <5 compatibility

sc1f:
This PR fixes compatibility issues in `perspective-python` between Python 2 and Python 3, as well as removing C++ features not present in older compilers.

- Fixes build and test process for Python 2, adds `--python2` flag to yarn build and yarn test
- Adds Travis CI task and Docker image for python2
- Adds `client` mode for PerspectiveWidget - instead of creating a Table, the data is sent to the browser and owned wholly by `perspective-viewer`
- Adds [HowardHinnant/Date](https://github.com/HowardHinnant/date) library for serializing datetimes to string, and paves way for rewriting our datetime parser in the future
- Removes `codecvt` and `std::put_time`, as well as unnecessary `std::wstring` conversions for Python.
- Fixes tests that previously depended on Python 3.7's [guaranteed dict insertion order](https://docs.python.org/3.7/tutorial/datastructures.html#dictionaries).
- Fixes broken string string by using `unicode()`
- Fixes superstore and streaming example on Python 2 and Python 3


cpp/perspective/CMakeLists.txt
```diff
@@ -101,6 +111,9 @@ if (NOT DEFINED PSP_PYTHON_BUILD)
 	set(PSP_PYTHON_BUILD OFF)
 elseif(PSP_PYTHON_BUILD)
 	set(PSP_CPP_BUILD ON)
+	if(NOT DEFINED PSP_PYTHON_VERSION)
+		set(PSP_PYTHON_VERSION 3)
```
timkpaine: Set to 3.7

cpp/perspective/CMakeLists.txt
```diff
@@ -273,10 +286,16 @@ elseif(PSP_CPP_BUILD OR PSP_PYTHON_BUILD)
 		#########################
 		# PYTHON BINDINGS BUILD #
 		#########################
-		include_directories("/usr/local/include/python3.7m")  # FIXME
+		set(PYTHON2 string_starts_with("${PSP_PYTHON_VERSION}" "2"))
+		if(PYTHON2)
```
timkpaine: This was an experiment from before, we shouldn't need it anymore

python/perspective/perspective/core/widget.py
```diff
@@ -52,10 +53,10 @@ class PerspectiveWidget(Widget, PerspectiveViewer):
     # Required by ipywidgets for proper registration of the backend
     _model_name = Unicode('PerspectiveModel').tag(sync=True)
     _model_module = Unicode('@finos/perspective-jupyterlab').tag(sync=True)
-    _model_module_version = Unicode('^0.3.9').tag(sync=True)
+    _model_module_version = Unicode('^0.4.0-rc.1').tag(sync=True)
```
timkpaine: We need to brainstorm something for this down the road, and try to avoid breaking changes in patch releases

python/perspective/perspective/src/accessor.cpp
```diff
@@ -74,14 +74,10 @@ infer_type(t_val x, t_val date_validator) {
...
+        t = t_dtype::DTYPE_FLOAT64;
     } else if (py::isinstance<py::int_>(x)) {
-        double x_float64 = x.cast<double>();
-        if ((std::fmod(x_float64, 1.0) == 0.0) && (x_float64 < 10000.0)
-            && (x_float64 != 0.0)) {
-            t = t_dtype::DTYPE_INT32;
-        } else {
-            t = t_dtype::DTYPE_FLOAT64;
-        }
+        t = t_dtype::DTYPE_INT32;
```
timkpaine: Do we have an overflow test? In python 3 ints are int64s

python/perspective/perspective/tests/core/test_widget_pandas.py
```diff
@@ -51,6 +51,10 @@ def superstore(count=50):
 class TestWidgetPandas:
 
     def test_widget_load_table_df(self):
+        '''Flaky on Python 2 - sometimes unicode encoding errors occur.
```
timkpaine: My superstore can generate non English character names, give it a bigger size to guarantee the error

python/perspective/perspective/core/widget.py
sc1f: maybe the `version_python` script would work here

python/perspective/perspective/src/accessor.cpp
sc1f: added overflow test - promotion takes care of overflow

python/perspective/perspective/tests/core/test_widget_pandas.py
sc1f: fixed a few commits ago - removed outdated comment

PR #791, opened 2019-11-04
Dataframe/Numpy Array Loader

sc1f:
This PR adds a separate load path for Pandas dataframe and numpy arrays (dictionaries of numpy arrays, numpy structured arrays, and numpy record arrays. 

- It will try to copy all arrays of integral/float/bool type, and fill other types through iteration.
- Arrays of `dtype=object` will always be iteratively filled using the `marshal` method, which coerces data into the correct format. 
- Support for mixed dictionaries of numpy arrays and lists has been deprecated: datasets should be fully composed of `numpy.ndarray` or `list`, but not both.
- Adds support for numpy structured arrays.
- Adds and fully tests support for all numpy datetime types from nanosecond (`datetime64[ns]`) to year (`datetime64[Y]`).
- Adds support for loading timestamps (in seconds or milliseconds since epoch) into columns already typed as `datetime`. To load timestamps as datetimes, either construct the table with a schema of {"name": datetime}, or with a valid datetime/Timestamp/datetime64 object.
- Adds and tests support for timedelta64 arrays, which are cast into string to preserve their units.
- Adds and tests support for `pandas.DateTimeIndex` and `pandas.PeriodIndex`.

### DIFF
--- a/cpp/perspective/CMakeLists.txt
+++ b/cpp/perspective/CMakeLists.txt
@@ -449,6 +449,7 @@ set (PYTHON_SOURCE_FILES
 	${PSP_PYTHON_SRC}/src/accessor.cpp
 	${PSP_PYTHON_SRC}/src/context.cpp
 	${PSP_PYTHON_SRC}/src/fill.cpp
+	${PSP_PYTHON_SRC}/src/numpy.cpp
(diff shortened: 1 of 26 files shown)


texodus:
Thanks for the PR!

PR #793, opened 2019-11-04
Add `update` for widget in client mode, fix `on_update` in Python 2

sc1f:
This PR fixes some remaining issues with the Python plugin:

- `on_update` now works in Python 2 after fixing a bug due to different implementations of `functools.partial` between Python 3 and Python 2
- `update` has been implemented for widgets running in client-only mode
- remove all build artifacts for python on each run of `yarn build`


texodus:

PR #802, opened 2019-11-08
Tweak date/datetime inference, remove dependency on non-core Numpy/Pandas API

sc1f:
This PR fixes the inference of dates and datetimes for pandas DataFrames, specifically for `DatetimeIndex` of various frequencies. Instead of inferring the data type based on the `datetime64` unit, we use the index's datetime frequency to decide whether to type the column as a date or a datetime.

Other fixes:
- remove `fillna` from our dataframe processing flow
- remove dependencies on `masked_invalid` and non top-level numpy/pandas API
- enforce all column names as strings, and throw an error otherwise
- add new tests + `util.make_dataframe` and `util.make_series` as test utilities

### DIFF
--- a/python/perspective/perspective/table/_accessor.py
+++ b/python/perspective/perspective/table/_accessor.py
@@ -10,7 +10,8 @@ import pandas
 import numpy
 from math import isnan
 from ._date_validator import _PerspectiveDateValidator
-from ..core.data import deconstruct_numpy
+from ..core.data import deconstruct_numpy, deconstruct_pandas
+from ..core.data.pd import _parse_datetime_index
 from ..core.exception import PerspectiveError
 
 try:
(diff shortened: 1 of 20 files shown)


python/perspective/perspective/table/_accessor.py
```diff
@@ -104,6 +96,29 @@ def __init__(self, data_or_schema):
 
         self._types = []
 
+        # Verify that column names are strings, and that numpy arrays are of type `ndarray`
+        if six.PY2:
+            valid_name_types = (str, unicode)  # noqa: F821
```
timkpaine: `isinstance(whatever, six.string_types)`

timkpaine: just keep this bookmarked for any time youre not sure https://six.readthedocs.io/#module-six

PR #807, opened 2019-11-11
Allow client mode widget to be constructed with schema

sc1f:
This PR allows widgets to be constructed with schemas when using the client-only flag. Previously, type objects were not serialized properly, and schema construction failed.


texodus:

Issue #810, opened 2019-11-12
Widget does not automatically re-render when calling load() after initializing with `None`


sc1f:
### Steps to Reproduce:
 1. Create a `PerspectiveWidget`, passing in `None` as the first parameter
 2. Rendering the widget.
 3. Call `load()` on the widget with a dataset.

### Expected Result:
Widget should automatically re-render with the new dataset visible.

### Actual Result:
Widget does not re-render, but it displays properly after manually rendering.

PR #811, opened 2019-11-12
Hypergrid no longer depends on `Object.keys` order of dataset

sc1f:
This PR changes the way Hypergrid renders column headers; instead of using `Object.keys` on an initial slice of the dataset, the `column_paths` method is added to the `View` instance. This method returns a vector of column paths that show exactly the column headers expected in the view. For pivoted views, it will prepend `__ROW_PATH__` to the array. 

Because this method respects the ordering of columns in the `columns` param of the View, hypergrid columns will no longer be out of order.

Other fixes:
- Fixes bug #810 by proactively sending new data/new table to the front-end on `load`.
- Added tests for column path in JS and Python.


texodus:

PR #812, opened 2019-11-12
Add puppeteer tests for `perspective-phosphor`

zemeolotu:


texodus:
Thanks for the PR!  More tests are always a welcome addition!

PR #813, opened 2019-11-12
Client mode supports dataframes, np.ndarray, structured and recarray

sc1f:
Add support for pandas and numpy data structures into client mode widget - previously these weren't getting serialized properly.

Additionally, add `clear()` and `replace()` to the viewer, as well as remove column promotion logic from updates - it will cause an error.


texodus:

PR #817, opened 2019-11-15
Re-Fix python build

timkpaine:
[This commit](https://github.com/finos/perspective/commit/966d533f8f56a3829cb24b07f530a0b437285154#diff-b23a161aa71df80df5c594015a1cdb7b) removed the sdist regression test. This PR is an attempt to fix this.


timkpaine:
Screen Shot 2019-11-17 at 1 46 03 PM

timkpaine:
Docker builds should be good once we update the docker files

docker/python/Dockerfile
```diff
@@ -24,21 +22,25 @@ RUN rm -fr /opt/cmake/bin/cpack
...
 
-RUN ln -s /usr/local/lib/libboost_python37.so /usr/local/lib/libboost_python.so
-RUN ln -s /usr/local/lib/libboost_numpy37.so /usr/local/lib/libboost_numpy.so
+# install pyarrow
+RUN wget https://github.com/apache/arrow/archive/apache-arrow-0.15.0.tar.gz >/dev/null 2>&1  || echo "wget arrow failed"
+RUN tar xfz apache-arrow-0.15.0.tar.gz
+RUN cd arrow-apache-arrow-0.15.0 && mkdir build && cd build && cmake ../cpp/ -DARROW_RPATH_ORIGIN=ON -DARROW_PYTHON=ON -DARROW_FLIGHT=ON -DARROW_COMPUTE=ON -DCMAKE_INSTALL_PREFIX=/usr -DARROW_DATASET=OFF -DARROW_BUILD_UTILITIES=OFF -DARROW_JEMALLOC=OFF -DARROW_DEPENDENCY_SOURCE=BUNDLED -DARROW_HDFS=OFF -DARROW_WITH_BACKTRACE=OFF -DARROW_WITH_BROTLI=OFF -DARROW_WITH_BZ2=OFF -DARROW_WITH_LZ4=OFF -DARROW_WITH_SNAPPY=OFF -DARROW_WITH_ZLIB=OFF -DARROW_WITH_ZSTD=OFF && make -j2 && sudo make install && cd ../python && sudo python3.7 setup.py install
 
-RUN python3 -m pip install codecov nose2 mock flake8 pytest pytest-cov traitlets ipywidgets faker psutil zerorpc
-RUN python3 -m pip install 'numpy>=1.13.1' 'pandas>=0.22.0' 'pyarrow==0.15.0'
```
texodus: Shouldn't we be testing the way users are wont to have pyarrow installed, namely `pip install pyarrow`, rather than build from source here?

cmake/modules/FindPyArrow.cmake
```diff
@@ -50,17 +50,20 @@ set(PYTHON_PYARROW_LIBRARY_DIR ${__pyarrow_library_dirs})
 
 if (CMAKE_SYSTEM_NAME MATCHES "Darwin")
   # Link against pre-built libarrow on MacOS
-  set(PYTHON_PYARROW_SHARED_LIBRARY ${__pyarrow_library_dirs}/${CMAKE_SHARED_LIBRARY_PREFIX}arrow.15${CMAKE_SHARED_LIBRARY_SUFFIX})
+  set(PYTHON_PYARROW_PYTHON_SHARED_LIBRARY ${__pyarrow_library_dirs}/${CMAKE_SHARED_LIBRARY_PREFIX}arrow_python.15.dylib)
+  set(PYTHON_PYARROW_ARROW_SHARED_LIBRARY ${__pyarrow_library_dirs}/${CMAKE_SHARED_LIBRARY_PREFIX}arrow.15.dylib)
+else()
+  # linux
+  set(PYTHON_PYARROW_PYTHON_SHARED_LIBRARY ${__pyarrow_library_dirs}/${CMAKE_SHARED_LIBRARY_PREFIX}arrow_python${CMAKE_SHARED_LIBRARY_SUFFIX}.15)
+  set(PYTHON_PYARROW_ARROW_SHARED_LIBRARY ${__pyarrow_library_dirs}/${CMAKE_SHARED_LIBRARY_PREFIX}arrow${CMAKE_SHARED_LIBRARY_SUFFIX}.15)
```
texodus: This step fails on the docker image for python3, as far as I can discern.

```bash
CMake Error at /opt/cmake/share/cmake-3.15/Modules/FindPackageHandleStandardArgs.cmake:137 (message):
  Could NOT find PyArrow: Found unsuitable version "", but required is at
  least "0.15" (found PYTHON_PYARROW_INCLUDE_DIR-NOTFOUND)
```

python/perspective/setup.py
```diff
@@ -146,31 +152,41 @@ def build_extension_cmake(self, ext):
...
         env['PSP_ENABLE_PYTHON'] = '1'
         env["PYTHONPATH"] = os.path.pathsep.join((os.path.join(os.path.dirname(os.__file__), 'site-packages'), os.path.dirname(os.__file__)))
 
         if not os.path.exists(self.build_temp):
             os.makedirs(self.build_temp)
-
         try:
-            subprocess.check_output([self.cmake_cmd, os.path.abspath(ext.sourcedir)] + cmake_args, cwd=self.build_temp, env=env, stderr=subprocess.STDOUT)
-            subprocess.check_call([self.cmake_cmd, '--build', '.'] + build_args, cwd=self.build_temp, env=env, stderr=subprocess.STDOUT)
+            out1 = subprocess.check_output([self.cmake_cmd, os.path.abspath(ext.sourcedir)] + cmake_args, cwd=self.build_temp, env=env, stderr=subprocess.STDOUT)
```
texodus: `check_output` has really shitty user experience when building from the console - errors lack highlighting and incremental output is suppressed.  If there is no functional reason to change this logging, I would prefer to revert these.

scripts/ci_python.js
```diff
@@ -39,7 +39,17 @@ try {
...
+        codecov --token 0f25973b-091f-42fe-a469-95d1c6f7a957`;
+    if (!PY2) {
+        // dont build docs/install check for python2
+        build_cmd += `&&\
+        make -C ./docs html &&\
+        ${python} -m pip install -U pip &&\
+        ${python} -m pip install . &&\
+        ${python} setup.py sdist &&\
+        ${python} -m pip install -U ./dist/perspective*`;
+    }
```
texodus: `ci_python` and `build` (with a proper [`.perspectiverc`](https://perspective.finos.org/docs/md/development.html#building)) are identical, except the former runs `install` and some other sdist tasks now.  Can we
1. Merge this with `build` so our CI and dev builds are the same?
2. Remove 'pip install` Install the necessary python modules as part of the docker image, so this is consistent, build time is conserved, etc?
3. Move the sdist tasks to `test` or at least put it behind a build flag so it is not mandatory?

docker/python/Dockerfile
timkpaine: Perspective is not compatible with the manylinux pyarrow wheel, if you can get perspective to build on centos6 with gcc 4.8 (a non C++11 compiler) then we can use the pyarrow wheel https://www.python.org/dev/peps/pep-0571/#id48

python/perspective/setup.py
timkpaine: I have no preference, but I'm done trial-and-erroring this code, whoever wants this to change can implement that change

scripts/ci_python.js
timkpaine: 1 sure
2 this makes dependency adding/changing more annoying, it will require a docker image upgrade anytime we add or modify a dependency. Doesn't matter to me, you'll be trading faster incremental builds for much longer occasional builds (when you have to rebuild and repush the docker image).
3 I don't care where they go so long as they are tested

docker/python/Dockerfile
texodus: I propose if this is something we want to maintain tests for on travis, we should build a separate custom container for it, and either way we want to maintain this container as-is for python3 testing.  E.g. this should be reverted.

python/perspective/setup.py
texodus: Not asking for a change - this reverts the current behavior on master for a reasons I don't understand yet.  Can I revert this?

timkpaine: The current master raises an exception if the subprocess call fails because out has no attribute decode.

cmake/modules/FindPyArrow.cmake
timkpaine: did you rebuild the docker image? running from scratch this works fine for me, the error you're getting suggest pyarrow was not installed

texodus:
Discussed offline & decided to close

PR #818, opened 2019-11-15
Fix misordered columns in update

sc1f:
Updates previously used the `names` property of the table, which may be of a different order than the schema of the `t_gnode`. Remove the use of `names` and replace with `get_columns()` from the gnode schema.


texodus:

PR #819, opened 2019-11-15
Update Perspective website with Python API and user guide

sc1f:
- Add Python documentation to the user/developer guide
- Makes developer guide clearer, pins our dependencies and versions
- Add auto-generated API documentation from docstrings
- Minor restyles of Perspective site with new fonts


texodus:

PR #821, opened 2019-11-18
Default to int64 in Python3, add `long` and `unicode` to schema and type inference

sc1f:
This PR changes our inference check to infer int32s in python 2 and int64 in python 3. Additionally, it allows `long` and `unicode` to be used in schemas; `long` maps to int64, and `unicode` maps to string. 

Actual storage semantics of the values are not changed, so the limit for numeric types remains at 64 bits. Passing in a value that overflows int64 is undefined behavior.

Additionally wraps the call to `iter` to generate more meaningful errors when assessing the type of a dataset.


texodus:

PR #822, opened 2019-11-18
Properly remove `on_delete` and `on_update` callbacks that fail.

sc1f:
This PR fixes behavior around update/delete callbacks in JS. Previously, only `remove_update` was called in case of a failing callback—this PR calls the right removal function based on the method (`remove_update` or `remove_delete`).


texodus:

PR #824, opened 2019-11-24
Add CSV/JSON renderer in JupyterLab

timkpaine:
A few open issues:
- [ ] accept column-oriented json instead of row-oriented json
- [ ] if perspective load fails, `showDialog`
- [ ] detect theme changes and switch between light and dark 

@texodus could you take a look at these, the second one do we need to wait for the async load to fail, is there a callback or something?


timkpaine:
tmp

timkpaine:
moving source branch

PR #826, opened 2019-11-26
Widget fixes for resizing, TS typings, boolean columns

sc1f:
- Fix widget resize by calling `notifyResize` properly
- Fix typing for Perspective Viewer by making table options optional
- Fix exception thrown when string booleans ("True", "False") were passed into Perspective
  - use [`distutils.util.strtobool`](https://docs.python.org/2/distutils/apiref.html#distutils.util.strtobool) to parse truthy/falsy values
- Make sure column pivots on booleans return `true` or `false` as their column name, not 0/1


texodus:

PR #829, opened 2019-12-03
Read date32, date64, decimal128 from Arrow datasets

sc1f:
This PR adds support for reading `date32`, `date64`, and `decimal128` columns from Arrow-serialized datasets. `date32` and `date64` columns are typed as `date`, while `decimal128` columns are typed as `int64`.

Additionally, utility functions have been added for loading Arrows from the filesystem (in JS tests) and generating Arrow binaries using PyArrow (for Python tests). This allows us to quickly add test cases for Arrow datasets in Perspective.


texodus:
Thanks for the PR!

PR #832, opened 2019-12-03
CSV/JSON renderer in JupyterLab

timkpaine:
Initial cut at a CSV/JSON viewer in jupyterlab

Standalone CSV/JSON viewer:
- Can parse records-oriented JSON or CVS, available from the right click menus for `Open With`
- Shows error for datatypes can't parse
- Grabs light/dark theme from environment
- Changes when theme is changed

Updates for `PerspectiveWidget`:
- If you don't pass in `dark=True` or `dark=False`, it will pick it up from the environment

### DIFF
--- a/packages/perspective-jupyterlab/src/ts/plugin.ts
+++ b/packages/perspective-jupyterlab/src/ts/plugin.ts
@@ -39,4 +39,3 @@ export const PerspectiveJupyterPlugin: IPlugin<Application<Widget>, void> = {
     autoStart: true
 };
 
-export default PerspectiveJupyterPlugin;
--- a/packages/perspective-jupyterlab/src/ts/renderer.ts
+++ b/packages/perspective-jupyterlab/src/ts/renderer.ts
@@ -1,0 +1,249 @@
+import {ActivityMonitor} from '@jupyterlab/coreutils';
+import {ILayoutRestorer, JupyterFrontEnd, JupyterFrontEndPlugin} from '@jupyterlab/application';
+import {IThemeManager, WidgetTracker, Dialog, showDialog} from '@jupyterlab/apputils';
+import {ABCWidgetFactory, DocumentRegistry, IDocumentWidget, DocumentWidget} from '@jupyterlab/docregistry';
+import {PerspectiveWidget} from "@finos/perspective-phosphor";
+
+/**
+ * The name of the factories that creates widgets.
+ */
+const FACTORY_CSV = 'CSVPerspective';
+const FACTORY_JSON = 'JSONPerspective';
+
+const RENDER_TIMEOUT = 1000;
+
+type IPerspectiveDocumentType = "csv" | "json";
+
+// create here to reuse for exception handling
+const baddialog = () => {
+          showDialog({
+          body: "Perspective could not render the data",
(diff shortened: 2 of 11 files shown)


timkpaine:
Don't merge yet, need to fix themes and squash

packages/perspective-jupyterlab/src/ts/plugin.ts
```diff
@@ -25,8 +25,10 @@ import {PERSPECTIVE_VERSION} from "./version";
  */
 export const PerspectiveJupyterPlugin: IPlugin<Application<Widget>, void> = {
     id: "@finos/perspective-jupyterlab",
-    requires: [IJupyterWidgetRegistry],
+    requires: [IJupyterWidgetRegistry], //the "as any" is a workaround for @jupyter-widgets phosphor changes
     activate: (app: Application<Widget>, registry: IJupyterWidgetRegistry): void => {
+        console.log(registry);
```
sc1f: Remove console.log

packages/perspective-jupyterlab/src/ts/view.ts
```diff
@@ -17,7 +17,7 @@ import {PerspectiveJupyterClient, PerspectiveJupyterMessage} from "./client";
  * the DOM.
  */
 export class PerspectiveView extends DOMWidgetView {
-    pWidget: PerspectiveWidget;
+    pWidget: PerspectiveWidget;  // this should be pWidget, but temporarily calling it pWidget for widgets incompatibilities
```
sc1f: Confused - what does this comment mean

timkpaine:
tmp

packages/perspective-jupyterlab/src/ts/renderer.ts
```diff
@@ -0,0 +1,249 @@
...
+      } else if (this._type === "json") {
+        const data = this._context.model.toJSON() as any;
+
+        if (Array.isArray(data) && data.length > 0) {
+          // already is records form, load directly
+          this._psp._update(data);
+
+        } else {
+          throw "Not handled";
+          // TODO
```
sc1f: Can we get rid of the comments/refactor it out to a separate function

timkpaine: replaced with a comment describing the intent of what would go here, should anyone else feel like implementing it

texodus:
Thanks for the PR and review!  This is a cool feature!

I will merge this now, but I have reservations about this design.  I do not think `dark` keyword argument is the right API for this, and I definitely do not think it belongs on `PerspectiveViewer` class.  IMO, `dark` is Jupyter-specific and should be implemented as `Widget.addClass()` calls [here where we already know the jupyter dark `class` names](https://github.com/finos/perspective/pull/832/files#diff-e7ba0b880cee75b4cbc76ed5073a626aR215), rather than manipulating an attribute `dark` that ultimately does the same thing.  

`PerspectiveViewer` inherits from `Widget` which makes this possible, as `<perspective-viewer>` inherits `HTMLElement`, and should similarly use `class` to manipulate DOM style.

timkpaine:
Yep, I think we want to expose full styles themselves into python rather than just light/dark . But light/dark was an existing api, not a new api, so it wouldn't make sense to implement that change in this PR.

PR #833, opened 2019-12-04
Preserve user columns and pivots in widget

sc1f:
When passing in `pandas.MultiIndex` to `PerspectiveWidget`, the parsed columns and pivots take precedence over all `kwargs` passed in by the user. This PR prefers user input for widgets created from dataframe for `columns`, `row_pivots`, and `column_pivots`.


texodus:
Thanks for the PR!

PR #834, opened 2019-12-05
Set Enums as values for Widget/Viewer, refactor test folder structure

sc1f:
Previously, `Aggregate`, `Sort`, and `Plugin` enums could be used at `__init__` for the viewer and widget, but not be set as attributes as they fail Traitlets type checking. This PR overrides `__setattr__` for `PerspectiveViewer` to proactively call validate on the Enum, thus converting it to a string before setting it on the instance.

Additionally, this PR refactors the test folder structure to match that of the Python source, and adds new tests for the Enums.


texodus:
Thanks for the PR and Review!  Looks good!

PR #835, opened 2019-12-06
Throttle fix

texodus:
Fixes event throttling in perspective-viewer to work as originally intended, as well as several fixes to scripts to make this change easier to build and test.


texodus:
Thanks for the review!

PR #837, opened 2019-12-07
Improvements to Arrow updates and indexed columns

sc1f:
- Updates using Apache Arrow use Perspective table schema, thus allowing single-column and partial updates with Arrow datasets
- Assert correct behavior for arrow updates on tables created from schema, where arrow int64 needs to supersede int32 columns defined in Perspective.
- Rewrite and centralize arrow loading in Perspective.js test suite
- Allow for null to be a valid value in columns used as index
- Remove vestigial code in `t_vocab`, unify behavior of `t_vocab` across JS and Python runtimes
- Add more arrow tests in Python, including for dictionary, file writer, and legacy-flagged arrows


texodus:
Thanks for the PR!  Very nice work and great test coverage!

PR #838, opened 2019-12-07
Asynchronously process updates when running in Tornado

sc1f:
This PR contains an integration between `perspective-python` and the Tornado event loop, namely allowing Perspective to call `_process` on updates (an expensive operation) asynchronously. 

When using `PerspectiveTornadoHandler`, the default `queue_process` method is replaced with a version that uses `tornado.ioloop.IOLoop.current()`, and the `_process` call becomes a callback for the next loop iteration. 

This should also allow for deeper integration with other event loop structures, which would implement their own version of `queue_process`, and then call `_set_queue_process` on an instance of `PerspectiveManager`. The tables under management (and future tables hosted on the manager instance) will now call the custom `queue_process` method. This structure also allows for manager instances that use the event loop and other instances that do not use the event loop all in the same runtime.

Various tests have been added for the async functionality.

### DIFF
--- a/examples/webpack/index.js
+++ b/examples/webpack/index.js
@@ -16,10 +16,7 @@ require("@finos/perspective-viewer-d3fc");
 
 window.addEventListener("WebComponentsReady", () => {
     const worker = perspective.worker();
-    const table = worker.table([
-        {x: 1, y: 2},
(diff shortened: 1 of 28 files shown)


texodus:
Thanks for the PR!

PR #839, opened 2019-12-10
Python build overhaul

timkpaine:
This PR overhauls the python build system. The primary features are:
- reenabling installable sdists (so we can publish to pypi)
- adding a `manylinux2010` build for [PEP 571](https://www.python.org/dev/peps/pep-0571/) to allow us to publish linux wheels for python 3.7
- the above, for python 2.7
- adding a `manylinux2014` build for [PEP 599](https://www.python.org/dev/peps/pep-0599/) to allow us to publish linux wheels for python 3.7

It also
- cleans up the `CMakeLists.txt` for cleaner multi-platform configuration
- cleans up the `setup.py` file according to @texodus suggestions for cleaner outputs
- cleans up env vars in the `travis.yml`
- cleans up and shortens the `boost` build in the python docker images (removes boost python build, only builds filesystem and system)
- fixes compatibility with both wheel installed pyarrow and source-installed pyarrow
- industrializes some hacks that were previously necessary for centos 6 (e.g. `-std=c++1y` flag) 
- fixes some pybind issues on mac when installing with brew vs pip vs vendored in build


docker/python2/official/Dockerfile
```diff
@@ -0,0 +1,55 @@
...
+RUN ln -s /opt/cmake/bin/ctest /usr/local/bin/ctest
+
+RUN python2.7 -m pip install numpy pandas cython 'mock==2.0.0' codecov nose2 flake8 pytest pytest-cov traitlets ipywidgets faker psutil zerorpc 'backports.shutil_which'
+
+# install boost
+RUN wget https://dl.bintray.com/boostorg/release/1.71.0/source/boost_1_71_0.tar.gz >/dev/null 2>&1
+RUN tar xfz boost_1_71_0.tar.gz
+# https://github.com/boostorg/build/issues/468
+RUN cd boost_1_71_0 && ./bootstrap.sh
+RUN cd boost_1_71_0 && ./b2 -j8 --with-program_options --with-filesystem --with-system install
```
texodus: Would be really nice if the diffs between these `Dockerfile` used a base image, which would also alleviate the build times.

python/perspective/setup.py
```diff
@@ -1,38 +1,79 @@
-################################################################################
+# *****************************************************************************
```
texodus: 👎

texodus:
Thanks for the PR!

PR #840, opened 2019-12-10
Async resize

texodus:
Makes `notifyResize()` and `toggleConfig()` methods async, such that redraws for size changes can `await` the finished render.  The PR also resolves several race conditions in this feature, and makes the test suite better able to track these sorts of issues by adding:

* `--interactive` will run Puppeteer in headed mode and pause when an error is detected.
* Puppeteer is now shared across tests, improving runtime speed.
* `--saturate` works across tests now for running test cases in a loop until they fail.


packages/perspective-phosphor/src/ts/widget.ts
```diff
@@ -368,7 +368,7 @@ export class PerspectiveWidget extends Widget {
         if (!viewer.notifyResize) {
             console.warn("Warning: not bound to real element");
         } else {
-            const resize_observer = new MutationObserver(viewer.notifyResize.bind(viewer));
+            const resize_observer = new MutationObserver(() => viewer.notifyResize.call(viewer));
```
zemeolotu: We probably don't need this block of code. Phosphors `onUpdateRequest` lifecycle method can probably replace this

texodus: Noted - I've removed this from the minimal-theme branch

PR #844, opened 2019-12-13
Node.js `table` unpin

### LINKED ISSUE #841: How to delete a hosted table?
## Support Question

I am trying to `delete` a table hosted by a `WebSocketServer` in NodeJS, and I get the error `Table still has contexts - refusing to delete.`. In my case, we seek to dynamically create and destroy tables of different names on the server. Later on, a new table with the same name might appear (with a different schema).

As far as I can tell (I'm new here...) a view is created here which can never be deleted: https://github.com/finos/perspective/blob/a912c9f23f1cab3e6ce97cdf65c6fe48f9ae9756/packages/perspective/src/js/perspective.node.js#L266

However... the view isn't returned so I'm not sure how to handle this case. The class doesn't seem to have an `unhost_table` method.

Is there anything better than using `clear` on the table and leaving it hosted in the server under that name?

texodus:
Fixes #841 

* Removes the empty `view()` which kept hosted `table` alive.  `<perspective-viewer>` no longer deletes it's `table` when it is [itself deleted](https://github.com/finos/perspective/blob/eaabd8b39349fc39cb23f4891b56430fa8fec993/packages/perspective-viewer/src/js/viewer/perspective_element.js#L458), so this is not default deleted (though it can still be explicitly deleted remotely - probably something to change later).
* Adds `on_delete` callbacks for both, unhosting objects when they are deleted.
* Adds `eject_table()` and `eject_view()` methods.

### DIFF
--- a/packages/perspective/src/js/perspective.node.js
+++ b/packages/perspective/src/js/perspective.node.js
@@ -263,27 +263,58 @@ class WebSocketServer extends Server {
         }
     }
 
-    /**
-     * Expose a Perspective table through the WebSocket, allowing it to be
-     * accessed by a unique name.
-     *
-     * @param {String} name
-     * @param {Perspective.table} table
-     */
-    host_table(name, table) {
-        this._tables[name] = table;
-        table.view({columns: []});
+    _host(cache, name, input) {
+        if (cache[name] !== undefined) {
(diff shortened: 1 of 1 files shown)


andyferris:
Awesome, thanks @texodus!

PR #842, opened 2019-12-12
Add `selectable` attribute to `perspective-viewer`

zemeolotu:
DO NOT MERGE YET! Waiting on a couple of fixes before merging

Adds a new 'selectable' to the perspective-viewer api. It needs to be implemented by the plugins and is currently only implemented by `perspective-viewer-hypergrid`.

In `perspective-viewer-hypergrid`, it enables row selection and maintains selection when row move about.

PR #845, opened 2019-12-14
Theme `material-dense`

texodus:
Adds a dense Material theme variant.  Also changes:

* Fixed side-effect order in `toggleConfig()` method to correctly set state when button is quicklly pressed (regression).
* Fixed `perspective-viewer-hypergrid` issue which caused console error (from #842).
* Adds `--hypergrid-row--height` style property.
* Adds `throttle` attribute, which overrides `perspective-viewer`'s default adaptive framerate.
* Reworked `perspective-phosphor` maximize functionality from toolbar to context menu using Phosphor's. `"single-document"` mode.
* Fixed a yarn version incompatibility (from #839)
* Fixed a phosphor error message due to over-aggressive default `faux-hypergrid` event binding.
* Made CMake's log output a little less brutal.

Old:

Screen Shot 2019-12-14 at 4 21 30 PM

New:

Screen Shot 2019-12-14 at 4 24 33 PM


texodus:
Reviewed offline with @zemeolotu

PR #852, opened 2019-12-18
Update versioning script for Python

sc1f:
This PR changes the versioning script to use the correct value for `rc`, as well as adding the changed Python version files to Git so that `lerna publish` can pick them up.


texodus:
Thanks for the PR!

PR #859, opened 2020-01-06
Add benchmark suite for Python, Refactor module loading for environments where C++ cannot be built

sc1f:
This PR adds a primitive benchmark suite for `perspective-python`, which follows the convention set by the Javascript benchmark suite - it runs operations for Table construction from different datasets, view construction with different pivot settings, and serializing data in all formats from each of those views. Benchmark results are written to a `perspective.Table` which is hosted via Tornado when the suite finishes (or when Ctrl-C is used to exit the script).

Additionally, this PR refactors the module loading semantic for `perspective-python`: when the C++ runtime is not present, a reasonable error message is logged, and symbols dependent on the C++ binding will not be exported in the module. `PerspectiveWidget` in client mode becomes the only usable symbol, and `Table`, `PerspectiveManager`, `PerspectiveTornadoHandler`, and `PerspectiveViewer` are not exported at all. A new test module has been added to assert correct behavior in this case.

The error message in question:

Screen Shot 2020-01-06 at 2 59 40 PM

Finally, minor refactors were made in `to_format` for performance enhancements.

### DIFF
--- a/python/perspective/bench/bench.py
+++ b/python/perspective/bench/bench.py
@@ -6,14 +6,28 @@
 # the Apache License 2.0.  The full license can be found in the LICENSE file.
 #
 import logging
+import json
 import os
 import sys
 import signal
+import tornado
 from time import perf_counter
 sys.path.insert(1, os.path.join(os.path.dirname(__file__), '..'))
-from perspective import Table  # noqa: E402
+from perspective import Table, PerspectiveManager, PerspectiveTornadoHandler  # noqa: E402
 
 logging.basicConfig(level=logging.INFO)
+
+
+class BenchmarkTornadoHandler(tornado.web.RequestHandler):
+    """Host the results of the benchmark suite over a websocket."""
+
+    def set_default_headers(self):
+        self.set_header("Access-Control-Allow-Origin", "*")
+        self.set_header("Access-Control-Allow-Headers", "x-requested-with")
+        self.set_header('Access-Control-Allow-Methods', 'POST, GET, OPTIONS')
+
+    def get(self):
+        self.render("benchmark.html")
 
 
 class Benchmark(object):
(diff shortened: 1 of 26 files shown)


python/perspective/bench/bench.py
```diff
@@ -0,0 +1,200 @@
...
+
+class Benchmark(object):
+    """A single Benchmark function. Use as a wrapper for stateless lambdas
+    with no parameters.
+
+    Example:
+        >>> func = Benchmark(lambda: self._view.to_records(), meta={
+                "name": "to_records",
+                "group": "view"
+            })
```
texodus: Are these for use as a decorator?  

```python
@Benchmark
def _to_records_test():
    self._view.to_records();
```

If so I think the signature of `__init__` is wrong - it wouldn't take a `func`.  I see `Benchmark` is not used as a decorator though here or in the perspective.py benchmarks themselves, but they are referenced as decorator in the docs.

python/perspective/bench/bench.py
```diff
@@ -0,0 +1,200 @@
...
+
+    def write_results(self):
+        if self._table is None:
+            return
+        logging.info("Writing results to `benchmark.arrow`")
+        arrow_path = os.path.join(os.path.dirname(__file__), "benchmark.arrow")
+        with open(arrow_path, "wb") as file:
+            arrow = self._table.view().to_arrow()
+            file.write(arrow)
+        self._WROTE_RESULTS = True
```
texodus: It would be nice if this method could incrementally write to an existing arrow so we can locally archive multiple versions, by checking to see if an `benchmark.arrow` file exists on load.  It would also enable cross-language benchmarks in the same archive ..

python/perspective/bench/bench.py
sc1f: The docs are a little out of sync; `Benchmark` wraps a parameterless lambda and a metadata dictionary, which is a simpler implementation I went with instead of decorators.

PR #884, opened 2020-01-28
updates and fixes for windows build

timkpaine:
A few updates and fixes for the windows build of perspective-python
- build dlls and pyd file
- set prefix for assets to `lib` for symmetry
- inline arrow dlls and tbb dll
- omit tests using missing platform functions
- swap slashes in `setup.py` for windows paths
- change `.appveyor.yml` to match travis

Still unresolved:
https://bugs.python.org/issue36439


scripts/script_utils.js
```diff
@@ -274,15 +275,28 @@ function run_suite(tests) {
...
-run_suite([
-    [resolve`a/b/c`, `${process.cwd()}/a/b/c`],
-    [resolve`${__dirname}/../cpp/perspective`, `${process.cwd()}/cpp/perspective`],
-    [resolve`${__dirname}/../python/perspective/dist`, _path.resolve(__dirname, "..", "python", "perspective", "dist")],
-    [resolve`${__dirname}/../cpp/perspective`, _path.resolve(__dirname, "..", "cpp", "perspective")],
-    [resolve`${__dirname}/../cmake`, _path.resolve(__dirname, "..", "cmake")],
-    [resolve`${resolve`${__dirname}/../python/perspective/dist`}/cmake`, _path.resolve(_path.resolve(__dirname, "..", "python", "perspective", "dist"), "cmake")],
-    [resolve`${resolve`${__dirname}/../python/perspective/dist`}/obj`, _path.resolve(_path.resolve(__dirname, "..", "python", "perspective", "dist"), "obj")]
-]);
+if (isWin){
```
timkpaine: @texodus can you take a look at this, the comments seem to suggest that this was supposed to work but it does not on windows

cpp/perspective/CMakeLists.txt
```diff
@@ -197,6 +196,10 @@ message(WARNING "${BUILD_MESSAGE}")
 #######################
 include_directories("${CMAKE_SOURCE_DIR}/src/include")
 
+if(NOT WIN32)
+	set(CMAKE_CXX_FLAGS_RELEASE "-O3 -DNDEBUG")
```
timkpaine: @sc1f note this

cpp/perspective/CMakeLists.txt
```diff
@@ -264,16 +267,31 @@ elseif(PSP_CPP_BUILD OR PSP_PYTHON_BUILD)
...
 	# VANILLA CPP BUILD #
 	#####################
-	if(CMAKE_BUILD_TYPE_LOWER STREQUAL debug)
-		set(OPT_FLAGS " \
-			-O0 \
-			-g3 \
-			")
+	if(WIN32)
+		if(CMAKE_BUILD_TYPE_LOWER STREQUAL debug)
+			set(OPT_FLAGS " \
```
timkpaine: @sc1f we should set  debug flags (see prev comment) instead of having this opt flags thing

cpp/perspective/CMakeLists.txt
```diff
@@ -548,7 +564,9 @@ elseif(PSP_CPP_BUILD OR PSP_PYTHON_BUILD)
             set(module_install_rpath "@loader_path/")
 	    else()
             set(module_install_rpath "\$ORIGIN")
-	    endif()
+		endif()
+	else()
+		set(CMAKE_SHARED_LIBRARY_PREFIX lib)
```
timkpaine: @sc1f we should consider not prefixing with lib, and setting the prefixes to "" (this will mean we need to change python imports from `libbinding` to just `binding`, or preferable `_binding`)

cpp/perspective/CMakeLists.txt
```diff
@@ -562,28 +580,45 @@ elseif(PSP_CPP_BUILD OR PSP_PYTHON_BUILD)
...
 		target_link_libraries(binding tbb)
 
+		target_link_libraries(binding psp)
+
 		add_custom_command(TARGET psp POST_BUILD COMMAND ${CMAKE_COMMAND} -E copy $<TARGET_FILE:psp> ${PSP_PYTHON_SRC}/table/)
 		add_custom_command(TARGET binding POST_BUILD COMMAND ${CMAKE_COMMAND} -E copy $<TARGET_FILE:binding> ${PSP_PYTHON_SRC}/table/)
+
+		if(WIN32)
+			# inline arrow dlls
+			file(GLOB ARROW_DLLS "${PYTHON_PYARROW_LIBRARY_DIR}/*.dll")
```
timkpaine: @sc1f @texodus  open to suggestions to get around copying all arrow dlls into our directory. In a "production" environment with a custom arrow build you won't have to worry about this, but with the packaged arrow wheel bringing a ton of stuff with it, we need to ensure all the links resolve.

cpp/perspective/src/include/perspective/scalar.h
```diff
@@ -207,49 +207,49 @@ template <>
 t_tscalar t_tscalar::coerce_numeric<bool>() const;
 
 template <>
-std::int64_t t_tscalar::get() const;
+PERSPECTIVE_EXPORT std::int64_t t_tscalar::get() const;
```
timkpaine: @sc1f i assume all these specializations are necessary?

cpp/perspective/src/include/perspective/sym_table.h
```diff
@@ -31,8 +31,8 @@ class t_symtable {
     t_mapping m_mapping;
 };
 
-const char* get_interned_cstr(const char* s);
-t_tscalar get_interned_tscalar(const char* s);
-t_tscalar get_interned_tscalar(const t_tscalar& s);
+PERSPECTIVE_EXPORT const char* get_interned_cstr(const char* s);
```
timkpaine: @sc1f this is currently being exposed as part of `libpsp`'s public interface (it is used in the python cpp code). should this be public?

python/perspective/package.json
```diff
@@ -3,7 +3,8 @@
   "name": "perspective-python.node",
   "version": "0.4.1",
   "dependencies": {
-    "@finos/perspective": "^0.4.1"
+    "@finos/perspective": "^0.4.1",
+    "zerorpc": "^0.9.8"
```
timkpaine: @texodus please confirm

python/perspective/perspective/include/perspective/python/numpy.h
```diff
@@ -29,7 +39,7 @@ namespace numpy {
     /**
      * NumpyLoader fast-tracks the loading of Numpy arrays into Perspective, utilizing memcpy whenever possible.
      */
-    class PERSPECTIVE_EXPORT NumpyLoader {
+    class PERSPECTIVE_BINDING_EXPORT NumpyLoader {
```
timkpaine: @sc1f is a class necessary?

scripts/test_python.js
```diff
@@ -31,11 +31,12 @@ try {
         const python_path = resolve`${__dirname}/../python/perspective`;
         // Tests for client mode (when C++ assets are not present) require
         // a new runtime.
-        execute`cd ${python_path} && TZ=UTC ${PYTHON} -m pytest \
+        process.env.TZ = "UTC";
```
timkpaine: @sc1f note for os-agnostic

cpp/perspective/src/include/perspective/scalar.h
sc1f: Yeah, they all need to be explicitly specialized.

python/perspective/perspective/include/perspective/python/numpy.h
sc1f: it follows the structure set by arrow_loader, and allows us to abstract all the numpy-specific stuff since its' all state-dependent

python/perspective/package.json
texodus: At this point I see honestly zero reason to maintain this feature - `perspective-python` is better in every way.

cpp/perspective/CMakeLists.txt
texodus: I don't understand what the requirement being solved by copying is, nor why a ton of stuff being brought is the instigator, so I'm not really in a position to offer constructive advice here.  Copying is bad of course.

PR #897, opened 2020-02-02
update copyright, update python makefile disclaimer

timkpaine:


python/perspective/Makefile
```diff
@@ -1,3 +1,18 @@
+################################################################################
```
timkpaine: the only non-year change

texodus:
Thanks for the PR!  I think "Copyright 2017" means "year of first publication of the work or compilation of derivative works", not "year most recently modified", in which case this does not need to be updated. [U.S.C.A. Title 17 § 401.b.2.](https://www.copyright.gov/title17/92chap4.html).

PR #903, opened 2020-02-04
Make `perspective-viewer-hypergrid` selection state save/restore compatible

zemeolotu:
This PR makes the `perspective-viewer-hypergrid` selection state save/restore compatible. 

If a row is selected and `save` is called on the `perspective-viewer`, the config returned will have `plugin_config` with the selected row index. Similarly to restore a selection state, you can call `restore` with the selected row index set on the `plugin_config` to set the selection state of the grid e.g.

`viewer.restore({plugin_config: {selected: "row_index"}})`

This PR also binds the `perspective-select` event to the `fin-row-selection-changed` event instead of the `cellClicked` callback. This allows selection events from arrow keys to be captured.
\
selection

### DIFF
--- a/packages/perspective-viewer-hypergrid/src/js/PerspectiveDataModel.js
+++ b/packages/perspective-viewer-hypergrid/src/js/PerspectiveDataModel.js
@@ -175,6 +175,22 @@ export default require("datasaur-local").extend("PerspectiveDataModel", {
         }
     },
 
+    getSelectedRowID() {
+        if (this._grid.selectionModel.hasRowSelections()) {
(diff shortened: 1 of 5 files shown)


timkpaine:
@sc1f is this something that should also be exposed in python?

texodus:
@timkpaine This can't currently be implemented as `save()`/`restore()` in python just serializes the `Widget` state, and `plugin-config` is only calculated when `save()` is called on the `<perspective-viewer>` element itself, e.g. it does not emit change events or re-calculate the `View`.

PR #906, opened 2020-02-05
Bump core-js to v3.6.4 and Babel to 7.8.4

sebinsua:
We should upgrade `core-js@2` to `core-js@3` because the (1) the [next major release of Babel will do so](https://github.com/babel/babel/issues/10746#issuecomment-573401724), and (2) there are [many V8 de-optimizations in `core-js@2` that have been fixed in `core-js@3`](https://github.com/babel/babel/issues/10746#issuecomment-573402372).

**Note**: Half-expecting there to be some kind of build issue, so opening a PR to see what's what.


sebinsua:
@texodus It seems to pass the tests.

Screenshot 2020-02-05 at 14 30 35

package.json
```diff
@@ -17,19 +17,15 @@
         "docs",
         "python/perspective"
     ],
-    "repository": {
-        "type": "git",
-        "url": "https://github.com/finos/perspective"
-    },
```
sebinsua: Something weird about this diff since this is [not where this property is within the current `package.json`](https://github.com/finos/perspective/blob/86cd2c1f9135db586ed0a8f74bea1c7f50759dde/package.json#L3-L6).

texodus:
@sebinsua Thanks for the PR!

I made a [`<perspective-workspace>` chart of the bundle size impact on bl.ocks.org](https://bl.ocks.org/texodus/fd312e621e00edacbd932b754aea4cd8), which shows the impact of this PR on Perspective's bundle size is not significant, but measurable - `perspective.wasm.worker.js`, the WebWorker runtime for the WASM engine, seems particularly badly hit.

Screen Shot 2020-02-05 at 4 48 56 PM

texodus:
Another view of this change via the upcoming `flatten()`

Screen Shot 2020-02-05 at 7 31 06 PM

PR #908, opened 2020-02-06
Remove yarn dependency duplication.

texodus:
Cleans up duplicate `yarn.lock` entries by tightening some dependency ranges, as discussed in #906

### DIFF
--- a/package.json
+++ b/package.json
@@ -11,6 +11,9 @@
             "internal": "Internal"
         }
     },
+    "engines" : {
+        "node" : ">=8.12.0"
+    },
     "workspaces": [
         "packages/*",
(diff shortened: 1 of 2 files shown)


sebinsua:
Btw [there is a useful package to get visibility of this kind of thing and which can automate fixing it](https://github.com/atlassian/yarn-deduplicate). Maybe you've used it before - just mentioning it because it's a lot easier than needing to do this manually.

texodus:
Thanks @sebinsua !  The key thing missing here was a `node.js` version constraint to our dev dependencies, as emscripten and our Docker runtimes are somewhat inflexibly tied to 8.x

PR #911, opened 2020-02-08
Fix bug where `perspective-viewer` doesn't resize

zemeolotu:


packages/perspective-viewer/src/js/viewer/dom_element.js
```diff
@@ -308,7 +308,7 @@ export class DomElement extends PerspectiveElement {
                     this._persisted_side_panel_width = this._side_panel.style.width;
                     this._side_panel.style.width = old || "";
                     app.classList.add("columns_horizontal");
-                    return true;
+                    return false;
```
texodus: This was supposed to return `false` because `_pre_resize()` above it itself calls `this.plugin.resize()` [here](https://github.com/finos/perspective/blob/8fc6672861ed0fa9f521edb2dd4af62f6370ac55/packages/perspective-viewer/src/js/viewer/action_element.js#L75), thus preventing resize logic form being applied twice.  Clearly that's not working correctly but I believe returning `true` here will render the plugin twice?

texodus:
Thanks for the PR!  I believe this is fixed in #914

packages/perspective-viewer/src/js/viewer/dom_element.js
zemeolotu: Gotcha, however `pre_resize()` isn't actually called in that block. Is the right fix to just call `pre_resize` then?

texodus:
Thank for the PR!  I misread this in review, apologies!

PR #916, opened 2020-02-11
Added Azure compatible reporting for tests

texodus:
Adds `jest` and `pytest` `junit` output, and publish these results to Azure Devops for [pretty build analytics](https://dev.azure.com/finosfoundation/perspective/_build/results?buildId=18&view=ms.vss-test-web.build-test-results-tab).

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -27,6 +27,12 @@ jobs:
       env:
         PSP_DOCKER: 1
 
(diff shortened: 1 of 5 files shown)


timkpaine:
nice! https://dev.azure.com/finosfoundation/perspective/_build/results?buildId=21&view=codecoverage-tab

PR #917, opened 2020-02-11
Upgrade Apache Arrow to 0.16.0

sc1f:
This PR upgrades our dependency on [Apache Arrow](https://arrow.apache.org/release/0.16.0.html) to version `0.16.0`. Docker images will need to be rebuilt for this patch to pass CI.


timkpaine:
@sc1f can you delete the requirements files? I dont think we use them anymore (requirements are inlined in the setup.py)

texodus:
Thanks for the PR!

Discussed offline, it is not possible to upgrade until the from-source builds of `Arrow` for our docker builds are removed, which I've been told is not possible itself.  Hence, let's close this and re-visit when there is a working `pyarrow` to build on Azure.

PR #920, opened 2020-02-12
Adds test coverage reporting for`@finos/perspective`

texodus:
Due to poor interplay between Webpack and Puppeteer, this will unfortunately only work for `@finos/perspective` for now, but better this than nothing!  In order to get meaningful module name reporting here, I've also gone ahead and removed the `perspective.node.js` webpack step entirely, and `node` part of this module now references the babel output directly.

[Example here](https://dev.azure.com/finosfoundation/perspective/_build/results?buildId=32&view=codecoverage-tab)

Also removes the FINOS/Operating badge from `README.md`, as it links to an access-restricted (dead?) Confluence page.


packages/perspective-workspace/src/js/workspace/workspace.js
```diff
@@ -411,15 +411,16 @@ export class PerspectiveWorkspace extends DiscreteSplitPanel {
             await widget.toggleConfig();
         }
 
+        this.detailPanel.close();
```
zemeolotu: I think this will introduce a bug, where the detail panel is closed when there's already a global filter widget

texodus: Sorry, this was a commit from another PR.  Good measure we have code review!

Issue #921, opened 2020-02-12
Incorrect semver for front end assets causing issues with voila


timkpaine:
These versions are stale (they are not auto-incremented) which can cause issues with Voila loading the wrong version. 

https://github.com/finos/perspective/blob/master/python/perspective/perspective/widget/widget.py#L144

texodus:
It also seems like the `bumpversion` script did not pick up `0.4.3` nor `0.4.2.post1`.

timkpaine:
Since we have the package.json, can we just ship that and then read the version out of it?

timkpaine:
Then lerna can manage all the versioning and we can delete bumpversion

sc1f:
That sounds much nicer. I'll work it in.

timkpaine:
It's generally a bad idea to load up random files at runtime, but we would be loading up the _version.py file anyway

PR #923, opened 2020-02-13
Fix for #921: unifies versioning between JS and Python libraries

sc1f:
This PR fixes our long-standing issue with Python versioning, Bumpversion, and assorted versioning quirks. 

By replacing `bumpversion` with a task run as part of Lerna's [lifecycle script](https://github.com/lerna/lerna/tree/master/commands/version#lifecycle-scripts), the semantics around maintaining two separate versioning configurations are removed.

Instead, the `version` script reads the version of `package.json` in the Python folder, which (at the time the lifecycle script runs) will have been updated to the new version specified in `lerna publish` or `lerna version`. 

`_version.py` will have the version string written into it exactly as specified in `package.json`, and `PerspectiveWidget` will look for the **EXACT** version as specified in `_version.py` when it is loaded in Jupyterlab. 

Thus, a distribution with _mismatched_ versions between Python and Javascript will break the Jupyterlab widget, and a distribution that does _not_ use `lerna version` or `lerna publish` as one of its steps will **NOT** have the version number reflected in `_version.py`. In short, Python distributions now must happen in lock-step with Javascript distributions, and with the exact same version string between the two distributions.

### Changelog
- Removes bumpversion and replaces it with a script that reads `version` from `package.json`
- `PerspectiveWidget`  now reads `__version__` from `_version.py`
- `docs` task and Sphinx config now read `__version__`
- `setup.py` now reads `__version__`


python/perspective/perspective/widget/widget.py
```diff
@@ -141,10 +142,10 @@ class PerspectiveWidget(Widget, PerspectiveViewer):
     # Required by ipywidgets for proper registration of the backend
     _model_name = Unicode('PerspectiveModel').tag(sync=True)
     _model_module = Unicode('@finos/perspective-jupyterlab').tag(sync=True)
-    _model_module_version = Unicode('^0.4.0-rc.2').tag(sync=True)
+    _model_module_version = Unicode(__version__).tag(sync=True)
```
sc1f: We need a tilde constraint here that allows for `rc`, `alpha`, etc.

timkpaine:
@texodus this is ready to go!

scripts/docs.js
```diff
@@ -12,13 +12,9 @@ const {execute} = require("./script_utils.js");
 try {
     execute`mkdirp docs/build docs/obj`;
     const project = process.env.PSP_PROJECT;
-    if (!project || project === "js") {
+    if (!project || project === "js" || project === "python") {
```
texodus: prob don't need js anymore here either due to #912

PR #939, opened 2020-02-21
Refactor `gnode` and `gnode_state`, remove C++ test suite

sc1f:
This PR removes the C++ test suite and cleans up the `t_gnode` and `t_gnode_state` classes, which handle processing of new `Table`s and calls to `update` before notifying all `View`s registered to the `Table`. No functionality changes have been implemented in this PR, simply refactoring the internals to make it easier to reason about.

- Add `t_process_state` struct to keep track of internal state in `_process_table`
- Remove all dead/unimplemented methods from gnode/gnode state
- Rename methods and members in gnode/gnode state for clarity
- Remove C++ test suite and C++-only build
- Add version support for Python benchmarks (creates venv and pulls from pip)


timkpaine:
@sc1f https://github.com/finos/perspective/blob/master/python/perspective/setup.py#L124

sc1f:
> @sc1f https://github.com/finos/perspective/blob/master/python/perspective/setup.py#L124

fixed

cpp/perspective/src/cpp/gnode.cpp
```diff
@@ -345,264 +236,150 @@ t_gnode::_process_table() {
...
                 default: { PSP_COMPLAIN_AND_ABORT("Unsupported column dtype"); }
             }
         }
 #ifdef PSP_PARALLEL_FOR
     );
 #endif
-    std::shared_ptr<t_data_table> flattened_masked
-        = mask.count() == flattened->size() ? flattened : flattened->clone(mask);
-    PSP_GNODE_VERIFY_TABLE(flattened_masked);
-#ifdef PSP_GNODE_VERIFY
```
texodus: Here too (I think this is removal not refactor?)

cpp/perspective/src/cpp/gnode.cpp
```diff
@@ -345,264 +236,150 @@ t_gnode::_process_table() {
...
-    {
-        auto updated_table = get_table();
-        PSP_GNODE_VERIFY_TABLE(updated_table);
-    }
-#endif
-    m_state->update_history(flattened_masked.get());
-#ifdef PSP_GNODE_VERIFY
```
texodus: Here

cpp/perspective/src/cpp/gnode.cpp
```diff
@@ -345,264 +236,150 @@ t_gnode::_process_table() {
...
+        m_gstate->update_master_table(flattened.get());
         _update_contexts_from_state(*flattened);
         m_oports[PSP_PORT_FLATTENED]->set_table(flattened);
         release_inputs();
         release_outputs();
-        
-#ifdef PSP_GNODE_VERIFY
-        auto stable = get_table();
-        PSP_GNODE_VERIFY_TABLE(stable);
-#endif
```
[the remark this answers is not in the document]
sc1f: As far as I can tell `PSP_GNODE_VERIFY` isn't actually defined in any of our builds. I agree with triggering these under PSP_DEBUG though - that would be helpful.

cpp/perspective/src/cpp/gnode.cpp
sc1f: I removed the call to verify table but refactored the structure of the `flattened_masked` assignment.

cpp/perspective/src/cpp/gnode.cpp
sc1f: Removed the verify table here as well, yes

PR #943, opened 2020-02-22
Deal with np.int_ on Windows, handle missing __INDEX__

### LINKED ISSUE #926: Numpy int defaults to int32 on Windows, int64 on *nix
this is causing some of the issues on windows:
https://github.com/finos/perspective/blob/master/python/perspective/perspective/src/utils.cpp#L23
https://software.intel.com/en-us/articles/size-of-long-integer-type-on-different-architecture-and-os

timkpaine:
- [x] fixes #926 
- [x] fixes missing \_\_INDEX\_\_ error during np update
- [x] reenable windows build on azure
- [x] fix carriage return inconsistency  in `to_csv`


texodus:
Thanks for the PR!

Issue #952, opened 2020-02-26
Prune bench from python manifest


timkpaine:
We are currently listing a python package called `bench`, this should be omitted

Issue #937, opened 2020-02-21
Failed to import C++ bindings for Perspective


alexeykulakov:
## Bug Report
I have successfully install perspective-python package with all dependencies (NumPy and PyArrow)
Then I get the warning message when import perspective
> CRITICAL:root:Failed to import C++ bindings for Perspective probably as it could not be built for your architecture.
> You can still use `PerspectiveWidget` in client mode using JupyterLab.

So I can't use Table and PerspectiveManager classes.
Command `from perspective import Table, PerspectiveManager` gives the result:

> File "<stdin>", line 1, in <module>
> ImportError: cannot import name 'Table' from 'perspective' (/home/akulakov/miniconda3/envs/finos/lib/python3.7/site-packages/perspective/__init__.py)

### Steps to Reproduce:
 1. conda create -n finos python=3.7.0
 2. conda activate finos
 3. python -m pip install numpy
 4. python -m pip install pyarrow==0.15.1
 5. python -m pip install perspective-python
No errors at this time
 6. exec python: from perspective import Table, PerspectiveManager

### Expected Result:
import packages with no errors

### Environment:
OS: Ubuntu 19.10
Virtualenv: conda
Python: 3.7.0

timkpaine:
@alexeykulakov most likely culprit is that it built against some system libraries instead of what is in the conda env. There is a conda wheel in flight but in the mean time can you `ldd` libbinding and libpsp in perspective/table and see what they say

alexeykulakov:
> @alexeykulakov most likely culprit is that it built against some system libraries instead of what is in the conda env. There is a conda wheel in flight but in the mean time can you `ldd` libbinding and libpsp in perspective/table and see what they say

I've tried another virtual environment `virtualenv venv` and got the same result.
Would you please to explain how to use ldd?

timkpaine:
@alexeykulakov 
First can you try to pip install perspective with `-v` so we can see logs:
`python -m pip install perspective-python -v`

If that doesn't have any errors run
```bash
cd `python3 -c "import perspective, os.path; print(os.path.dirname(perspective.__file__))"`/table && ldd libpsp.so
```

alexeykulakov:
@timkpaine 
1) As I could understand installation finished with no errors.
Full log here https://raw.githubusercontent.com/alexeykulakov/temp/master/install_log.txt

2) command return the same message
> CRITICAL:root:Failed to import C++ bindings for Perspective probably as it could not be built for your architecture (check install logs for more details).
[...]

PR #953, opened 2020-02-27
Bugfix sdist

timkpaine:
This should fix:
- #937 

May also help fix:
- #928 

Fixes (unrelated):
- #952 

Don't know how possible it is to test this so i've added an extensive comment about the two library copies

### DIFF
--- a/cpp/perspective/CMakeLists.txt
+++ b/cpp/perspective/CMakeLists.txt
@@ -614,6 +614,10 @@ elseif(PSP_CPP_BUILD OR PSP_PYTHON_BUILD)
 
 		target_link_libraries(binding psp)
 
+		# The compiled libraries will be put in CMAKE_LIBRARY_OUTPUT_DIRECTORY by default. In the
+		# setup.py file, we designate this to be in the build/lib.<platform> directory. However,
+		# since we want to be able to test perspective in-source, we also copy the libraries into
+		# the source folder. These two commands do that.
 		add_custom_command(TARGET psp POST_BUILD COMMAND ${CMAKE_COMMAND} -E copy $<TARGET_FILE:psp> ${PSP_PYTHON_SRC}/table/)
 		add_custom_command(TARGET binding POST_BUILD COMMAND ${CMAKE_COMMAND} -E copy $<TARGET_FILE:binding> ${PSP_PYTHON_SRC}/table/)
 
--- a/python/perspective/MANIFEST.in
+++ b/python/perspective/MANIFEST.in
@@ -24,6 +24,10 @@ prune docs/build
 prune docs/gh-pages
 prune docs/dist
 
+# don't ship benchmark
+prune bench
+
+
 # Use package.json for versioning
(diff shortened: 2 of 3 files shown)


ceball:
> any suggestions on how to assert this behavior in a test in the future?

@texodus I think the overall best test of these kinds of things (in terms of coverage and practicality) is to actually install and test the package as end users would do. That sounds painful, but I would say that @timkpaine's conda package will already help a lot there. 

At the risk of saying things everyone already knows...

The package build process (i.e. running just "conda build" once there is a recipe) results in:
  * creation of a clean (isolated) build environment
  * building of package in that build environment
  * creation of clean (isolated) installation environment
  * installation of package built above into clean installation environment
  * tests of that installed package (at least: does it import? but could be more, e.g. unit tests)

"Isolated" means as isolated as reasonably possible (but not docker levels of isolation).

The conda build process above will happen automatically on conda-forge (on all of the major platforms) whenever a new release of the package is uploaded to pypi - a PR at conda-forge will be opened automatically for the perspective package maintainers to review.

So that seems pretty good to me. Thanks Tim :)

ceball:
I also have some related thoughts. Tacking them onto the end of a closed PR isn't the best idea, but I just happen to be here now.

The above process isn't perfect. E.g. it could result in having to upload a new release after putting out a bad one, which would be annoying. To avoid that, the above workflow could be done first with a dev or pre-release tag, i.e. pushing pre-releases (e.g. 0.1.2rc1 or whatever) to pypi and making sure they successfully go through the above process before pushing a release tag. Or, perspective's own CI could be set up to build the conda package locally from time to time.

It might also be worth considering having automated uploading to pypi from CI, e.g. based on a git tag.

Also, the above is mainly conda focused. Although the process will effectively be testing some parts of the python distribution and installation process, it's not a full test from the point of view of someone who will be doing "pip install perspective-python". However, the same kinds of ideas (e.g. testing what users will be installing) are covered in the python world via projects like tox (to the extent it's possible to cover "system level dependencies" with pip).

I'd be happy to help out with any of the above (will talk to Tim).

texodus:
Is it possible to "install and test the package as a end users would do" without first _disting it to conda-forge_?  I have no doubt this is the best way - but it doesn't prevent us from regressing this build support if we can only install & test the last version we disted?

ceball:
Yes, it is. My comment was too long :) 

> perspective's own CI could be set up to build the conda package locally from time to time.

This is what most projects I personally work on do, rather than relying only on conda-forge. I.e. we keep the recipe (yaml file) in our github repo, and can then just run "conda build" whenever we want (locally as developers, and on CI). 

We don't usually do it on every commit, just when we push a dev tag, becase generally the projects I'm talking about are pure python with quick test times, so the overhead of full package building is not worth it. But that might be different for perspective.

timkpaine:
We already do this for perspective, just only on one specific platform after doing an in-place build. Basically need to modify it to undo the in-place prior to running sdist

PR #961, opened 2020-03-04
Fully clear queued updates before adding a new computed column

sc1f:
When a computed column is added, a new `Table` is created with the computed column, and Perspective internally invalidates and cleans up the old `Table`. 

However, if updates were queued on the old `Table` and then called for process, the state between the old `Table` and the new `Table` (both of which have the computed column) may fall out of sync and cause errors at random.

This PR adds a call to `call_process` before creating the computed column, ensuring that all queued updates are processed before the old `Table` is invalidated.

It also reverses a change on the `streaming.html` example, by reducing the tick rate to something approaching sanity and adding a `date` column for full type coverage.


texodus:
Fixes #972

PR #970, opened 2020-03-11
update dependencies to jupyterlab 2.0, phosphor -> lumino

timkpaine:
Screen Shot 2020-03-11 at 10 40 04 AM


timkpaine:
@zemeolotu please verify workspaces has no weird bugs (the package change should be name only, its the same version)

texodus:
Thanks for the PR!

PR #975, opened 2020-03-14
Reference python objects directly in perspective tables

timkpaine:
Adds support for storing python objects in perspective tables by consolidating `DTYPE_OBJECT` and `DTYPE_PTR` into just `DTYPE_OBJECT`.

Python objects are stored by pointer (e.g. `PyObject `*) as `uint64_t` then materialized from the underlying scalar, e.g. if `scalar.m_type == DTYPE_OBJECT` we cast the `uint64_t` back to a `PyObject *` 

Also added is support for value extraction via the `_psp_dtype_` and `_psp_repr_` magic methods, the presence of which indicates how perspective should treat the returned value, along with the ability to return arbitrary values (`__repr__` is limited to strings, so if you want to pull an integer out of your object you can define `_psp_dtype_` to return `int` and have `_psp_repr_` return an `int`).

Along these lines, type inference is slightly modified to support promotions, e.g. if you return a custom object that supports `__int__` or `__float__`, we'll look for those methods if you tell perspective to store as an integer or floating type. 

Example behaviors in the [test file](https://github.com/finos/perspective/blob/1805e686fccdc9ae4737e1570ab261e4bc3f3172/python/perspective/perspective/tests/table/test_table_object.py)

Some vestigial code was removed as well, and some hardcoded `py::object` were converted to `t_val` for consistency.

Todos:
- [x] `id(val)` won't increment the reference counter, so it will be possible for a scalar to store a pointer that has been cleaned up. When we store a `DTYPE_OBJECT`, we'll need to make sure to `INCREF` the object.
- [x] Add documentation about `_psp_dtype_` and `_psp_repr_`
- [x] Decide whether default behavior should be to extract (current behavior) or store reference
- [x] turn TBB back on


timkpaine:
@sc1f can you take a look and let me know your thoughts

python/perspective/perspective/table/_accessor.py
```diff
@@ -229,6 +235,10 @@ def marshal(self, cidx, ridx, dtype):
             if not isinstance(val, bool) and isinstance(val, _PerspectiveAccessor.INTEGER_TYPES):
                 # update float columns with either ints or floats
                 return float(val)
+        elif dtype == t_dtype.DTYPE_OBJECT:
+            # store id as pointer
+            print(id(val))
```
sc1f: remove print

cpp/perspective/src/cpp/view.cpp
```diff
@@ -465,6 +465,10 @@ View<CTX_T>::data_slice_to_arrow(
                 fields.push_back(::arrow::field(name, ::arrow::dictionary(::arrow::int32(), ::arrow::utf8())));
                 arr = arrow::string_col_to_dictionary_array(slice, cidx, stride, extents);
             } break;
+            case DTYPE_OBJECT: {
+                fields.push_back(::arrow::field(name, ::arrow::uint64()));
```
sc1f: This just serializes the uint64 pointers to arrow right? I assume there's no way to reconstitute the object after it goes into arrow/from an arrow (not that we want to, just curious)

python/perspective/perspective/tests/table/test_table_object.py
```diff
@@ -0,0 +1,204 @@
...
+    def _psp_dtype_(self):
+        return float
+
+    def __int__(self):
+        return int(self._value)
+
+    def __repr__(self):
+        return 'test'
+
+class TestTableObjectsExtract(object):
```
sc1f: Is there a way to test the "storing an object whose pointer has been cleaned" behaviour you mention in the PR notes?

cpp/perspective/src/cpp/view.cpp
timkpaine: need @texodus to weigh in. Ideally if the underlying object supported a serialization protocol we could turn it into json/protobuf/arrow/whatever and reconstitute it later, but the object would need to define that.

e.g. if a given library's object supported json really well, the python object could be serialized to json using the library's functionality, transmitted across the wire by psp in an arrow (just storing the json as bytes or whatever), then lets say it transmitted to js, it could then be reconstituted by the the library's js apis into their JS object.

python/perspective/perspective/tests/table/test_table_object.py
timkpaine: yes, we can explicitly `del` the object but we don't want to support this e.g. if we increment the reference count, python won't delete the object so long as psp has a copy of the pointer. We'll then need to make sure that when the table is destroyed/column destroyed/value destroyed/value overwritten that we decrement the reference count so python can deallocate it. 

`pybind`'s `object` class will do this automatically (and its `handle` class will not), but we're not using either of those internally so we'll need to use the underlying APIs (https://docs.python.org/3.7/c-api/refcounting.html)

timkpaine: I'll try to think of a good test around our use of reference counting, we can probably just write a test to make a table and count references (should be previous+1), and then destroy the table (which should subtract 1).
https://docs.python.org/dev/library/sys.html#sys.getrefcount

cpp/perspective/src/cpp/view.cpp
texodus: Lacking a specific user need for serializing python objects as Arrows, I would make this an error in `to_arrow()`, though it will be hard to prevent as `DTYPE_PTR` does not have a Perspective API type string like `"object"` yet.

timkpaine: @texodus part of this eliminates DTYPE_PTR and solely relies on DTYPE_OBJECT

timkpaine:
@texodus ready for you to review

cpp/perspective/src/cpp/column.cpp
```diff
@@ -444,33 +471,32 @@ t_column::clear(t_uindex idx, t_status status) {
...
-            std::uint16_t v = 0;
-            set_nth<std::uint16_t>(idx, v, status);
+            set_nth<std::uint16_t>(idx, 0, status);
         } break;
         case DTYPE_BOOL:
         case DTYPE_UINT8:
         case DTYPE_INT8: {
-            std::uint8_t v = 0;
-            set_nth<std::uint8_t>(idx, v, status);
+            set_nth<std::uint8_t>(idx, 0, status);
```
texodus, line 489: Any particular reason these need to be zeroed?  Just curious.

python/perspective/perspective/src/serialization.cpp
```diff
@@ -21,19 +21,6 @@ namespace binding {
...
-
-template <>
-t_val
-get_column_data(std::shared_ptr<t_data_table> table, const std::string& colname) {
-    py::array arr = py::array();
-    // TODO
-    // auto col = table->get_column(colname);
-    // for (auto idx = 0; idx < col->size(); ++idx) {
-    //     arr[idx] = py::cast(col->get_scalar(idx));
-    // }
```
texodus, line 33: Deadbeef

python/perspective/perspective/table/_accessor.py
```diff
@@ -191,13 +191,19 @@ def marshal(self, cidx, ridx, dtype):
         if val is None:
             return val
 
+        # if item implements custom repr, use it
+        if hasattr(val, "_psp_repr_"):
+            val = val._psp_repr_()
```
texodus, line 196: I am not a Python expert - is this canonical?  `__perspective_repr__`?

PR #979, opened 2020-03-17
Datagrid row selection and column resize

texodus:
Adds 2 new features to `@finos/perspective-viewer-datagrid`:

Row selection add support for the `selectable` attribute and support for `<perspective-workspace>`, plus a nifty sub-group selection style.  This fires the same `perspective-select` event as `@finos/perspective-viewer-hypergrid`, and will work as a drop-in replacement as a Global Filter in a `<perspective-workspace>`.

select

Column resize allows the user to override auto-sized columns by clicking on the column header separator and dragging left or right.  Double-click will remove this override and re-enable column auto-sizing.

Special care was given to performance for this feature.  `overflow: hidden` on a `<td>` causes incredible strain on the Blink layout engine it seems, so the `pd-cell-clip` CSS class which enables this overflow behavior (and `text-overflow:ellipsis`) are only applied when the override width is less than the auto-detected width.  Ergo, column auto-width is not disabled even when an override is applied;  when `pd-cell-clip` is applied, there is an ~30% drop in fps for this frame, though in practice this is usually not perceivable.

resize2


packages/perspective-viewer-datagrid/src/js/datagrid.js
```diff
@@ -101,27 +103,62 @@ function _tree_header(td, path, is_leaf, is_open) {
...
         for (let i = 0; i < path.length; i++) {
-            html += `<span style="margin-left:5px;margin-right:15px;border-left:1px solid #eee;height:19px"></span>`;
+            html += `<span class="pd-tree-group"></span>`;
         }
         td.innerHTML = `<div style="display:flex;align-items:stretch">${html}<span class="pd-row-header-icon">${icon}</span><span class="pd-group-name">${name}</span></div>`;
     }
     td.classList.add("pd-group-header");
 }
 
+async function getCellConfig(view, config, row_idx, col_idx) {
```
JHawk: This is prime for a refactor as some of the code exists in different forms in the other plugins.

texodus: agreed

PR #983, opened 2020-03-19
Expression-based Computed Columns

sc1f:
This PR is a large-scale overhaul of Perspective's computed columns API and UI. 

ezgif com-optimize

## API Overview

This PR **deprecates** the `Table`-based API for creating computed columns, and introduces `computed_columns` as a configuration option on the `View` constructor. 

This formalizes the computed column creation API and unifies the semantics of creating a computed column with that of aggregates, pivots, and other `View` config options. It eliminates overhead from the old API, which had to create and return a new `Table` for each call to `Table.add_computed_column`. 

In `perspective-workspace`, the old API had issues when multiple viewers are linked to the same table, which would be recreated each time a computed column was added, creating issues around unnecessary shared state. The new API allows each viewer in the workspace to maintain its' own set of computed columns, even when linked to the same underlying `Table`.

The (unpublicized) ability for the user to create custom computation functions is **removed** as part of this PR, as computations are no longer performed in the binding language (JS or Python), but completely in C++. Perspective provides a wide array of computation functions and aggregations, and adding more functions is part of our roadmap for the near future.

Preliminary benchmarks of `View`-based computed columns indicate a 400% speedup over the previous API, and almost zero overhead for creating computed columns on flat, sorted, or filtered views. As number of pivots increase, computation overhead increases to an acceptable amount (as is expected).

## UI Overview

This PR contains significant improvements to the computed columns UI in `PerspectiveViewer`.

Most notably, a custom **expression engine** and minimal **expression language** has been added to `PerspectiveViewer`, and is accessible through the new UI as well as programmatically through the `computed-columns` attribute API (by passing an array of string expressions). The expression syntax is extremely simple, and outlined in [this document](https://github.com/finos/perspective/blob/expressions/packages/perspective-viewer/Expressions.md). 

The expressions UI offers on-the-fly syntax validation and recursive type checking of input columns (making it impossible to write an expression that tries to get the square root of a datetime, for example).

Introduction of the expressions UI makes creating multiple computed columns (which involved a large amount of drag/drop interactions in the previous UI) extremely easy:

`(sqrt(("Profit" % "Sales" as "profit%sales"))) * (bin100th("Discount")) as "custom column"`

When decomposed, this expression contains four individual computed columns, each of which depend on the output of the previous column. In conjunction with the new computed columns API, Perspective will parse the syntax tree generated by the expression, compute dependencies, and name columns either automatically or by aliases provided in the expression editor (through use of `as`).

Once a computed column is created, it can be pivoted, aggregated, sorted, filtered, etc. like any other column, and it can also be used as an input to another computed expression column. While the UI does not offer expression completion at this time, additions/improvements to the UI are on the roadmap, and far easier to achieve given the completion of this major block of both internal refactor and feature-building.

## `perspective-python`

Computed column functionality has been enabled in Python. Though the expression engine is not available in Python (as it is built in Javascript), passing computed column configurations is supported:

```python
from perspective import table
t = table(data)
view = t.view(computed_columns=[{
  "column": "new computed column",
  "computed_function_name": "bin100",
  "inputs": ["Sales"]
}])
view.schema()
>>> { "Sales": float, "new computed column": float }
```

## `PerspectiveWidget`

expressions-widget

The expression engine and computed columns have been fully enabled in `PerspectiveWidget`, available inside Jupyterlab. Users can use the new computed columns UI inside the widget to create computed columns, and they can also create columns using the widget's `computed_columns` kwarg:

```python
from perspective import PerspectiveWidget
# initialize using the `computed_columns` kwarg
PerspectiveWidget(data, computed_columns=['sqrt((invert("Sales") as "inv")) as "computed"])
# or set the computed_columns attribute
widget =  PerspectiveWidget(data)
widget.computed_columns = ['sqrt((invert("Sales") as "inv")) as "computed"]
```

Because the widget is tied into the `PerspectiveViewer` API, the full expression engine is available through `PerspectiveWidget`.

### Improvements

- Brand new computed columns API and UI
- Computed columns in Python and `PerspectiveWidget`
- Expressions language and engine in `PerspectiveViewer`
- Type checking and column validation for computed columns

### Breaking API changes

- #### `table.add_computed` API has been completely **removed**, along with `func` and any code related to creating/handling a new table as part of adding computed columns.
- #### All handling for custom functions as part of computations has been completely **removed**.


sc1f:
Don't merge yet - I found a bug with restore in `perspective-workspace` that I need to fix

texodus:
Thanks for the PR!  Very cool new feature - and alot more on the way!

PR #986, opened 2020-03-20
Add update callback

timkpaine:
This adds an update callback hook to the widget. There may be alternative ways of doing this.


sc1f:
Do you have an example/test of when this would be useful? I'm wondering if it would make more sense to make `PerspectiveManager.on_update` an unimplemented signature, and then create a child class that implements `on_update` with the ipywidgets updater. 

That way we can segment widget-specific code from the manager (and later allow other callback handlers to be added as well).

timkpaine:
This isn't widget specific. Just need a callback of when the table gets updated from a UI, which can happen either from the Widget UI or from the regular PerspectiveViewer UI through a manager. 

Basically subscribe to an update event coming from the front end.

sc1f:
If we implement `on_update` on a child of manager, that way we can use custom update handlers instead of defaulting to the ipywidgets handler. I can take a crack at it as well - just let me know

timkpaine:
how would that differ from this? why would we implement it twice rather than implementing it once?

timkpaine:
@texodus this is ready to merge, will put together an example using it in a separate PR

texodus:
Yes, I understand what it does - what is the problem for which this feature is useful or necessary to solve?  Why not this?

```javascript
view.on_update(received)
```

timkpaine:
@texodus this seg faults
Screen Shot 2020-05-28 at 09 13 07

timkpaine:
and the `TABLE.on_update` method straightup doesn't exist
Screen Shot 2020-05-28 at 09 13 58

timkpaine:
so without this PR, how does one subscribe in python to updates to a table (not view) coming from the JS client?

texodus:
Surely it is reasonable to address the segfault then, as this is core, documented functionality of the Python API with heavy test coverage?

timkpaine:
If I host a view called "my_special_view", will `view.on_update` include this information so I can distinguish between hosted things? I don't see why this wouldn't make more sense in the manager, which could defer to view or table as necessary

texodus:
I don't understand - please define 
* "this information" - what information?  view.on_update provides which row IDs were updated, the rows themselves, and allows arbitrary query of the underlying view at that point in time, from a specific update step.  It doesn't tell you specifically which methods are called by a client - if this is the information you want, what is the use case for this?  I don't want to argue a straw man, so please be specific in what exactly you want.
* "hosted things" - I assume you mean `DTYPE_OBJECT` here, but I do not understand why you are trying to intercept this at such a low level, below perspective's remote object multiplexing even.

timkpaine:
closing, as discussed will utilize ports and refactor the manager (@sc1f) to make an equally user-friendly but less "exposes internal apis" version of this feature.

PR #995, opened 2020-03-30
Look for PyArrow relative to Perspective in @rpath

sc1f:
This PR adds an entry to `INSTALL_RPATH ` that allows the binary to look for PyArrow in a directory relative to `libbinding.so`. 

I ran into this problem when trying to push a Manylinux 2010 wheel of Perspective onto a Heroku dyno, which stores its Python packages in `/app/.heroku/`. Perspective was only looking for `pyarrow` using the path of `pyarrow` supplied to it from the machine that _built_ the wheel, which was different from the machine that the wheel was installed to.

Adding the new entry to `INSTALL_RPATH` makes the assumption that both `perspective-python` and `pyarrow` are installed to the same site-packages folder. This change is only useful for installing from wheel; if a user builds Perspective from source, the path to `pyarrow` has to be correct for the build to succeed.

Additionally, this PR adds a `_wheel_python` script to allow quick generation of wheels to the local filesystem using Perspective's Manylinux and Python docker containers. It supports the same command line arguments as `build_python`, allowing selection between Python 2/3 as well as Manylinux 2010/2014.

### DIFF
--- a/cpp/perspective/CMakeLists.txt
+++ b/cpp/perspective/CMakeLists.txt
@@ -552,10 +552,11 @@ elseif(PSP_CPP_BUILD OR PSP_PYTHON_BUILD)
         set(CMAKE_INSTALL_RPATH_USE_LINK_PATH TRUE)
         set(CMAKE_INSTALL_NAME_DIR "@rpath/")
 
+		# module_origin_path is the location of the binary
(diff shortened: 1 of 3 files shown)


texodus:

PR #999, opened 2020-04-03
Add `lock` to PerspectiveManager

sc1f:
This PR adds the `lock` attribute to `PerspectiveManager`, which will reject the following remote messages that can mutate the state of `Table`s and `View`s under management:

- `table`
- `table.update`
- `table.clear`
- `table.replace`
- `table.reset`
- `table.delete`

`PerspectiveManager`s exposed over the internet should be locked, either by calling `.lock()` or initializing with `lock=True`. This PR also includes tests and a compact example demonstrating message rejection when the manager is locked.

### DIFF
--- a/python/perspective/perspective/manager/manager.py
+++ b/python/perspective/perspective/manager/manager.py
@@ -19,11 +19,6 @@ from ..table import Table, PerspectiveCppError
 from ..table.view import View
 from .session import PerspectiveSession
 
-
-def gen_name(size=10, chars=string.ascii_uppercase + string.digits):
-    return "".join(random.choice(chars) for x in range(size))
-
-
 _date_validator = _PerspectiveDateValidator()
 
 
@@ -34,6 +29,10 @@ class DateTimeEncoder(json.JSONEncoder):
             return _date_validator.to_timestamp(obj)
         else:
             return super(DateTimeEncoder, self).default(obj)
+
+
+def gen_name(size=10, chars=string.ascii_uppercase + string.digits):
+    return "".join(random.choice(chars) for x in range(size))
 
 
 class PerspectiveManager(object):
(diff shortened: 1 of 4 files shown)


timkpaine:
I understand the concept but if we want to prevent js client modifications, this should be a construction time attribute that cannot be undone (e.g. no unlock). As is this current PR feels like "authentication-light" which is not something we want to be doing.

sc1f:
> I understand the concept but if we want to prevent js client modifications, this should be a construction time attribute that cannot be undone (e.g. no unlock). As is this current PR feels like "authentication-light" which is not something we want to be doing.

Agreed - wondering if we should just default the manager to locked as well

timkpaine:
> Agreed - wondering if we should just default the manager to locked as well

I think so.

The followup question is do you want to implement a token-granting system for perspective? This would be just on the client-server api, basically the ability to:
- restrict read/write access (separately) using a token
- ask the server for a new token
- ask the server to revoke a token
- tell the server about existing tokens, how to auth them

Then web servers could either provide a token scheme, or otherwise put the token granting behind an authenticated api (using their own authentication). This keeps us out of the authentication business while still allowing perspective to work nicely in authenticated contexts.

sc1f:
Yeah this is definitely working up to a token-based authentication scheme for the remote API.

Adding a lock is mostly for being able to host python-based perspective remote examples on the internet without having to worry about getting the example table cleared/deleted remotely.

python/perspective/perspective/manager/manager.py
```diff
@@ -57,11 +56,31 @@ class PerspectiveManager(object):
...
         self._callback_cache = _PerspectiveCallBackCache()
         self._queue_process_callback = None
+        self._lock = lock
+
+    def lock(self):
+        """Block messages that can mutate the state of `Table`s and `View`s
+        under management.
+
+        All `PerspectiveManager`s exposed over the internet should be locked to
+        prevent content from being mutated by clients.
```
texodus, line 76: This language simplifies things a bit :)

PR #1000, opened 2020-04-03
Fix incorrect sort order when hidden, pivoted columns are sorted

sc1f:
This PR fixes an issue where the incorrect `count` aggregate was used to sort hidden columns used as row/column pivots. Seen in the [Perspective COVID example](https://bl.ocks.org/sc1f/raw/6a04603d60967d3356a6156cb5399884/):

download

Because the `count` of each unique item in the column was used - and not the values of each unique item - row/column headers were displayed in the incorrect order. This PR fixes that issue in `t_view_config` and adds tests that assert the correct behavior:

Screen Shot 2020-04-03 at 5 09 20 PM


texodus:
Thanks for the PR!

PR #1011, opened 2020-04-10
Update Python benchmarks to include results for `update` calls

sc1f:
This PR improves the Python benchmarks by adding tests for calls to `update`, which is now benchmarked for both JSON and `pandas.DataFrame` when 10 views of various pivots are created and forced to be notified on each update. The benchmark measures the amount of time it takes to call `update` and then `Table.size()`, which forces processing of the update before returning a value.


texodus:
Thanks for the PR!

PR #1012, opened 2020-04-10
When updating tables typed as `int` with `float64` numpy arrays, copy instead of filling iteratively

sc1f:
This PR speeds up `perspective-python`'s numpy loading by adding a code path for a very common requirement - when a column is typed as an integer and updated with a `numpy.float64` array, the array is not trivially copyable into the column's underlying memory, and must be filled iteratively.

This slows down update speeds for dataframes/numpy arrays dramatically, and is an extremely common case given that most numpy arrays are defined as `float64`, and that users may prefer the truncation of unnecessary decimal points. By converting the `float64` array to `int64` before the C++ numpy loader attempts to load the update dataset, the array is made copyable and thus can be loaded with a massive (4-6x) speed increase over iterative fills.

Converting arrays from float to int maintains order and correctness of the values, as well as position of `numpy.nan`s - Perspective constructs a null mask _before_ the array is converted, and the indices of that null mask are used to apply `null` values on top of indices where the float-to-int conversion turns `numpy.nan` into garbage values.

### DIFF
--- a/cpp/perspective/src/cpp/base.cpp
+++ b/cpp/perspective/src/cpp/base.cpp
@@ -170,43 +170,43 @@ get_dtype_descr(t_dtype dtype) {
             return "none";
         } break;
         case DTYPE_INT64: {
-            return "i64";
+            return "int64";
         } break;
         case DTYPE_INT32: {
-            return "i32";
+            return "int32";
         } break;
         case DTYPE_INT16: {
-            return "i16";
+            return "int16";
         } break;
         case DTYPE_INT8: {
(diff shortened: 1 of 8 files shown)


timkpaine:
can we verify that this works on windows before merging, given the windows/numpy issues of the past

sc1f:
Yes - I added tests for it and we shouldn't merge until the branch fixing azure has been merged

texodus:
Thanks for the PR!  Reminder that we should start publishing benchmarks with each release.

PR #1017, opened 2020-04-21
Adds -Wall and fixes C++ build warnings

texodus:
As per title, adds `-Wall` build flag and fixes relevant warnings.

### DIFF
--- a/cpp/perspective/CMakeLists.txt
+++ b/cpp/perspective/CMakeLists.txt
@@ -217,6 +217,7 @@ if (PSP_WASM_BUILD)
 		--bind \
 		--source-map-base ./build/ \
 		--memory-init-file 0 \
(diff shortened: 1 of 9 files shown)


timkpaine:
Did you adjust/remove the cpp strict warnings flag in the cmake list?
https://github.com/finos/perspective/blob/master/cpp/perspective/CMakeLists.txt#L133

texodus:
No - it seems to me, no more necessary now, then it would have been to add the flag to emscripten build when the change flag was originally added to Python?

timkpaine:
I think we didn't build Wall for python because there were some "acceptable" warnings we didn't feel like dealing with. We can just make Wall a default flag and remove this cmake flag entirely.

PR #1023, opened 2020-04-27
Add perspective-click event to datagrid

jspillers:
* fix row selection for un-pivoted views
* make selection details mirror hypergrid
* update superstore custom grid example's perspective-viewer element to be selectable


texodus:
Thanks for the PR!

This feature has been deprecated by refactor to use `regular-table`, which has its own existing API for click events.

PR #1027, opened 2020-04-28
add perspective-update-complete event example to js guide

jspillers:
Adding a small example for update complete events


texodus:
Thanks for the PR!

PR #1028, opened 2020-04-28
Fix minor bugs in Perspective datetime + NaN handling

sc1f:
This PR fixes a few small bugs in `perspective-python`:

- `NaN`, `Infinity`, and `-Infinity` are no longer serialized into JSON by `PerspectiveManager`, and any remote calls into `PerspectiveManager` that returns a result with `NaN` inside will be rejected on the client-side. The server will not raise an Exception, but the exception will be visible inside the client browser console. 

#### A Short Explanation

Perspective's data loading and model treats `None` or `null` as invalid values, and the Javascript value `undefined` as a value for "nothing exists in this cell". To the extent possible, Perspective will strip and treat `NaN`s as `None` (in Python) or `null` (in Javascript). 

A case where `NaN`s cannot be coerced to `None` is in Arrows with columns that contain `NaN` values but do not have the `null_count` or the Arrow array's [validity bitmap](https://arrow.apache.org/docs/format/Columnar.html#validity-bitmaps) correctly set. To take advantage of the speed benefits provided by Arrow, Perspective treats Arrow-formatted data as strongly typed within each column, and where invalid values are indicated through use of the validity bitmap. This allows Perspective to `memcpy` arrays from Arrow into Perspective, i.e. a basically free operation that massively speeds up Arrow loading.

At this time, there are no plans to support a distinction between `NaN` and None in Perspective, especially as the visualization value of `NaN` and `None` are equivalent - both denote invalid values. Thus, using `perspective-python` and attempting to serialize data containing `NaN` values will fail, and this change provides an explicit error message.

Additionally, PyArrow's [default behavior](https://arrow.apache.org/docs/python/data.html#none-values-and-nan-handling) when converting from Pandas is to treat `NaN` values as `None`.

#### Other Changes

- Datetimes before the year 1900 are treated automatically as UTC using `calendar.timegm`, instead of `time.mktime` which raised an error whenever dealing with years before 1900.

- `datetime.min` values passed in to Perspective used to return "2001/1/1" due to a edge case in how timestamps were calculated before being passed into C++. Now, datetime.min values (whenever year, month, and day are `1` and hour, minute, seconds are `0`) return a timestamp of 0, i.e. the Unix epoch.

### DIFF
--- a/python/perspective/perspective/manager/manager.py
+++ b/python/perspective/perspective/manager/manager.py
@@ -174,21 +174,17 @@ class PerspectiveManager(object):
         cmd = msg["cmd"]
 
         if self._is_locked_command(msg) is True:
-            error_message = "`{0}` failed - access denied".format(
+            error_string = "`{0}` failed - access denied".format(
                 msg["cmd"] + (("." + msg["method"]) if msg.get("method", None) is not None else ""))
-            post_callback(json.dumps(self._make_error_message(
-                msg["id"], error_message), cls=DateTimeEncoder))
+            error_message = self._make_error_message(msg["id"], error_string)
+            post_callback(self._message_to_json(msg["id"], error_message))
             return
 
         try:
             if cmd == "init":
                 # return empty response
-                post_callback(
-                    json.dumps(
-                        self._make_message(
-                            msg["id"],
-                            None),
-                        cls=DateTimeEncoder))
+                message = self._make_message(msg["id"], None)
+                post_callback(self._message_to_json(msg["id"], message))
             elif cmd == "table":
                 try:
                     # create a new Table and track it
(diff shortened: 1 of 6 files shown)


python/perspective/perspective/tests/table/test_table_infer.py
```diff
@@ -185,3 +186,9 @@ def test_table_datetime_infer_no_false_positive(self):
         data = {"a": [" . - / but clearly not a date"]}
         tbl = Table(data)
         assert tbl.schema() == {"a": str}
+
+    @mark.skip
+    def test_table_datetime_infer_from_string_with_time(self):
+        data = {"a": ["11:00 ABCD"]}
+        tbl = Table(data)
+        assert tbl.schema() == {"a": str}
```
texodus, line 194: Yeah, we definitely need to unify on one implementation of this and port it to C++

python/perspective/perspective/manager/manager.py
```diff
@@ -392,6 +382,32 @@ def _make_error_message(self, id, error):
...
+        '''
+        try:
+            return json.dumps(message, allow_nan=False, cls=DateTimeEncoder)
+        except ValueError as error:
+            error_string = str(error)
+
+            # Augment the default error message when invalid values are
+            # detected, i.e. `NaN`
+            if error_string == "Out of range float values are not JSON compliant":
+                error_string = "Cannot serialize `NaN`, `Infinity` or `-Infinity` to JSON."
```
texodus, line 404: Very user friendly! Tragic that we must resort to this tho ...

PR #1041, opened 2020-05-08
Expressions ui improvements

Vishv07:
I aiming to solve a couple of issues


texodus:
This is neither a valid PR, nor your own work.

PR #1043, opened 2020-05-12
Implement Client/Server Editing

sc1f:
One of Perspective's main selling points is interoperability between `perspective-viewer` in the browser and `perspective-python` on the server. This PR furthers inter-op by enabling editing between a client and a server without the latency overhead of using a remote `Table`, i.e. proxying all operations to the server. By creating `ports` on the client and server, clients are able to make edits to a Perspective `Table` in browser memory, and automatically transfer their edits to the server and all other connected clients. Similarly, the server is able to propagate live updates to each of their connected clients, and state will remain automatically synchronized.

For a concise example, see the [Python Examples](https://github.com/finos/perspective/tree/master/python/perspective/examples) folder, where the `client_server_edits` Python and HTML files show the simplest possible implementation of client/server editing. The benefits of using client/server editing as opposed to traditional remote table editing are numerous, including:

- Lightning-fast performance on the client side, as the render is no longer tied to round-trips from the server.
- All clients receive updates from the server as well as edits from all other clients.
- Because the table the client sees is constructed within the client, it can be a subset/slice of the main dataset on the server.

The main tradeoff is loading time - remote tables load incredibly quickly, as there is no data to transport across the network besides what is being requested by the viewer. For a table to be constructed in the client-side, however, the client needs access to the entire table's dataset in Arrow form. The example provides a simple implementation of this, and for most datasets the tradeoff in loading time is far outweighed by increased performance.

# Changelog

- Add the `make_port` API to `Table`, and the `getEditPort` method to `perspective-viewer`, allowing clients and servers to create ports and enable client/server editing.
- **Changes the `on_update` API** - depending on the language, parameters to `on_update` callback functions need to be changed in order for proper operation. issues with callback parameters are easily debuggable, as they will raise exceptions in both Javascript and Python.
  * In **Javascript**, on_update callbacks now receive only **one** parameter: an object containing the keys `port_id` and `delta`, depending on if `cell` or `row` mode is selected:

```javascript
view.on_update(updated => {
  console.log(updated.port_id, updated.delta instanceof ArrayBuffer);
}, {mode: "row"});
```
  * In **Python**, on_update callbacks either receive the `port_id` argument by default, or they receive both `port_id` and `delta` (if the `row` mode was selected):

```python
def callback_with_delta(port_id, delta):
  print(port_id, delta)

def callback_without_delta(port_id):
  print(port_id)

view.on_update(callback_with_delta, mode="row")
view.on_update(callback_without_delta)
```
- Change the way we `process` batched updates and notify users in `on_update`, which will always fire with a port_id.
- Adds APIs on `perspective-viewer` so that each viewer automatically gets a port that is used for editing.
- Added support for passing arrow binaries to and from `PerspectiveWebsocket` (in JS) and `PerspectiveTornadohandler` (in Python).
- Fixes a regex mismatch that caused issues on Firefox but not Chrome.
- Adds tests + examples in Javascript and Python


cpp/perspective/src/cpp/gnode.cpp
```diff
@@ -597,10 +646,16 @@ t_gnode::get_table_sptr() {
 void
 t_gnode::promote_column(const std::string& name, t_dtype new_type) {
     PSP_TRACE_SENTINEL();
-    PSP_VERBOSE_ASSERT(m_init, "touching uninited object");
+    PSP_VERBOSE_ASSERT(m_init, "Cannot `promote_column` on an uninited gnode.");
```
texodus, line 649: Much better error message!  I assume you changed these to trace the segfault?  Reminds me that a future improvement may be to completely replace `PSP_VERBOSE_ASSERT` which check object initialization only and replace with proper valgrind support in the build scripts, stricter clang warnings and our existing excellent test coverage?

cpp/perspective/src/cpp/gnode.cpp
```diff
@@ -1098,10 +1153,21 @@ t_gnode::get_id() const {
...
 
+t_uindex
+t_gnode::num_input_ports() const {
+    return m_input_ports.size();
+}
+
+t_uindex
+t_gnode::num_output_ports() const {
+    return m_oports.size();
+}
```
texodus, line 1164: I am OK with removing these too if you're keen.

packages/perspective-viewer/src/js/computed_expressions/lexer.js
```diff
@@ -32,7 +32,7 @@ const UpperLowerCaseTokenType = createToken({
 });
 
 // Create tokens for column names and computed function names
-const column_name_regex_pattern = /(["'])(?<column_name>.*?[^\\])\1/y;
+const column_name_regex_pattern = /(["'])(.*?[^\\])\1/y;
```
texodus, line 35: Chrome bug?

cpp/perspective/src/cpp/gnode.cpp
```diff
@@ -532,19 +567,33 @@ t_gnode::_process_column<std::string>(
...
+
+    t_process_table_result result = _process_table(port_id);
+
+    if (result.m_flattened_data_table) {
+        notify_contexts(*result.m_flattened_data_table);
+    } 
+    
+    // Whether the user should be notified - False if process_table exited
+    // early, True otherwise.
+    return result.m_should_notify_userspace;
```
texodus, line 596: We could maybe even remove the concept of no-ops and drop `m_should_notify_userspace` altogether;  instead, we can just say if `_process_table` made it past the empty-table-checks earlier in this method, we call `notify_userspace`. This may simplify some of the aggregate logic in the `context_*` classes as well since they won't need to report this information.

packages/perspective-bench/bench/versions.js
```diff
@@ -34,7 +34,7 @@ const JPMC_VERSIONS = [
 
 const FINOS_VERSIONS = ["0.3.1", "0.3.0", "0.3.0-rc.3", "0.3.0-rc.2", "0.3.0-rc.1"];
 
-const UMD_VERSIONS = ["0.4.5", "0.4.4", "0.4.2", "0.4.1", "0.4.0", "0.3.9", "0.3.8", "0.3.7", "0.3.6"];
+const UMD_VERSIONS = ["0.4.8", "0.4.7", "0.4.6", "0.4.5", "0.4.4", "0.4.2", "0.4.1", "0.4.0", "0.3.9", "0.3.8", "0.3.7", "0.3.6"];
```
texodus, line 37: Thanks for updating this - reminds me, we could just use this and dump the pre-finos versions I think now?

```
$ npm view @finos/perspective  versions  --json
[
  "0.3.0-rc.1",
  "0.3.0-rc.2",
  "0.3.0-rc.3",
  "0.3.0-rc.4",
  "0.3.0",
  "0.3.1",
  "0.3.4",
  "0.3.5",
  "0.3.6",
  "0.3.7",
  "0.3.8",
  "0.3.9",
  "0.4.0-rc.1",
  "0.4.0-rc.2",
  "0.4.0-rc.3",
  "0.4.0-rc.4",
  "0.4.0-rc.5",
  "0.4.0-rc.6",
  "0.4.0",
  "0.4.1",
  "0.4.2",
  "0.4.4",
  "0.4.5",
  "0.4.6",
  "0.4.7",
  "0.4.8"
]
```

packages/perspective-viewer/src/js/viewer.js
```diff
@@ -850,6 +852,16 @@ class PerspectiveViewer extends ActionElement {
...
     }
+
+    /**
+     * Returns a promise that resolves to the element's edit port ID, used
+     * internally when edits are made using Hypergrid or DataGrid.
+     *
+     * @async
+     */
+    async getEditPort() {
+        return this._edit_port_lock;
```
texodus, line 863: This doesn't need to be `async` - `_edit_port_lock` is a `Promise`.

packages/perspective-viewer/src/js/viewer/perspective_element.js
```diff
@@ -267,6 +267,14 @@ export class PerspectiveElement extends StateElement {
             this.filters = this.getAttribute("filters");
         }
 
+        // Get an edit port from the table, and set it on the viewer so that
+        // all grid edits use the specified port.
+        this._edit_port = await table.make_port();
+
+        // Resolve the edit port lock, which allows for `get_edit_port` to be
+        // called in arbitary order without ever returning a null value.
+        this._edit_port_lock.resolve(this._edit_port);
```
texodus, line 276: I think this is unnecessary if we remove the `await` on line 272 above?  I would also like to understand how this can lead to `null` for port, since it seems (on inspection) that the `Promise` does not defer the actual call to `make_port()`.

I tried removing this on my branch and did not see errors in the tests, but the message implies this error condition is race-y ..

packages/perspective/src/js/api/server.js
```diff
@@ -153,7 +153,8 @@ export class Server {
                         // post transferable data for arrow
                         if (msg.args && msg.args[0]) {
                             if (msg.method === "on_update" && msg.args[0]["mode"] === "row") {
-                                this.post(result, [ev]);
+                                // actual arrow is in the `delta`
+                                this.post(result, [ev.delta]);
```
texodus, line 157: Most Understated Addition in this patch!

texodus:
Thanks for the PR!

PR #1044, opened 2020-05-12
Remove duplicate `psp_okey` column from arrow updates

sc1f:
This PR reproduces and fixes a segmentation fault with arrow updates and string reallocation - previously, the `psp_okey` column was being duplicated incorrectly on an Arrow update, and this turned out to be the cause of a few different segfaults that became trivially reproducible (and has been reproduced and tested in the Python tests).

Removing the incorrect duplication fixes the tests and seems to (for now) prevented any further segfaults related to this issue.

### DIFF
--- a/cpp/perspective/src/cpp/emscripten.cpp
+++ b/cpp/perspective/src/cpp/emscripten.cpp
@@ -1064,7 +1064,8 @@ namespace binding {
             
             // Always use the `Table` column names and data types on up
             if (table_initialized && is_update) {
-                auto schema = gnode->get_output_schema();
+                auto gnode_output_schema = gnode->get_output_schema();
+                auto schema = gnode_output_schema.drop({"psp_okey"});
                 column_names = schema.columns();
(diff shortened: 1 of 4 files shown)


cpp/perspective/src/cpp/emscripten.cpp
```diff
@@ -1064,7 +1064,8 @@ namespace binding {
             
             // Always use the `Table` column names and data types on up
             if (table_initialized && is_update) {
-                auto schema = gnode->get_output_schema();
+                auto gnode_output_schema = gnode->get_output_schema();
+                auto schema = gnode_output_schema.drop({"psp_okey"});
```
texodus, line 1068: It is amazing to me that this coerces to `std::set<std::string>` ..

Since `column_names` is simply iterated on 10loc below this, it may be cheaper to simple `continue` on `"psp_okey"`, since `drop()` iterates through the set and reconstructs both name and type vectors from scratch.

PR #1057, opened 2020-05-21
Enable Manylinux wheel builds

sc1f:
This PR makes Perspective's linux wheels fully Manylinux2010 and Manylinux2014 compatible, adds wheel build tasks for MacOS 10.14 (Mojave) and MacOS 10.15 (Catalina), as well as ensures passage of the `conda-forge` Mac build.

# Changelog
- Edits dockerfiles to include PyArrow from source, as well as force an install of `auditwheel`
- Fixes `CMakeLists` to allow full Manylinux builds to pass due to a difference between `PYTHON_INCLUDE_DIRS` and `Python_INCLUDE_DIRS`.
- Runs `auditwheel show` and `auditwheel repair` whenever Linux wheels are built, which ensures that all wheels are fully portable and compliant to the Manylinux PEPs.
- Adds a Python 2 wheel task for Manylinux2010
- Adds Azure pipelines tasks for MacOS wheel builds similar to Manylinux builds
- Adds a MacOS 10.15 Azure pipeline
- Removes `CLOCK_MONOTONIC` from `compat_impl_osx`, which is breaking [builds](https://github.com/conda-forge/staged-recipes/pull/10884/files) on `conda-forge`. This is due to `clock_gettime` and `CLOCK_MONOTONIC` not being implemented in MacOS until 10.12. Because the method that calls `clock_gettime` is not used anywhere in the compiled code, I commented out the original implementation and forced the offending method to return a 0.


cpp/perspective/CMakeLists.txt
```diff
@@ -352,13 +352,22 @@ elseif(PSP_CPP_BUILD OR PSP_PYTHON_BUILD)
 			message("${Red}Manylinux build has no python shared libraries${ColorReset}")
 			find_package( Python ${PSP_PYTHON_VERSION} REQUIRED COMPONENTS Interpreter )
 			find_package( PythonHeaders ${PSP_PYTHON_VERSION} REQUIRED)
+			message("${Cyan}Using PYTHON_INCLUDE_DIRS: ${PYTHON_INCLUDE_DIRS}, Python_LIBRARIES: ${Python_LIBRARIES} ${ColorReset}")
+
+			# find_package(PythonHeaders) sets the uppercased version of
```
timkpaine, line 357: we wrote this cmake package, just modify it to match

sc1f: Done

timkpaine:
windows 3.7 and 3.8 passing...need to fix a python bug on 3.6

timkpaine:
@texodus 
New docker image names:
- python2 - Python2.7
- python2_manylinux2010 - Python2.7 manylinux2010
- python36 - Python 3.6 (**new**)
- python36_manylinux2010 - Python 3.6 manlinux2010 (**new**)
- python36_manylinux2014 - Python 3.6 manlinux2014 (**new**)
- python37 - Python 3.7 (**renamed from python3**)
- python37_manylinux2010 - Python 3.7 manlinux2010 (**renamed from python3_manylinux2010**)
- python37_manylinux2014 - Python 3.7 manlinux2014 (**renamed from python3_manylinux2014**)
- python38 - Python 3.8 (**new**)
- python38_manylinux2010 - Python 3.8 manlinux2010 (**new**)
- python38_manylinux2014 - Python 3.8 manlinux2014 (**new**)

cpp/perspective/src/include/perspective/last.h
```diff
@@ -0,0 +1,30 @@
...
+   to come after any other perspective includes, which 
+   themselves have to precede any arrow includes.
+
+   The preprocessor is fun.
+
+   https://bugs.python.org/issue24643
+*/
+
+#ifdef WIN32
+#if _MSC_VER >= 1900
```
timkpaine, line 25: @wesm had to duplicate this from feather, happens when including pybind and arrow on windows python3.6. I know its a python bug, but might make sense to make a jira in case people look for it down the road

timkpaine:
@texodus @sc1f this is ready for a docker rebuild/review

texodus:
Thank y'all for your hard work on this, it is very cool!

I am working on updating the Docker images for this now.  Before I merge this - I propose strongly we limit the Azure Pipelines Python builds to a subset reasonably likely to cover possible failure states.

* `python2_manylinux2010` - Python2.7 manylinux2010
* `python36_manylinux2014` - Python 3.6 manlinux2014 (new)
* `python37_manylinux2014` - Python 3.7 manlinux2014 (renamed from python3_manylinux2014)
* `python38_manylinux2014` - Python 3.8 manlinux2014 (new)

These `dockerfile` in their current state are complex and heavily duplicitous.  They take hours to build, and much longer to debug.  Running all suites on them locally takes hours, so there is 0 possibility they will be run before push by any developer.  Regarding Azure specifically, in net, one commit to Azure Pipelines now relies on ~60gb worth of docker images.  It also runs 24 complete iterations of the build and test suite, where the only variable that is changed affects maybe 0.01% of the codebase.

I understand these builds and tests have value, but the variation vs. cost ratio here seems out of balance.  I can still build and push these images, and I can run these tasks before we dist or publish, and perhaps this merits a `yarn prod` which scripts all of these build/tests, which we can eventually wire into a "dist" pipeline.

Regarding running these locally, removing the docker startup cost of 11 containers would be a start towards making it practical.  I'd also love to see if we can combine the python version and/or manylinux images into single containers with multiple python installations.

timkpaine:
I can probably consolidate to just a manylinux2010 and manylinux2014 image, since they have all python versions installed on them anyway

sc1f:
We can have the additional builds that we don't need for per-commit CI to run just on a tag release - that way we have the wheels pre-generated for dist.

timkpaine:
updates:
- [x] filter down the number of required builds for non-release builds
- [x] remove "official"
- [x] default to `manylinux2010`
- [ ] consolidate all python versions into 2 docker images, `manylinux2010` and `manylinux2014`

wesm:
Do you all plan to get yourselves synced up with pyarrow mainline?

sc1f:
@texodus this should be good to go

sc1f:
@wesm the custom arrow build we have for Emscripten might take a little fixing, but it shouldn't be too hard to get Perspective on 0.17 in the near future

PR #1074, opened 2020-06-02
Use local time for column/row headers and computed functions

sc1f:
This PR fixes several issues with computed functions and computed columns in `perspective-python`:

- the `hour_of_day`, `day_of_week`, and `month_of_year` functions now use local time to perform their calculations. Previously, all calculations were in UTC which could cause issues with datetimes that were on the cusp of one day/month to the next, as well as returning the wrong hour of day when compared with the input column.
- row and column headers for datetimes (i.e. when a datetime column is pivoted on) are now in local time, consistent with the behavior of datetimes across Perspective:

Screen Shot 2020-06-02 at 6 13 47 PM

- Additional tests have been added for both features
- Edits to documentation around datetime handling in the Python guide
- Fixes a race condition where computed columns created via attribute were created but not preserved after the view changes.
- Fixes a scrolling issue within Jupyterlab (and other environments where Perspective is both > page width and the page is scrollable vertically) where using the arrow keys to navigate autocomplete would move the entire page.


andyferris:
Is it possible to have both behaviours implemented and be able to choose?

sc1f:
> Is it possible to have both behaviours implemented and be able to choose?
@andyferris 

Which behaviours are you referring to exactly? For the computed functions, even though it would be possible to preserve the old behaviour of UTC, it just isn't correct when compared with the underlying data:

_Before_
Screen Shot 2020-06-03 at 10 47 02 AM

_After_
Screen Shot 2020-06-03 at 10 46 53 AM

The issue is the same with row headers - we don't have an API for controlling their output from userland (they are generated in the C++ part of the codebase), and UTC output would be inconsistent with the underlying data.

PR #1077, opened 2020-06-04
Refactor manager internal API, speed up string filters in UI,  add manager API tests

sc1f:
This PR reduces the public API/code footprint of `PerspectiveManager` by moving all internal logic to a separate base class, which should make the Manager API more "manageable" and easier to read. Additionally, it reduces the amount of computation required for string filters by making the suggestions list quicker to compute and serialize.

### Python Changelog
- Add `PerspectiveManagerInternal`, which hosts all internal methods that maintain the wire API.
- Add tests around `PerspectiveSession` being able to open/close multiple sessions
- Add tests around `PerspectiveManager` handling `on_update` and `remove_update` over the wire API
- `__ROW_PATH__` values are now python values that match the underlying schema (`datetime.datetime` for "datetime" columns, etc.) instead of strings. This behavior is now consistent with `perspective.js`.
- Makes `_PerspectiveCallbackCache` easier to use by removing the need to get the underlying list before operating on its contents
- Changes `_PerspectiveCallbackCache.remove_callback` to remove the callbacks for which the lambda returns True. This changes its behavior and makes it no longer consistent with `filterInPlace`, which is used in the Javascript version of Perspective to remove callbacks.
- Removes `numpy.isnat` and fixes several pandas and numpy tests around datetimes
- Replaces `isinstance(str)` with `isinstance(six.string_types)` for Python 2 compatibility

### JS Changelog
- Makes filter suggestions for string columns faster by a magnitude of ~10x (on large datasets)
  * Because we don't use the actual columns of the dataset to generate the suggestion, make the new `view` with `columns=[]` instead of serializing all columns and immediately disposing of them.


sc1f:
@texodus this branch is rebased on top of #1082 - please merge that PR first so we have consistent line endings, otherwise the `eol=lf` in `.gitattributes` is preventing me from most git operations because line endings are automatically changed, even if I discard those changes in the working directory

PR #1090, opened 2020-06-18
Computed expressions respect left-to-right associativity and operator precedence

### LINKED ISSUE #1072: Expressions do not support associativity as expected
### Steps to Reproduce:
 1. On `superstore.html`, open the config 
 2. Open the `New Column` editor and enter:
```
"Discount" + "Quantity" + "Profit" as "broken sum"
```
 3. Save the `New Column` 
 4. Remove unnecessary pivots and put the columns in order `Discount`, `Quantity`, `Profit`, `broken sum`

### Expected Result:
The column should show the sum of all 3 of the selected column.
Correct behavior can be seen by creating a new column with the following:
```
("Discount" + "Quantity") + "Profit" as "correct sum"
```

Screen Shot 2020-05-29 at 4 09 07 PM

### Actual Result:
The sum presented is the sum of only the first two columns `Discount` + `Quantity`
Screen Shot 2020-05-29 at 4 11 30 PM

Screen Shot 2020-05-29 at 4 03 32 PM

sc1f:
This PR fixes #1072 by rewriting the expression parser to allow arbitary operator associativity without parentheses, as well as correct operator and function precedence: `a + b * c ^ d / e`, for example, evaluates to `a + (b * ((c ^ d) / e))`. This allows expressions to be much more complex and arbitrary than previously possible:

newcomputedexpressions

# Changes
- Expressions respect left to right associativity and operator precedence
- Functions can have arbitary expressions inside them without additional parentheses
- Expression editor no longer destroys already-rendered tokens when an error occurs - so text no longer moves under the cursor while typing
- Adds parser + autocomplete tests
- Adds invariant tests for the parser
- Cleans and refactors the expression editor/parser


texodus:
Thanks for the PR!

PR #1095, opened 2020-06-19
Fix Windows build on Azure

sc1f:
`BOOST_ROOT` on Azure win64 got moved one level down into `Boost/1.69.0/x86_64`, which is breaking Azure builds.

This branch is on top of #1082  - that should be good to merge without doing a rebuild/re-CI, which would fail without this patch.


sc1f:
@timkpaine all the PRs need to go on top of manylinux and CRLF fix - I couldn’t do any local development without fixing CRLF first

texodus:

Issue #1097, opened 2020-06-19
date columns with 'unique' aggregate breaks datagrid plugin 


panda311:
### Steps to Reproduce:
 In Perspective superstore example, adding Ship Date to the columns list and making column's 
 aggregate 'unique' and minimize configurations, this breaks the datagrid component and Scroll 
 also.

The console gives invalid time value Error, This might be possible that the associated function might not be able to able to handle value as null.

sc1f:
Thanks for the report! We are aware of the issue (#1068) - it should be fixed in the next major update to datagrid as Perspective is ported to use [regular-table](https://github.com/jpmorganchase/regular-table).

PR #1103, opened 2020-06-24
Fix #1068

### LINKED ISSUE #1068: `unique` aggregate on datetime columns breaks pivoted datagrid
### Steps to Reproduce:
 1. On `superstore.html`, add `Ship Date` or `Order Date` to the columns list.
 2. Set the datetime column's aggregate to `unique`.

### Expected Result:
A rendered datagrid, similar to this Hypergrid output:

Screen Shot 2020-05-27 at 1 06 09 PM

### Actual Result:
Render is broken, as `_format` in `view_model` cannot call `Intl.DateTimeFormat` on the value "-".:

Screen Shot 2020-05-27 at 1 06 22 PM

I'm assuming at the point `_format` is called, any `null` and `undefined` values are converted to "-" already. I'm also assuming a much simpler repro of this bug exists if a row in a datetime column is null.*

Screen Shot 2020-05-27 at 1 12 09 PM

texodus:
Removes a special-cased `unique` aggregate.

### DIFF
--- a/cpp/perspective/src/cpp/extract_aggregate.cpp
+++ b/cpp/perspective/src/cpp/extract_aggregate.cpp
@@ -15,7 +15,6 @@ namespace perspective {
 t_tscalar
 extract_aggregate(
     const t_aggspec& aggspec, const t_column* aggcol, t_uindex ridx, t_index pridx) {
-    static std::string non_unique("-");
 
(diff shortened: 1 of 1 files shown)


sc1f:
Also fixes #1097

PR #1104, opened 2020-06-25
Unpin arrow version

sc1f:
This PR unpins Perspective from Arrow 0.16.0, and uses the found arrow version to infer the version of the PyArrow library to be pulled in. This allows a little more flexibility in terms of PyArrow versions we allow - `perspective-python` will always install 0.16.0 by default, but in this case earlier/later versions are allowed.


texodus:
Thanks for the PR!

PR #1105, opened 2020-06-29
Generate/test Manylinux wheels, clean up wheel/build process

sc1f:
This PR fixes (hopefully for the final time) our build process to build Manylinux-compliant wheels.

~~Due to an [issue](https://github.com/pypa/auditwheel/pull/245) in Auditwheel (fixed and released in auditwheel 3.2.0), our internal C++ bindings were not being picked up after `auditwheel repair` replaces the RPATHs of the C++ binding.~~

A previous iteration of this PR required a rebuild of our docker images, however I removed a hack from our Linux build where it inserted `./../../pyarrow` into the RPATH, which was originally designed as a "hint" to look for pyarrow immediately relative to the install location (usually site-packages). By removing this hack from the Linux build (but keeping it in MacOS), Auditwheel was able to properly delocate the shared libraries and repair the wheel.

Testing on a clean `python:3.6-buster` and a `python:3.7-buster` image show that the wheels install properly from `pip install`, and after pulling down the `perspective-python` test suite and running it against the installed version from wheel, all tests are shown to pass.

Additionally, this PR also adds `test_wheels.sh` which reduces our bloat in `azure-pipelines` by externalizing the build/install in venv/test process for wheels and making it more maintainable.


texodus:
Thanks for the PR!

PR #1109, opened 2020-07-02
Add client/server editing example, fix hypergrid to allow editing in workspace

sc1f:
This PR adds `examples/workspace-editing`, which implements the `client_server_editing` example from `perspective-python` in a `perspective-workspace`. It is designed to be run in parallel with `client_server_editing.py` in the Python examples folder, and provides a simple, self-contained and annotated demonstration of Perspective's client/server editing capabilities.

This PR also fixes an issue in the Hypergrid plugin that was preventing edits from being tracked when using a remote server and in `perspective-workspace`.

### DIFF
--- a/examples/workspace-editing/README.md
+++ b/examples/workspace-editing/README.md
@@ -1,0 +1,12 @@
+# `perspective-workspace` client-server editing example
+
+For this example to work, you'll need a `perspective-python` server running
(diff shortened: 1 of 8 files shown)


texodus:
Thanks for the PR!  Great example of a very cool new feature!

PR #1110, opened 2020-07-02
slight UX change, dont end with stack trace as it looks like it raise…

timkpaine:
…d the error instead of catching it


texodus:
Thanks for the PR!

PR #1112, opened 2020-07-06
Explicitly floor start row/col and ceil end row/col

sc1f:
This PR prevents `to_format` from being called with float start/end row and column indices, and unifies behavior across Javascript and Python bindings. This fixes an issue with `perspective-datagrid` where float start/end row indices were being sent to a `perspective-python` Table, and rejected by a C++ function whose parameters were integers.

For `start_row` and `start_col`, `floor` is called - so requesting a start row of `0.3` would start at `0`, and not at `1`. For `end_row` and `end_col`, `ceil` is called so that an end row of `1.2` would return `2`, and not 1. This means that the data window should never be smaller than requested by the plugin.


sc1f:
Fixes #1115

texodus:
Thanks for the PR!

PR #1114, opened 2020-07-08
Datetime computed functions use local time

sc1f:
This PR fixes all datetime computed functions to compute in local time instead of UTC, through the use of `std::localtime`. 

Previously, computed functions that returned strings/ints from datetimes were fixed to return in local time, but I left out the functions that returned datetimes (as I assumed they were correct). This PR fixes that mistake and makes sure that day/week/month/year bucket use local time. The second/minute/hour buckets are unaffected, as timezones do not affect the correctness of their calculation.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -369,7 +369,7 @@ jobs:
     - script: npm install -g yarn
       displayName: "Install Yarn"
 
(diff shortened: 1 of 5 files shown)


sc1f:
Closing and replacing with a PR with more fixes

PR #1116, opened 2020-07-09
Properly display computed axis on charts, use local time for all datetime functions

sc1f:
This PR makes several fixes to computed functions and highcharts/d3fc:

- Fixes an issue where computed columns would not be displayed correctly in chart axis, both in highcharts and d3fc. This was due to computed columns no longer being present in the table schema.
- Fixes an issue where datetime computed functions calculated buckets in UTC and not local time. This caused issues with datetimes that were part of one bucket in local time, but another in UTC - i.e. `03/15/2020 23:59 EST` becomes  `03/16/2020 04:59 UTC`.
- Distinguishes between `computed_schema` on the view and on the table in `PerspectiveManager`, as they have different signatures. A future improvement would be renaming `Table.computed_schema` to something like `Table.get_computed_schema`, as the table method receives a set of computed columns and calculates their output types without actually creating the columns themselves.
- Cleans up the internals of `PerspectiveManager` and returns an error message if `table.delete()` is called on a remote table. Previously, `table.delete()` would fail due to code flow reasons, which prevented the method from being called but is opaque. This provides a transparent error message and explanation.
- Fixes the `last by index` aggregate to use `psp_okey` instead of `psp_pkey` in the aggspec, as `psp_pkey` does not exist in some of the intermediate data structures used during a live update. Additionally, rename `AGGTYPE_LAST` to `AGGTYPE_LAST_BY_INDEX`, as the `last` aggregate uses `AGGTYPE_LAST_VALUE`.


azure-pipelines.yml
```diff
@@ -369,7 +369,7 @@ jobs:
     - script: npm install -g yarn
       displayName: "Install Yarn"
 
-    - script: yarn
+    - script: yarn --network-timeout 600000
```
texodus, line 372: ??

packages/perspective-viewer-d3fc/src/js/charts/treemap.js
```diff
@@ -48,6 +48,8 @@ function treemap(container, settings) {
         .each(function({split, data}) {
             const treemapSvg = d3.select(this);
 
+            console.log(split, data);
```
texodus, line 51: Debug leak

texodus:
Thanks for the PR!  This won't be the last time we address this aggregate type, due ot the large body of work necessary to properly integrate this with sort, but its a great start!

PR #1118, opened 2020-07-10
Simplify python build

timkpaine:
This PR seeks to replace the CMake build system with vanilla distutils for python. 

*Benefits*
- Much more portable than before
- Vendored the header-only dependencies so no outside fetches at build time
    - boost no longer required as a system dependency
    - cmake no longer required as a system dependency
- Should support conda on all platforms "out of the box"

*Negatives*
- Slower builds potentially (using a numpy hack to have parallel builds)
- No modification tracking (must recompile all files every time)
    - Can work around this for our day-to-day development with `ccache`
- tbb required as a system dependency


texodus:
Thanks for the PR!  I think this patch has been overtaken by #1207 so I will close for now - feel free to re-open if I've missed something or this is ready for review.

PR #1140, opened 2020-07-28
Enable updates from arrows that have more/less columns than the Table schema

sc1f:
This PR implements and tests the ability for tables to be updated with arrows that have more/less columns than the table's schema - i.e. the table's schema will always be respected when reading through the arrow binary. If a column exists in the arrow binary but does not in the schema, it will be skipped, and vice versa - if a column exists in the schema and not in the arrow binary, it will not be filled. Additionally, this PR fixes an issue where an integer index column would cause an error when it was updated with an arrow - index columns will not be promoted, and filled through iteration instead.

### DIFF
--- a/python/perspective/perspective/tests/table/test_update_arrow.py
+++ b/python/perspective/perspective/tests/table/test_update_arrow.py
@@ -13,6 +13,7 @@ import pyarrow as pa
 import pandas as pd
 import numpy as np
 from datetime import date, datetime
+from pytest import mark
 from perspective.table import Table
 
 SOURCE_STREAM_ARROW = os.path.join(os.path.dirname(__file__), "arrow", "int_float_str.arrow")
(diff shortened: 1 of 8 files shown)


python/perspective/perspective/tests/table/test_update_arrow.py
```diff
@@ -667,6 +839,88 @@ def test_update_arrow_updates_dictionary_stream(self, util):
...
+        tbl.update(arrow_data)
+
+        assert tbl.size() == 3
+        assert tbl.view().to_dict() == {
+            "a": [None, "a", "b"],
+            "b": ["z", "x", "y"]
+        }
+
+    @mark.skip
+    def test_update_arrow_partial_updates_dictionary_stream_duplicates(self, util):
```
texodus, line 863: Painful issue.  As discussed, we can detect this if we care to sacrifice a bit of memory for a HashSet to detect collisions, but I'm far from convinced we want to.  I am curious if a dictionary encoded column in arrow is even considered valid?

Issue #1139, opened 2020-07-28
Chromium browser doesn't load perspective.wasm.worker.js in Python remote.py/remote.html example


pgagarinov:
### Steps to Reproduce:
Same as in #1128 

### Actual Result:

<img width="1350" alt="image" src="https://user-images.githubusercontent.com/4868370/88724967-07f19980-d134-11ea-853e-03333a853fe9.png">

### Environment:
perspective 0.5.1

### Caveats
Same example works in Brave browser.

pgagarinov:
Hello @texodus. I'm not sure if it is a bug or not, from reading the documentation my impression was that perspective.wasm.worker.js should be loaded from CDN when Python server is started. This is what Brave browser does: 
<img width="1218" alt="image" src="https://user-images.githubusercontent.com/4868370/88725429-c31a3280-d134-11ea-8e61-3ed8949c89e9.png">

If it is a bug is there any workaround for that? Thanks.

texodus:
Thanks for the report, I caught this last night also.  It appears to be a regression in Chrome 84 - I've tested in 83 and it works fine.  For some reason, `__webpack_public_path__` is unset when hosted from alias un `unpkg.com`.  A work around for now is to use the direct links instead of the aliases - e.g. use [https://unpkg.com/@finos/perspective/dist/umd/perspective.js](https://unpkg.com/@finos/perspective/dist/umd/perspective.js) instead of [https://unpkg.com/@finos/perspective](https://unpkg.com/@finos/perspective)

texodus:
Fix has been released in 0.5.3

Issue #1128, opened 2020-07-17
Python example remote.py/remote.html doesn't work in Firefox


pgagarinov:
### Steps to Reproduce:
 1. run `remote.py` script from `python\perspective\examples` subfolder (I had to remove `#sys.path.insert(1, os.path.join(os.path.dirname(__file__), '..'))` line though as it is not necessary if perspective from conda-forge is used

 2. try to open `http://localhost:8888` in Firefox
page opens but no data is shown

 3. open the console to see the following
<img width="887" alt="image" src="https://user-images.githubusercontent.com/4868370/87801205-0e406580-c858-11ea-84b6-13b30b72f5b7.png">

### Expected Result:
In chrome-based browsers (tested in Brave) everything works

### Environment
Firefox 78.0.1
Perspective 0.5.1 from conda-forge

timkpaine:
1. is required for a local development build so not an issue

texodus:
Can't repro, this has likely bit-rotted. In general, while I do want to fix any obvious errors on web platforms like Firefox, as we cannot currently run our test suite on this platform, it is not technically supported.

PR #1141, opened 2020-07-29
Fix path resolution for cross-origin assets

texodus:
Updates Perspective's cross-origin Web Worker and Wasm loading logic to use [`document.currentScript`](https://developer.mozilla.org/en-US/docs/Web/API/Document/currentScript).  In the before times, the long-long ago, we supported a browser called Internet Explorer 8, which led to some highly questionable logic for determining the path of these assets, throwing an `Error` and parsing the stack trace ...

I expect this will fix issues #1139 and possibly #1128 - but I can't say for sure until 0.5.3 is released, as the specific interplay of this issue was only detected due to how `unpkg.com` resolves `browser` alias fields in `package.json`.

### DIFF
--- a/packages/perspective-webpack-plugin/src/js/utils.js
+++ b/packages/perspective-webpack-plugin/src/js/utils.js
@@ -7,53 +7,14 @@
  *
  */
 
-/**
- * An Object for capturing details of the invoking script's origin.
- *
- * Returns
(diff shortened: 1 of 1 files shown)


texodus:
Thanks @sc1f for discovering the root cause for this!

PR #1153, opened 2020-08-13
Fix segfault and errors on 2-sided sorted views when data window is invalid

sc1f:
Retrieving data from a `View` can be controlled via the `start_row`/`end_row` and `start_col`/`end_col` arguments to `to_json`, which are parsed in a similar fashion to Python's `range` - inclusive of `start` and stopping at `end - 1`. 

Invalid data windows - where `start > end` or `start == end` - are handled by returning empty data. However, in the special case of two-sided sorted views, it did not correctly handle invalid data windows while constructing a column indices vector, and would segfault or throw a ValueError.

This does not raise an issue with the viewer/plugins - the data window requested by `perspective-viewer` will be valid, even if the viewport is extremely small. However, this interferes with a Websocket testing suite I'm currently building, which randomly generates data windows with the expectation that they are handled correctly by Perspective.

Because there are many situations where a data window can be _technically_ invalid, such as requesting a `to_json` on a view with 0 rows, or a view with 0 columns (such as the one created when generating autocomplete results for string filtering), I think it's better to have consistent behavior rather than throwing an error.

### Changes

- Fixes segfault and ValueError when data windows are invalid for 2-sided sorted views
- Adds tests for all view contexts to check the behavior for invalid data windows are consistent

### DIFF
--- a/cpp/perspective/src/cpp/view.cpp
+++ b/cpp/perspective/src/cpp/view.cpp
@@ -389,11 +389,27 @@ View<t_ctx2>::get_data(
         }
 
         cols = column_names(true, depth);
-        column_indices = std::vector<t_uindex>(column_indices.begin() + start_col,
-            column_indices.begin() + std::min(end_col, (t_uindex)column_indices.size()));
+
+        // Filter down column indices by user-provided start/end columns
+        column_indices = std::vector<t_uindex>(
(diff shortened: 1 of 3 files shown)


texodus:
Thanks for the PR!

PR #1171, opened 2020-08-29
Fix mac wheels, force arrow ~.17 for consistency between wheel and reqs, bump to arrow to 0.17.1 for js

timkpaine:


sc1f:
Tangentially related: I built a manylinux2010 wheel using the docker image, and tried installing it on a clean docker `python-3.7-buster` image. Because of `auditwheel`, Perspective will ship with its own `libarrow.so.16` and link to it, but it installed Pyarrow `0.17.1`:

Screen Shot 2020-08-29 at 2 27 59 PM

Screen Shot 2020-08-29 at 2 27 51 PM

Screen Shot 2020-08-29 at 2 29 32 PM

timkpaine:
it shouldnt, it should install pyarrow 0.17 and build against that except on py2 when it will use 0.16

timkpaine:
ah right, we build from source in the manylinux images

sc1f:
Pulled the latest version of this branch and built a MacOS wheel, running `otool -l libbinding.so`:

```
Load command 17
          cmd LC_RPATH
      cmdsize 32
         path @loader_path/ (offset 12)
Load command 18
          cmd LC_RPATH
      cmdsize 40
         path @loader_path/../../pyarrow (offset 12)
Load command 19
          cmd LC_RPATH
      cmdsize 64
         path /usr/local/lib/python3.7/site-packages/pyarrow (offset 12)
Load command 20
          cmd LC_RPATH
      cmdsize 120
         path /usr/local/Cellar/python/3.7.5/Frameworks/Python.framework/Versions/3.7/lib/python3.7/site-packages/pyarrow (offset 12)
```
Looks like the Pyarrow used at build time is still in the rpath.

timkpaine:
thats ok, we could go the extra mile to prune it but the `@loader_path/../../pyarrow` should come first and work

sc1f:
The manylinux wheel I built is from master, but I left a comment here since I didn’t think it would install 0.17.1, just 0.17.0. Let me try it on this branch - sorry for confusion

On Aug 29, 2020, at 4:26 PM, Tim Paine <notifications@github.com> wrote:

﻿

ah right, we build from source in the manylinux images

sc1f:
Screen Shot 2020-08-29 at 8 37 37 PM

okay - built a new manylinux wheel from this branch, did not rebuild the docker images. Looks like it's pulling arrow 17 now, although I have no idea why considering that it's not installed in the docker image. `0.17.1` was installed as part of the wheel installation though, which seems correct.

texodus:
As configured, this will not compile the WebAssembly build due to the introduced deprecation warnings that the python build allows, and updating the syntax to remove the warnings [breaks `0.16.0`/python2](https://dev.azure.com/finosfoundation/perspective/_build/results?buildId=1222&view=logs&j=a7525133-5bc3-5a3d-ed30-a8509984a81f&t=58993f68-0165-5319-a03a-bf27844b461c).  Can we consider deprecating `python2`, given that it is not supported by `pyarrow`, in perspective `0.6.0`?

timkpaine:
yep I think we can deprecate py2

timkpaine:
closing to separate wheel fixes from arrow upgrade

PR #1177, opened 2020-09-04
Add websocket stresstests for Tornado server

sc1f:
# Feature

This PR adds a stresstest suite for the `perspective-python` tornado server, allowing for clients to be spawned against any remote `perspective-python` server to test performance and load handling. Much like the benchmark suite, results are logged to a Perspective table and can be visualized in the browser.

## Changelog

- Adds test client and test servers
- Adds a `PerspectiveManager` and `PerspectiveTornadoHandler` implementation that provides telemetry to the client


texodus (approved):
Looks good!

texodus:
Thanks for the PR!

PR #1178, opened 2020-09-06
Move Python examples to `/examples/`

texodus:
Refactors `python/perspective/examples` to two top-level example Python projects, `examples/tornado-python` and `examples-tornado-streaming-python`.


sc1f (approved):
Tested locally - looks good!

PR #1182, opened 2020-09-06
Shhhhhhh!

texodus:
This PR removes all non-error `message()` calls from our cmake files, so that cmake does not log its entire life's voyage.


timkpaine:
installing via `pip` without verbose flags already silences the build output

sc1f:
I think we should keep these logs - they are extremely useful for debugging on machines where cmake might pick up something from where we don’t expect, which is a lot of places

On Sep 6, 2020, at 4:34 PM, Tim Paine <notifications@github.com> wrote:

﻿

installing via pip without verbose flags already silences the build output

texodus:
👎

PR #1195, opened 2020-09-14
Server mode for Jupyterlab `PerspectiveWidget`

texodus:
* [@sc1f] Adds `server` kwarg to `PerspectiveWidget()` Jupyterlab widget.  When `True`, the browser `<perspective-viewer>` will use a python, server-side `Table()` virtually, rather than creating a WebAssembly `Table()` and cloning the contents.  This mode is great for extra-large datasets, but comes at the cost of client interactive performance, and server memory/CPU. 
* Fixes 2 discrepancies with the Javascript API:
  - Fixes missing offset (called `psp_offset` in wasm) for correcting column_only `end_col` values.
  - Adds `id` kwarg for `to_columns()`, which is equivalent to `__index__` for 0-sided contexts, and `__row_path__` otherwise, and is necessary for GUI features like persistent selection and editing.
* Fixes a longstanding bug in `perspective-jupyterlab` which caused `Create New Output` derived viewers to be mis-sized.
* Fixes `perspective-viewer-datagrid` to not swallow scroll events when fully-scrolled, for convenient notebook scrolling.
* Switches `perspective-jupyterlab` to `material-dense` theme.

Screen Shot 2020-09-14 at 4 15 30 AM

### DIFF
--- a/packages/perspective-jupyterlab/package.json
+++ b/packages/perspective-jupyterlab/package.json
@@ -25,7 +25,7 @@
         "bench:run": "echo \"No Benchmarks\"",
         "clean:screenshots": "rimraf \"screenshots/**/*.@(failed|diff).png\"",
         "test:build": "cpx \"test/html/*\" dist/umd && cpx \"test/csv/*\" dist/umd && cpx \"test/css/*\" dist/umd && webpack --color --config src/config/webpack.config.js",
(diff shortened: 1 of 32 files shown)


sc1f (approved):
Tested locally - looks good! Proper styling for the additional output view is nice.

PR #1197, opened 2020-09-17
Improve Python Examples

sc1f:
This PR fixes the `workspace-editing` example by bundling in a Python server, and improves/clarifies the `tornado-python` example with more inline comments. It also establishes a distinction between "server mode" and "distributed mode" for `perspective-python` client/server setups, which should make setting up these servers a little easier to understand conceptually.


texodus (approved):

PR #1198, opened 2020-09-18
GIL release for perspective-python Async mode

texodus:
Adds `set_loop_callback()` public method to the `PerspectiveManager` class, which takes as an argument a loop-queueing function, such as asyncio `call_soon()` and Tornado `add_callback()`.  Once invoked, Perspective will presume to be owned by the calling thread, and will release the Python GIL for the bulk of reading/writing Apache Arrow data, as well as calculating
new `View` contexts.  There are some new restrictions in this mode, namely that Perspective is no longer thread safe, covered in the documentation.

### DIFF
--- a/docs/md/python.md
+++ b/docs/md/python.md
@@ -223,6 +223,42 @@ websocket connection.
 [operate remotely](/docs/md/js.html#remote-perspective-via-perspective-python-and-tornado)
 using a websocket API.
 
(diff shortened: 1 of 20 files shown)


sc1f (approved):
Looks good! Tested locally & offline and enables some really great parallel performance over large datasets.

PR #1199, opened 2020-09-18
Add Python client implementation

sc1f:
This PR introduces `PerspectiveClient`, which mimics the Javascript client implementation by offering a fully async (`asyncio` or `yield`-based) API on top of the core `Table` and `View` public API. This is useful for testing `perspective-python` servers, as it offers programmatic access to Perspective's public API in Python.

`PerspectiveTornadoClient` is an implementation of the client API that allows users to open up remote `Table`s and `View`s on a `perspective-python` server, and treat them as if they were objects in the client runtime using either `await` or `yield`.

The interface is extremely simple, using `async` and `await`:

```python
import asyncio
from perspective.tornado_handler import websocket

current_loop = asyncio.get_event_loop()
ws = websocket("ws://.../", ioloop=current_loop)

async def run():
    table = ws.open_table("data_source")
    schema = await table.schema()
    print(schema)  # {"a": "float"}
    view = table.view(row_pivots=["a"])
    data = await view.to_arrow()
    print(type(arrow)) # bytes

current_loop.run_until_complete(run())  # or current_loop.call_soon(run)
```

or `yield`-based coroutines:

```python
from tornado import gen
from tornado.ioloop import IOLoop
from perspective.tornado_handler import websocket

current_loop = IOLoop.current()
ws = websocket("ws://.../", ioloop=current_loop)

@gen.coroutine
def run():
    table = ws.open_table("data_source")
    schema = yield table.schema()
    print(schema)  # {"a": "float"}
    view = table.view(row_pivots=["a"])
    data = yield view.to_arrow()
    print(type(arrow)) # bytes
current_loop.run_sync(run)  # or current_loop.add_callback(run)
```


python/perspective/perspective/client/client.py
```diff
@@ -0,0 +1,145 @@
...
+                handler["callback_id"] = msg["callback_id"]
+
+            self._msg_id += 1
+            self._handlers[self._msg_id] = handler
+
+        msg["id"] = self._msg_id
+
+        self.send(msg)
+
+    def table(self, data, index=None, limit=None, name=str(random())):
```
texodus, line 136: The default argument to `name` here is evaluated once when the method is created, so all proxy tables will have the same name.

python/perspective/perspective/client/table_api.py
```diff
@@ -0,0 +1,147 @@
...
+# This file is part of the Perspective library, distributed under the terms of
+# the Apache License 2.0.  The full license can be found in the LICENSE file.
+#
+from random import random
+from functools import partial
+from .dispatch import async_queue, subscribe, unsubscribe
+from .view_api import view as make_view
+
+
+def table(client, data, index=None, limit=None, name=str(random())):
```
texodus, line 14: Same, and also it's probably a good idea to just make `name` a required arg here as this `table` is private to the library.

python/perspective/perspective/client/table_api.py
```diff
@@ -0,0 +1,147 @@
...
+        self._subscribe = partial(subscribe, self._client, self._name)
+        self._unsubscribe = partial(unsubscribe, self._client, self._name)
+
+    def make_port(self):
+        return self._async_queue("make_port", "table_method")
+
+    def remove_port(self):
+        return self._async_queue("remove_port", "table_method")
+
+    def compute(self):
```
texodus, line 61: Unrelated - but I think we can deprecate this "feature flag" - its implementation returns true for JS and Python

python/perspective/perspective/tests/client/test_client.py
```diff
@@ -4,226 +4,4 @@
 #
 # This file is part of the Perspective library, distributed under the terms of
 # the Apache License 2.0.  The full license can be found in the LICENSE file.
```
texodus, line 6: Empty file

python/perspective/perspective/tests/table/test_remove.py
```diff
@@ -15,11 +15,13 @@ def test_remove_all(self):
         tbl = Table([{"a": "abc", "b": 123}], index="a")
         tbl.remove(["abc"])
         assert tbl.view().to_records() == []
+        # assert tbl.size() == 0
```
texodus, line 18: Only adds commented blocks

texodus (approved):
Thanks for the PR!  Looks great and is a super useful addition!  There is a small bug I think with table names, but I will take care of fixing these in a cleanup PR.

texodus:
Also - I believe the FOSSA check is either malfunctioning or has recently received an update to its detection criteria.  This PR specifically does not introduce dependencies, so if there are legitimate failures, they are pre-existing regardless.  Thus, I'm going to override this check.

python/perspective/perspective/client/client.py
sc1f: yup - that is a silly mistake on my part

PR #1205, opened 2020-09-25
Update script_utils.js

xingxp:


texodus:
Thanks for the effot - but no description of the issue nor test.  This is also not how the `bash` template string function works - the missing space tells the parser that `-v` should be removed when `CMD` is `undefined`.

PR #1207, opened 2020-09-25
Upgrade WebAssembly build to Arrow 1.0.1

texodus:
Upgrades Apache Arrow to v1.0.1 for all targets _except_ Python 2.7.  For the WebAssembly/Emscripten build, CSV loading and writing (via `to_csv()`) has been rewritten to use Arrow over C++, resulting in a ~10x load performance increase, depending on whether you provide a schema and what data types your columns are.

For Python 2.7, Arrow requirements are `pyarrow<2` only, which is 0.16.1.  I've done some limited testing to ensure this is cross-compatible with Arrow Table data generated or read-from 1.0.1, but this is something we should probably keep an eye on; especially regarding the JupyterLab plugin `@finos/perspective-jupyterlab`, whose client/server distributed mode would use engines compiled to both versions simultaneously when Python 2.7 runs the kernel.

Screen Shot 2020-09-24 at 2 30 43 AM


cmake/modules/FindPyArrow.cmake
```diff
@@ -24,7 +24,7 @@ execute_process(
           "from __future__ import print_function\ntry: import pyarrow; print(' '.join(pyarrow.get_libraries()), end='')\nexcept:pass"
   OUTPUT_VARIABLE __pyarrow_libraries)
 
-# And the version
+  # And the version
 execute_process(
   COMMAND "${Python_EXECUTABLE}" -c
```
timkpaine, line 29: we don't really need to do any version matching, we should just link all against arrow_python.dylib or arroy_python.so since arrow will include both a numbered and unnumbered version and we want the version picked up to be dictated by python dependency checking anyway.

cpp/perspective/CMakeLists.txt
```diff
@@ -631,8 +637,11 @@ elseif(PSP_CPP_BUILD OR PSP_PYTHON_BUILD)
 			# `site-packages`, the relative search path should be able to pick
 			# up pyarrow. This is only enabled for MacOS, as `auditwheel`
 			# will not delocate libarrow properly if it is in the rpath.
-            set_property(TARGET psp PROPERTY INSTALL_RPATH ${CMAKE_INSTALL_RPATH} ${module_origin_path} ${module_origin_path}../../pyarrow/ ${PYTHON_PYARROW_LIBRARY_DIR} ${PSP_PYTHON_ARROWINSTALLDIR})
-            set_property(TARGET binding PROPERTY INSTALL_RPATH ${CMAKE_INSTALL_RPATH} ${module_origin_path} ${module_origin_path}../../pyarrow/ ${PYTHON_PYARROW_LIBRARY_DIR} ${PSP_PYTHON_ARROWINSTALLDIR})
+			#
```
timkpaine, line 640: lots of white space

cpp/perspective/CMakeLists.txt
```diff
@@ -631,8 +637,11 @@ elseif(PSP_CPP_BUILD OR PSP_PYTHON_BUILD)
...
+			# We also include the install-time path to pyarrow so that pep-518
```
timkpaine, line 641: lots of white space

cpp/perspective/CMakeLists.txt
```diff
@@ -642,8 +651,9 @@ elseif(PSP_CPP_BUILD OR PSP_PYTHON_BUILD)
 
 		target_link_libraries(psp tbb)
 		target_link_libraries(binding tbb)
-
 		target_link_libraries(binding psp)
+		target_compile_options(psp PRIVATE -Wno-deprecated-register)
```
timkpaine, line 655: what are these for?

cpp/perspective/src/cpp/emscripten.cpp
```diff
@@ -10,6 +10,7 @@
 #include <perspective/emscripten.h>
 #include <perspective/arrow_loader.h>
 #include <perspective/arrow_writer.h>
+#include "arrow/csv/api.h"
```
timkpaine, line 13: match style with <>

python/perspective/setup.py
```diff
@@ -49,14 +50,16 @@ def which(x):
...
     "python-dateutil>=2.8.0",
     "six>=1.11.0",
     "traitlets>=4.3.2",
 ]
 
 if sys.version_info.major < 3:
     requires.append("backports.shutil-which")
+    requires.append("pyarrow<2")
+else:
+    requires.append("pyarrow>=1.0.1,<2")
```
timkpaine, line 62: this should match pyproject.toml

azure-pipelines.yml
```diff
@@ -181,7 +182,7 @@ jobs:
 #       displayName: "Which python"
 
 #     - script: |
-#         python -m pip install numpy "pyarrow>=0.16.0,<1"
+#         python -m pip install numpy "pyarrow>=1.0.1,<2"
```
timkpaine, line 185: match pyproject.toml

azure-pipelines.yml
```diff
@@ -338,8 +344,14 @@ jobs:
       displayName: "Which python"
 
     - script: |
-        python -m pip install delocate wheel numpy "pyarrow>=0.16.0,<1"
-      displayName: "Python deps"
+        python -m pip install delocate wheel numpy "pyarrow>=1.0.1,<2"
```
timkpaine, line 347: match pyproject.toml

timkpaine (changes requested):

timkpaine:
@texodus the only issue really is that `pyproject.toml` has to match wherever you install pyarrow. So if you install <2 for python2, and >1<2 for python3, you need to reflect this in the pyproject.toml file so that the build system matches the setup.py requirements matches what we install in CI

cmake/modules/FindPyArrow.cmake
```diff
@@ -24,37 +24,25 @@ execute_process(
...
 if(${CMAKE_SYSTEM_NAME} MATCHES "Windows")
   # windows its just "arrow.dll"
   set(PYTHON_PYARROW_PYTHON_SHARED_LIBRARY "arrow_python")
   set(PYTHON_PYARROW_ARROW_SHARED_LIBRARY "arrow")
   set(PYTHON_PYARROW_LIBRARIES ${PYTHON_PYARROW_PYTHON_SHARED_LIBRARY} ${PYTHON_PYARROW_ARROW_SHARED_LIBRARY})
 elseif (CMAKE_SYSTEM_NAME MATCHES "Darwin")
   # Link against pre-built libarrow on MacOS
-  set(PYTHON_PYARROW_PYTHON_SHARED_LIBRARY ${PYTHON_PYARROW_LIBRARY_DIR}/${CMAKE_SHARED_LIBRARY_PREFIX}arrow_python.${PYARROW_VERSION_MINOR}.dylib)
-  set(PYTHON_PYARROW_ARROW_SHARED_LIBRARY ${PYTHON_PYARROW_LIBRARY_DIR}/${CMAKE_SHARED_LIBRARY_PREFIX}arrow.${PYARROW_VERSION_MINOR}.dylib)
+  set(PYTHON_PYARROW_PYTHON_SHARED_LIBRARY ${PYTHON_PYARROW_LIBRARY_DIR}/${CMAKE_SHARED_LIBRARY_PREFIX}arrow_python.100.dylib)
```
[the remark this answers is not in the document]
texodus: We are going to have to cope with the conditional naming for these libraries, as they are not disted in this version under this name (and don't seem to exhibit any consistent version convention at all).

cmake/modules/FindPyArrow.cmake
```diff
@@ -45,23 +45,38 @@ if(${CMAKE_SYSTEM_NAME} MATCHES "Windows")
   # windows its just "arrow.dll"
   set(PYTHON_PYARROW_PYTHON_SHARED_LIBRARY "arrow_python")
   set(PYTHON_PYARROW_ARROW_SHARED_LIBRARY "arrow")
-  set(PYTHON_PYARROW_LIBRARIES ${PYTHON_PYARROW_PYTHON_SHARED_LIBRARY} ${PYTHON_PYARROW_ARROW_SHARED_LIBRARY})
+elseif (CMAKE_SYSTEM_NAME MATCHES "Darwin" AND ${PYARROW_VERSION_MAJOR} EQUAL "1")
+    # Link against pre-built libarrow on MacOS
+  set(PYTHON_PYARROW_PYTHON_SHARED_LIBRARY ${PYTHON_PYARROW_LIBRARY_DIR}/${CMAKE_SHARED_LIBRARY_PREFIX}arrow_python.100.dylib)
+  set(PYTHON_PYARROW_ARROW_SHARED_LIBRARY ${PYTHON_PYARROW_LIBRARY_DIR}/${CMAKE_SHARED_LIBRARY_PREFIX}arrow.100.dylib)
```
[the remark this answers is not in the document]
texodus: See above

cmake/modules/FindPyArrow.cmake
texodus: Arrow 1.0.1 does not include both such versions, as above.

cpp/perspective/CMakeLists.txt
texodus: This is because this is the only file in the project which uses spaces instead of tabs ... probably not a fix that should be included in this particular PR but I will address this in a cleanup patch.

cpp/perspective/CMakeLists.txt
texodus: Silences python2 deprecation warnings which make our build logs un-readable.

python/perspective/setup.py
texodus: This version is conditional on python version - does this not make the looser bounds `"pyarrow<2"` for `pyproject.toml` correct?  Perspective under this PR will compile with either `pyarrow`.

cpp/perspective/CMakeLists.txt
timkpaine: You're doing this on windows too though, are you sure this flag will work there?

python/perspective/setup.py
timkpaine: `pyproject.toml` specifies it as `pyarrow<2` regardless of python version. So it could satisfy with `pyarrow<0.16` even though that won't work. 

You have a couple of options, something like:
`"pyarrow>0.16.0,<2; python_version < '3'", "pyarrow>1.0.1,<2; python_version >= '3'"` would work ok.

Stylistically, every place where we specify requirements should match exactly the behavior of pyproject.toml, otherwise we can very easily start to break our builds and wheels like the state we were in before.

cpp/perspective/CMakeLists.txt
texodus: I don't know ... our windows azure builds do not work ...

texodus:
@timkpaine I've moved the `pyarrow` upgrade to [a new PR](https://github.com/finos/perspective/pull/1215).  What's left in the PR is just the C++ and Javascript changes necessary to upgrade, plus the new CSV parser.  Since this was already compilable for `ARROW_VERSION_MAJOR < 1`, the PR continues to compile fine with `perspective-python`'s `pyarrow` dependency bounds as they were.

timkpaine:
@texodus do you want me to make the python changes on that branch or do you want them in a separate PR?

cmake/modules/FindPyArrow.cmake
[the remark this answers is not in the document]
timkpaine: Let's link this into an apache arrow issue, for pyarrow < 1 they had symlinks and we'll need those for some platforms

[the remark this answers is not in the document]
timkpaine: https://github.com/apache/arrow/pull/7334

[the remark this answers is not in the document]
timkpaine: @wesm this makes linking against these libraries super ugly. Since the version is managed by python anyway (e.g. in the version of the python package), wouldn't it be better to ship just the unversioned binary (e.g. libarrow.dylib)?

texodus:
I already made most of them - just the `pyproject.toml` one is left.

cmake/modules/FindPyArrow.cmake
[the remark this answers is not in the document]
wesm: See http://arrow.apache.org/docs/python/extending.html#building-extensions-against-pypi-wheels. We had to make changes to prevent shared libraries from being duplicated in the wheels. It might be possible to change things to ship unversioned shared libraries instead, but it will require some surgery on the Cython modules to get them to link with the unversioned libs (because they look for the libraries with the ABI version when they link). Feel free to open a JIRA issue in Apache Arrow with a proposal -- I have definitely hit the limit for how much time I can personally spend on it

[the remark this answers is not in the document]
texodus: Thanks @wesm

@timkpaine I am not particularly bothered by this, it is easy to fix and this file must be manually updated for new Arrow versions regardless.

PR #1209, opened 2020-09-28
Send large arrow binaries as chunks, add client-level Websocket heartbeat

sc1f:
This PR implements chunked sending/chunked loading of binaries in Perspective's websocket implementations in both Python and Javascript. Apache Arrow binaries above a certain size will be sent to the Perspective client in chunks, which prevents the Websocket from locking on a particularly large binary and closing due to a timeout, as once that binary is on the wire all heartbeat messages/other messages are blocked from being sent/received. 

The default chunk threshold is configured to be 200MB, and arrows will be transported in 50MB chunks. In Python, these settings can be altered in the `PerspectiveManager` constructor by initializing a new `PerspectiveManager` with `chunk_threshold` and `chunk_size`, which are defined in bytes:

```python
manager = PerspectiveManager(
    chunk_threshold=500 * 1000 * 1000,  # in bytes
    chunk_size=100 * 1000 * 1000
)
```

In `NodeJS`, these settings can be altered on the `WebsocketManager`:

```javascript
const manager = new WebsocketManager();
manager.chunk_threshold = 500 * 1000 * 1000;  // in bytes
manager.chunk_size = 100 * 1000 * 1000;
```

Additionally, this PR implements a client-level heartbeat message (ping/pong) between Perspective client and server, which should aid in debugging as heartbeat messages now show up in the browser debug console. In Tornado, servers previously set up with `websocket_ping_interval` should remove their `websocket_ping_interval` setting and depend on the client-level heartbeats to keep the Perspective websocket connection alive.

### Changelog
- Implements chunked binary sending on `PerspectiveTornadoServer` in Python and `WebSocketManager` in JS
- Implements chunked binary receiving in `WebSocketClient` in JS and `PerspectiveTornadoClient` in Python
- Adds tests for chunked send/receive in Python
- Adds ping/pong heartbeat for Websockets

### DIFF
--- a/examples/tornado-python/server.py
+++ b/examples/tornado-python/server.py
@@ -59,8 +59,7 @@ def make_app():
                 tornado.web.StaticFileHandler,
                 {"path": "./", "default_filename": "index.html"},
             ),
-        ],
-        websocket_ping_interval=15,
+        ]
     )
 
 
--- a/packages/perspective-jupyterlab/src/ts/client.ts
+++ b/packages/perspective-jupyterlab/src/ts/client.ts
@@ -6,6 +6,7 @@
  * the Apache License 2.0.  The full license can be found in the LICENSE file.
  *
  */
+/* eslint-disable @typescript-eslint/camelcase */
 
 import {DOMWidgetView} from "@jupyter-widgets/base";
 import {Client} from "@finos/perspective/dist/esm/api/client";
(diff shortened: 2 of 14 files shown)


texodus:
Combined `chunk_size` and `chunk_threshold` kwargs, as per review.  I've additionally added the private `_chunk_sleep` for debugging, though this has been set to 0 by default [see here](https://www.tornadoweb.org/en/stable/gen.html#tornado.gen.moment)

Fix & test also for an issue with asynchronous logic, namely that once a client is in "pending_binary" state, it cannot process other messages.  This is only sometimes an issue, as clients tend to serialize themselves by `await`-ing responses to queries;  when it occurs though, the client will hang.  Note this change does not affect `"ping"` messages, since they are processed before the `post()` locking and response dispatch logic is run.

Before:  a call to `num_rows()` after an un-`await`ed call to `to_arrow()` interleaves the response, between messages 1 and 2 of a 3 message binary stream.

Screen Shot 2020-10-05 at 11 09 38 PM

After:  `num_rows()` is deferred until the strea is finished, in an elaborated `_chunk_sleep`.  The response to `num_rows()` is deferred until the stream is complete.

Screen Shot 2020-10-07 at 2 16 00 AM

texodus (approved):

PR #1212, opened 2020-10-01
Add `perspective-plugin-update` event for d3fc

zemeolotu:


texodus (approved):

PR #1214, opened 2020-10-04
Fix `computed-column` and `table()` constructor Javascript bugs

texodus:


timkpaine:
Thanks for the PR!

A description, even brief, would be helpful though!

PR #1223, opened 2020-10-12
Add Python client-based stresstest

sc1f:
This PR adds a new stresstest module for `perspective-python` using the new `PerspectiveTornadoClient`, which is a far superior way to write Perspective stresstests and remote clients in Python. Instead of measuring per-message time, the new stresstest allows the user to write arbitary tasks and measure the progress of those tasks over multiple concurrent clients.

To use the new stresstest, run `yarn bench -h` to see the available options and configurations. Examples are located in `python/perspective/bench/tornado`, and can be run as follows:

```bash
yarn bench python/perspective/bench/tornado/distributed_mode.py -c10 ws://localhost:8080/websocket
```

### Changelog
- Added new Python client-based stresstest suite
- Removed all code referring to old, custom-written stresstest suite
- Split Python benchmarks into `runtime` and `tornado`
- Added new-style benchmarks for distributed mode and server mode


texodus (approved):
Looks good!

texodus:
Thanks for the PR!

PR #1224, opened 2020-10-15
Improve `datetime` parsing performance

texodus:
Improves datetime parsing performance by discarding format string candidates per-column as they fail on non-empty strings, which has a nice performance boost across the board for data types that use string datetime encodings.

Screen Shot 2020-10-15 at 4 56 12 AM


texodus:
After further review, this approach is irreconcilable.  Back to the drawing board!

PR #1227, opened 2020-10-19
Explicitly import `tornado.locks`

sc1f:
`tornado.locks` is not included in the `__init__.py` of older Tornado versions, and this causes an issue when hosting Perspective servers as we don't explicitly import `tornado.locks`. This patch makes the import explicit and pins a version of Tornado for our Python 2 builds.

### DIFF
--- a/python/perspective/setup.py
+++ b/python/perspective/setup.py
@@ -53,13 +53,12 @@ requires = [
     "python-dateutil>=2.8.0",
     "six>=1.11.0",
     "traitlets>=4.3.2",
+    "tornado>=4.5.3",
 ]
 
 if sys.version_info.major < 3:
-    requires.append("backports.shutil-which")
-    requires.append("tornado==4.5.3")
-else:
-    requires.append("tornado>=4.5.3")
+    requires.pop()
+    requires += ["tornado==4.5.3", "backports.shutil-which"]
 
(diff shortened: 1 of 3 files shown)


python/perspective/setup.py
```diff
@@ -53,10 +53,12 @@ def which(x):
     "python-dateutil>=2.8.0",
     "six>=1.11.0",
     "traitlets>=4.3.2",
+    "tornado>=4.5.3",
```
timkpaine, line 56: needs to match pyproject.toml

python/perspective/setup.py
```diff
@@ -53,10 +53,12 @@ def which(x):
...
 ]
 
 if sys.version_info.major < 3:
-    requires.append("backports.shutil-which")
+    requires.pop()
+    requires += ["tornado==4.5.3", "backports.shutil-which"]
```
timkpaine, line 61: don't hard equals here

timkpaine (changes requested):

python/perspective/setup.py
timkpaine: also keep alphabetical

timkpaine (approved):

texodus:
Thanks for the PR!

PR #1228, opened 2020-10-21
Add unit context for 0-sided Views, remove cell delta, refactor C++

sc1f:
This PR enhances the performance of certain 0-sided Views in Perspective by 2x-10x, depending on data size.

Each `View` is backed by a context object, which maintains its own traversal of the underlying master table. This traversal allows the user to read data out based on the order of primary keys, allows for pivots to traverse the underlying datasets, allows for sorts to be applied to the subset of data in a context, etc. In the case where the context maintains a basically _trivial_ traversal, where the order that it reads data out is equivalent to the order data is stored in the underlying table, and when it does not have to apply any sorts, filters, or computed columns, we can skip the creation of a traversal entirely, and avoid the overhead of storing primary keys, sorting them, and converting row indices to primary keys.

The unit context is a context object that has no traversal and reads directly from the underlying master table of the gnode. Internally, it offers the same API as all other context types, and all construction around unit contexts occurs in internal code and has no bearing to the public API.

Externally, the unit context offers a massive performance improvement in a large use case—when the View has no pivots, sorts, filters, or computed columns, and the Table does not have a user-specified index. On a Table with a user-specified index, data must be read out in the same order as primary keys, which is different from the underlying stored order in the master table. However, this PR will allow for future improvements to this behavior.

Additionally, this PR contains a few performance-based refactors, including the removal of cell delta/step delta from 0-sided contexts and the public API, use of direct vector indexing whenever possible, and the removal of minmax calculations (which caused a 30-40% decrease in memory allocated per `View`.

### Changelog
- Add and implement unit context for JS and Python
- Multitude of tests for unit context in different scenarios to ensure correct behavior
- Change instances of `push_back` to `reserve()` and direct vector indexing whenever possible, i.e. the vector length is always known at runtime
- Removes minmax calculation from all contexts, which used to be still enabled for Python but disabled for Javascript
- Removes {mode: "cell"} from `on_update`, as row deltas are faster by orders of magnitude, and provide for the same function—a cell delta and a row delta from the same update will create equivalent updates when applied to a new table.
- Splits the Javascript websocket implementation into two files - `client` and `manager`, which allows it to more closely mirror the organization of the Python Tornado handler.

### Benchmarks

Javascript benchmarks show a massive improvement in View creation time, and slight improvements in serialization time and time to create a delta.

Screen Shot 2020-10-16 at 1 44 55 PM

In Python, where I've benchmarked this PR against much larger datasets (5m rows), the performance of `view()` is almost equivalent to the performance of `open_view()`, which simply provides a handle to an already-created view on the server. Over large datasets and multiple, parallel clients, the unit context massively reduces the overhead of `view()`, resulting in a 5x-10x improvement in performance over a regular `ctx0`.


sc1f:
Closing and replacing with two smaller PRs that are more logically segmented.

PR #1230, opened 2020-10-22
Add `get_index()` and `get_limit()` to the public API, internal code cleanup

sc1f:
This PR adds `get_index()` and `get_limit()`, which allows for Table consumers to access the value of user-provided index columns or limits, or `null` if not set by the user. This is useful in the distributed Perspective use case, as clients need to mirror the Table exactly:

```javascript
const remote_table = websocket.open_table("table");
const remote_view = remote_table.view();

const arrow = await remote_view.to_arrow();
const index = await remote_table.get_index();
const client_table = worker.table(arrow, {index: index});
```

Additionally, this PR does a little internal cleanup of Perspective core code in both Javascript and Python, and adds a large amount of new tests.

### DIFF
--- a/packages/perspective/src/js/api/table_api.js
+++ b/packages/perspective/src/js/api/table_api.js
@@ -107,8 +107,6 @@ table.prototype.columns = async_queue("columns", "table_method");
 
 table.prototype.clear = async_queue("clear", "table_method");
 
-table.prototype.on_clear = subscribe("on_clear", "table_method", true);
-
 table.prototype.replace = async_queue("replace", "table_method");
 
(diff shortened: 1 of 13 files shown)


texodus (approved):

I don't know that I am completely sold on `get_index` and `get_limit` vocabulary specifically.  Since these properties are _immutable_, they could be set directly on the proxy `table.index`, `table.limit`, though it introduces some race-y behavior with initialization due to the `table.view()` returning a proxy synchronously.  For now though, this method-based design is simple to encode in our current client code-generators!

PR #1231, opened 2020-10-23
Restrict `viewer.load()` to only accept `Table()`

texodus:
This PR removes `@finos/perspective` as a build-time dependency of `@finos/perspective-viewer`, which in practice means 
* `perspective.wasm` WebAssembly binary and WebWorker JS runtime are no longer packed with the `@finos/perspective-viewer` module, or any module that imports `@finos/perspective-viewer`.  You must now also import `@finos/perspective` explicitly.
* `<perspective-viewer>`'s `load()` method now _only_ allows arguments of type `Table()`, which must be created explicitly from a WebWorker `worker()` or Websocket `websocket()`.
* Reciprocal `Table()` methods on `<perspective-viewer>` have been removed: `replace()`, `clear()`, `update()`.  Use e.g. `viewer.table.clear()` instead.


packages/perspective-jupyterlab/src/ts/view.ts
```diff
@@ -305,7 +314,12 @@ export class PerspectiveView extends DOMWidgetView {
                 } else {
                     // Just load the view into the widget, everything else
                     // is handled.
-                    this.pWidget.load(kernel_view, table_options);
+
+                    // TODO this should lilely be removed
+                    kernel_view.to_arrow().then((arrow: ArrayBuffer) => {
```
sc1f, line 319: This breaks - there is no `on_update` callback so when the table on the kernel updates, it will not reflect in the widget.

packages/perspective-jupyterlab/test/js/unit/view.spec.js
```diff
@@ -365,8 +372,14 @@ describe("PerspectiveView", function() {
...
-                expect(load_args[0]).toEqual(data);
-                expect(load_args[1]).toEqual(options);
+                const result = await load_args[0].view().to_columns();
+                result.b = result.b.map(x => new Date(x));
+
+                expect(result).toEqual(data);
+
+                // TODO there is no way to verify this in perspective API
+                // currently ...
+                // expect(load_args[1]).toEqual(options);
```
sc1f, line 382: We can use get_index() or get_limit() here to make sure it's been applied properly.

sc1f (changes requested):
`PerspectiveWidget` is currently broken on this branch - verified locally and through code review.

Intuitively, we should also make sure that `PerspectiveWidget` follows this API change - i.e. it should only `__init__` or `load` with a table and not a dataset.

sc1f (approved):
Looks good! This should be merged before #1240 - there will be some conflicts on that PR (in the jupyterlab plugin and tests) that I will fix after this branch is merged.

packages/perspective-jupyterlab/src/ts/psp_widget.ts
```diff
@@ -152,14 +152,14 @@ export class PerspectiveWidget extends Widget {
      * @param data
      */
     _update(data: TableData): void {
-        this.viewer.update(data);
+        this.viewer.table.update(data);
```
timkpaine, line 155: `TypeError: this.viewer.table is undefined`

timkpaine: this introduces a bug as `table` isn't guaranteed to exist on the viewer. I don't think this was checked?

PR #1233, opened 2020-10-23
Refactor C++, misc. C++ fixes

sc1f:
This PR contains a few performance-based refactors: the removal of cell delta/step delta from 0-sided contexts and the public API, use of direct vector indexing whenever possible, and the removal of minmax calculations (which caused a 30-40% decrease in memory allocated per View.

### Changelog
- Change instances of push_back to reserve() and direct vector indexing whenever possible, i.e. the vector length is always known at runtime
- Removes minmax calculation from all contexts, which used to be still enabled for Python but disabled for Javascript
- Removes {mode: "cell"} from on_update, as row deltas are faster by orders of magnitude, and provide for the same function—a cell delta and a row delta from the same update will create equivalent updates when applied to a new table.
- Adds tests around restore and clear when computed columns are created
- Splits the Javascript websocket implementation into two files - client and manager, which allows it to more closely mirror the organization of the Python Tornado handler.


texodus (approved):

PR #1240, opened 2020-10-29
Remove host_view and open_view from public API

sc1f:
This PR removes `host_view` and `open_view` from the public API.

With the improvements in the speed to create a flat view in #1235, the difference between using a `View` created on a server and creating a `View` on the client using a handle to the server's table is minimal, if non-existent in most use-cases. This allows us to remove the `host_view` API, which doesn't fit in very well with our established public API semantics of "table as data container, view as query and data serializer."

'host_view' shoe-horns in the ability to access a `View` over the network so that we can use it to create a table, and makes the explanations around distributed/server/client mode more complicated. Because the old context implementation added a small overhead (scaling linearly with the size of the underlying data), `host_view` made sense for performance reasons.

With the removal of `host_view` and `open_view`, Perspective's remote API only allows the hosting and opening of `Table` objects. This makes it far easier to reason about, especially when it comes to the distinctions between distributed, server, and client mode while maintaining the same feature set in the public API.

Both distributed and server mode is easier to implement:

```python
manager = PerspectiveManager()
manager.host_table("data_source", TABLE)
```

```js
const worker = perspective.worker();
const websocket = perspective.websocket(URL);

// Open the table on the server - this is all that is needed for server mode
const remote_table = websocket.open_table("data_source");

// For distributed mode, create a view
const view = remote_table.view();

// Pass it into the JS table constructor
const client_table = worker.table(view);
```

It is also trivial to host subsets of a table based on already-created views:

```python
# Table with large dataset that might be too big to transmit in one go
table = perspective.Table(big_dataset)

# Create a view on a subset of the data
filtered_view = table.view(filter=[["Sales", ">", 100]])

# Create a table using the filtered view
filtered_table = perspective.Table(filtered_view.to_arrow())

# Add an `on_update` callback to keep the tables in sync
def updater(port_id, delta):
    filtered_table.update(delta)

filtered_view.on_update(updater, mode="row")

# Easy to host the tables
manager.host_table("data_source", table)
manager.host_table("data_source_filtered", filtered_table)
```

### Changelog
- Remove `host_view` and `open_view` from JS and Python
- Refactor Jupyterlab widget to create a new view instead of using `open_view`
- Fix regression in `PerspectiveWidget.delete()`to delete all views created on widget's manager instance before deleting table
- Update examples to use `open_table` and view creation
- Update documentation to include distributed/server mode in Python docs

### DIFF
--- a/packages/perspective-jupyterlab/src/ts/view.ts
+++ b/packages/perspective-jupyterlab/src/ts/view.ts
@@ -242,21 +242,30 @@ export class PerspectiveView extends DOMWidgetView {
         const table_options = msg.data["options"] || {};
 
         if (this.pWidget.client) {
-            // In client mode, retrieve the serialized data and the options
-            // passed by the user, and create a new table on the client end.
+            /**
+             * In client mode, retrieve the serialized data and the options
+             * passed by the user, and create a new table on the client end.
+             */
             const data = msg.data["data"];
             this.pWidget.load(data, table_options);
         } else {
             if (this.pWidget.server && msg.data["table_name"]) {
-                // Get a remote table handle, and load the remote table in
-                // the client.
+                /**
+                 * Get a remote table handle, and load the remote table in
+                 * the client for server mode Perspective.
+                 */
                 const table = this.perspective_client.open_table(msg.data["table_name"]);
                 this.pWidget.load(table, table_options);
-            } else if (msg.data["table_name"] && msg.data["view_name"]) {
-                // Get a remote view handle from the Jupyter kernel, and
-                // create a client-side table using the view handle.
+            } else if (msg.data["table_name"]) {
+                /**
+                 * Get a remote table handle from the Jupyter kernel, and
+                 * create a view on that handle to run Perspective in
+                 * distributed mode.
+                 */
                 const kernel_table: Table = this.perspective_client.open_table(msg.data["table_name"]);
-                const kernel_view: View = this.perspective_client.open_view(msg.data["view_name"]);
+                const kernel_view: View = kernel_table.view();
+
+                console.log(kernel_view);
 
                 // If the widget is editable, set up client/server editing
                 if (this.pWidget.editable) {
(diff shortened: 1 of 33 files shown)


packages/perspective-jupyterlab/src/ts/view.ts
```diff
@@ -242,21 +242,28 @@ export class PerspectiveView extends DOMWidgetView {
...
                 this.pWidget.load(table, table_options);
-            } else if (msg.data["table_name"] && msg.data["view_name"]) {
-                // Get a remote view handle from the Jupyter kernel, and
-                // create a client-side table using the view handle.
+            } else if (msg.data["table_name"]) {
+                /**
+                 * Get a remote table handle from the Jupyter kernel, and
+                 * create a view on that handle to run Perspective in
+                 * distributed mode.
+                 */
```
texodus, line 264: I can't (AFAIK) enforce this in our lint rules, but please use `//` exclusively for non-docstring comments.
Specifically, the `/** .. */` format has a specific [historical context of referring to a docstring](https://en.wikipedia.org/wiki/Javadoc), so this extra confusing to ex-java developers like myself!

packages/perspective/README.md
```diff
@@ -31,6 +31,8 @@ For more information, see the
         * [.remove_delete(callback)](#module_perspective..view+remove_delete)
     * [~table](#module_perspective..table)
         * [new table()](#new_module_perspective..table_new)
+        * [.get_index()](#module_perspective..table+get_index)
+        * [.get_limit()](#module_perspective..table+get_limit)
```
texodus, line 35: Though I know I have not done this consistently in the past, I will make sure the docs are disted as part of the release cycle so we do not need these inter-mingled in our everyday review process.

texodus (changes requested):
Thanks for the PR!  I would like to see 2 small changes before merging, which I'll add myself on this PR in a follow-up:

* Rename `host_table()` and `open_table()` to `host()` and `open()` respectively.
* Split out the autogenerated (docs) and unrelated lint-fix changes into a separate commit.  I've done a poor job of keeping on top of both of these lately but I aim to fix this!

texodus:
Thanks for the PR!

PR #1031, opened 2020-04-30
Add abs sum as an aggregator

harsh-jindal:
Fixes #750

PR #1246, opened 2020-11-04
Exposing abs sum

zepaz:
Fixes: #1031 
Changes
- Exposes abs sum to the perspective-viewer
- Added tests for abs sum


zepaz:
tests failing

PR #1249, opened 2020-11-07
Fix #1213 - improve render warning behavior

### LINKED ISSUE #1213: Improve the behavior of the render warning on charts
## Feature Request
The render warning shown for large datasets on charts obstructs the chart and often reappears after its initial dismissal making data exploration while viewing a chart a bit tedious - causing the user to dismiss it repetitively.

This is a request to improve this behavior.

### Description of Problem:
As a user viewing a chart
When my dataset is large 
And I select a chart
Then I see the "Rendering {x} points ... render all points" warning
When I click the "X" to dismiss the warning
Then it closes the warning
When I change a config option
Then the warning shows again obstructing the chart

I do NOT want to be hassled by the "Rendering {x} points..." warning reappearing

### Potential Solutions:

Make the dismissal of the warning persist for the session or the time that the chart is visible.
Make it less obstructive.

sc1f:
This PR fixes #1213 by improving the behavior of `perspective-viewer`'s render warnings, which display when the chart is attempting to render a dataset that is larger than specified bounds and might have performance implications. 

As outlined in #1213, render warnings would keep popping up even after a user has dismissed them, and it becomes an annoyance for the user. Additionally, the presence of the "Close" button allows there to be a state where the user has closed the warning that a subset of the data is being displayed, but if the viewer's config is not changed then there are no other indications that the user is looking at a subset of the data and not the whole.

This PR changes the behavior of the render warning in the following ways:

- Render warnings are now persistent per-plugin: after a user clicks "Render all points", the warning will not appear again for that plugin, even after the user switches to a different plugin.
- The "Close" button has been removed, so that there is always a warning present when the user is looking at a subset of data.
- A new `render_warning` option has been added to each plugin config, which allows the user to specify whether the warning should be shown at all (defaults to `true`):

```javascript
const plugin = window.getPlugin("d3_heatmap");

// Change the bounds at which the warning should appear
plugin.max_cells = 10000;
plugin.max_columns = 200;

// Or disable the warning entirely
plugin.render_warning = false
```
- A new set of tests around the render warning has been added to several chart types in `perspective-viewer-d3fc`.


texodus (approved):
Looks good!  

I think this implementation is an improvement over whats on `master` currently, a few things stick out to me re: the UX itself, that I'd like to answer in a future PR:
* The `X` ("dismiss the warning") button does indeed remove any indication the visualization is compromised, but the actual current implementation crops the top 30px of the view.  I can buy not wanting this to go away, but if its going to be persistent I think it needs a redesign - maybe integration with the upcoming status bar?
* Is `render_warning` keyword a more obvious implementation of this feature than, e.g., `max_cells: Infinity` ?  In essence, `max_cells` is a conceit to Perspective's somewhat asymmetric performance w.r.t. columns vs rows, so I can imagine a future simplification of this feature that only exports `max_cells`.  At minimum, `render_warning` and `max_*` are exclusive and might benefit from some validation (and maybe also default to `true` 💯 ).  Obviously, as this is a developer facing API, we have some room to iterate here!

PR #1250, opened 2020-11-09
Add perspective-click event to grid

### LINKED ISSUE #1076: perspective-click event on perspective-viewer-datagrid
Hi guys,

Unless I am mistaken, the `perspective-click` event is not supported with `perspective-viewer-datagrid`, as it was previously with `perspective-viewer-hypergrid`. I tried searching for references to `perspective-click` in the repo and also confirmed it with replacing the hypergrid script with datagrid in one of the [linked examples](https://bl.ocks.org/texodus/08fb5f1afccbd33e333453dc70db88ea).

Should [the docs](https://perspective.finos.org/docs/md/js.html#click-events) be updated accordingly (as it might be misleading to new users, considering hypergrid is legacy) or should a similar change to [this PR](https://github.com/finos/perspective/pull/1023) be made?

Thanks :)

Sam-Ogden:
Fixes #1076 
This PR adds the perspective-click event to perspective-viewer-datagrid as specified in [the docs](https://perspective.finos.org/docs/md/js.html#click-events).

### DIFF
--- a/packages/perspective-viewer-datagrid/src/js/click.js
+++ b/packages/perspective-viewer-datagrid/src/js/click.js
@@ -1,0 +1,34 @@
+/******************************************************************************
+ *
+ * Copyright (c) 2017, the Perspective Authors.
+ *
+ * This file is part of the Perspective library, distributed under the terms of
+ * the Apache License 2.0.  The full license can be found in the LICENSE file.
+ *
+ */
+
+import getCellConfig from "./getCellConfig";
+
+async function clickListener(table, viewer, event) {
+    const meta = table.getMeta(event.target);
+    if (!meta) return;
+    const {x, y} = meta;
+
+    const {row, column_names, config} = await getCellConfig(this, y, x);
+
+    viewer.dispatchEvent(
+        new CustomEvent("perspective-click", {
+            bubbles: true,
+            composed: true,
+            detail: {
+                row,
+                column_names,
+                config
+            }
+        })
+    );
+}
+
+export function configureClick(table, viewer) {
+    table.addEventListener("click", clickListener.bind(this, table, viewer));
+}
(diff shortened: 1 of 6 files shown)


sc1f (approved):
Tested locally - looks great, works well for arbitrary pivots and filters. Thanks!

KrishnaPG:
Came across this while facing a problem with click. The click event is not being raised. Used the below code from the [documentation)(https://perspective.finos.org/docs/md/js.html), but event is not getting triggered.
````
elem.addEventListener("perspective-click", function(event) {
    var config = event.detail.config;
    elem.restore(config);
});
````
Then came across this PR. Wondering if the `click` functionality is available yet. If not, could some one kindly merge this PR if this is what enables that click functionality specified in the documentation?

**Comment by texodus** [comment]
Thanks for the PR!  Apologies for the delay in merging, this is a much-requested feature addition.

Following the pattern of existing `perspective-*` events for `click` is a natural addition to the existing API.  Orthogonal to this PR (and also completely undocumented ..), ideally this would not be necessary, as we use the  [Custom Elements API](https://developer.mozilla.org/en-US/docs/Web/Web_Components/Using_custom_elements) to avoid needing to recreate DOM-like functionality.  `click` is a good example of this:

```javascript
const viewer = document.querySelector("perspective-viewer");
const table = document.querySelector("perspective-viewer regular-table");
viewer.addEventListener(x => {
    console.log(table.getMeta(x.target));
});
```

As it stands however, this is not particularly useful for a few reasons:

1) You must do this separately for each plugin type.
2) It is difficult to get the cell "config", a metadata dictionary of a single coordinate in a `View`.

For the latter at least, I propose `getCellConfig()` should move to `@finos/perspective` and perhaps undergo a name change.

PR #1253, opened 2020-11-24
Update Azure Windows boost to 1.72

sc1f:
This PR fixes the Windows build on Azure by updating the version of boost to 1.72, which is the only pre-installed version after [this PR](https://github.com/actions/virtual-environments/issues/1847) on `virtual-environments`.

Keeping a link to the [software installed on WIndows 2016 image](https://github.com/actions/virtual-environments/blob/main/images/win/Windows2016-Readme.md) for future reference.

**@texodus please merge this first and make sure builds work for subsequent PRs.**

PR #1260, opened 2020-12-06
Fix build for pyarrow 2.0.0

timkpaine:
Defaults to pyarrow>1 naming conventions. 

Merge #1253 first


texodus (approved):

I propose it's time we ditch the pyarrow-linking experiment, now that the format is stable and we are several Arrow major releases into its lifecycle.  I continue to be amazed that this worked, but at this point we stand to reduce build complexity significantly at a minor library size cost (in Python only).

PR #850, opened 2019-12-16
Implement `to_arrow` in C++ for JS/Python

sc1f:
This PR uses the C++ Apache Arrow library to serialize Perspective views into the Arrow format. It includes:

- *Upgrade Arrow to `0.15.1`*, which requires a rebuild of Docker images.
- Serialize to arrow for 0, 1, and 2-sided contexts, respecting start/end row and column.
- Generate row delta (changed rows after `update()`) using C++ to_arrow, and implemented in Python.
- Fix issue with `WebsocketClosedError` breaking in streaming example - clients closing websocket should no longer break lingering callbacks.
- Refactor `PerspectiveManager` to use `_PerspectiveCallbackCache` instead of a dictionary.
- add `remote.py` example in Python, which uses `to_arrow` and `row_delta` to stream changed rows to clients every millisecond.
- Splitting `arrow.cpp` into `arrow_loader` and `arrow_writer` for cleanliness.
- Removes `scalar_vec_to_val` and `scalar_vec_to_string` methods in C++ - use `scalar_to_val` instead.

PR #755, opened 2019-10-07
Apache Arrow native reads for JS+Python

texodus:
Ports Apache Arrow reading to use the C++ (rather than the TS) version of the library.  In addition to significant load time performance improvements, this makes arrow loading consistent & portable between Python & JS.

PR #1261, opened 2020-12-08
Add Rust client-side data accessor to read Apache Arrow binaries

sc1f:
This PR adds `arrow-accessor`, a small library written in Rust and compiled to WASM using [wasm-pack](https://github.com/rustwasm/wasm-pack) so that it can be used as part of the broader Perspective.js library inside the Datagrid and the D3FC plugins.

Using `arrow-accessor`, the Datagrid and D3FC plugins will be able to retrieve serialized data from the Perspective engine using `to_arrow()`, which returns an Arrow-formatted binary of the requested dataset. The `arrow-accessor` library, running in WASM, reads the Arrow binary and returns a 2D Javascript array containing the arrow binary’s dataset, ready to be rendered by the plugin.

This allows Perspective’s visualization plugins to take advantage of `to_arrow`’s 10x speed improvement over `to_columns` and `to_json`, which are currently used by the plugins to request data from the engine. Moreover, [async mode](https://perspective.finos.org/docs/md/python.html#async-mode) in `perspective-python` enables Arrow generation without holding the GIL, allowing Perspective servers to serve parallel requests faster.

Because this is the first exploration of using `to_arrow` as a data provider for Perspective's visualization plugins, more work is needed on the engine in order to fully use Arrow across the viewer. However, this is a huge step forward for the library as it explores Rust—not only in standalone modules compiled to WebAssembly, but compiled alongside existing C++ as part of the core engine.

### Why Rust?

The Perspective codebase already contains massive amounts of code spread across C++, Javascript, and Python. For the time being, we do not have plans to port Perspective into Rust or offer a Perspective API in Rust in the fashion of `perspective-python`.

In our opinion, however, Rust is extremely attractive for two main use cases:

- Small, standalone modules compiled to WASM that allow for performance increases in Perspective's Javascript library
- Adding new features into the core engine using Rust's FFI to allow compilation of Rust and C++ into the same shared libraries.

This PR and the attached changes clearly demonstrate that Rust is more than suitable for the first use case. It has actively-maintained and well-documented WebAssembly bindings, a user/developer base interested in Rust and WASM, and—when compared to the setup required for our Emscripten build—a massively simplified build/test/deploy process that does not require Docker or system dependencies outside of Rust itself. Moreover, Apache Arrow's Rust library can be used in WASM with zero setup, which is a massive win considering the effort involved in getting Arrow's C++ library to build under Emscripten. In short, Rust allows us to explore optimizations in our Javascript libraries which were previously infeasible or overly complicated using Emscripten.

What remains to be fully explored is the use of Rust's C FFI and how it integrates into Perspective's C++ core library. Because we heavily use features from C++11 and above, there may be difficulties in consuming Perspective's existing C++ data structures within Rust, or in creating features in Rust that return more complex structures than allowed by the FFI. Some early work has been done (though not inside this PR) on integrating Rust into our CMake build, and we can successfully compile Rust and C++ into the same WASM library, and call functions from Rust in our C++ code.

Without getting into the depths of the language debate, my personal experience working with Rust for the first time has been overwhelmingly positive, and I see Rust's potential for encouraging contributions from the open-source community into parts of Perspective's codebase that might have been previously inaccessible, given Perspective's complexity.

### Integrating Rust into Perspective

Though `arrow-accessor` has been tested locally in Perspective's datagrid plugin, I chose not to include changes to the existing codebase inside this PR. In terms of formalizing the Rust integration, I want to spend some time figuring out the best way to not only build/test Rust as part of our current build (both locally and in CI), but also how we consume the compiled WASM asset and JS wrapper in a way that does not add unnecessary complexity to the Perspective build for developers.

For users, I except nothing to change—Perspective will still be trivially installable from NPM, consumable from CDN, and wheels will continue to be deployed to PyPi. In this sense, integrating Rust is an "internal" change, but the performance benefits and developer experience will be substantially improved.

### Next Steps

- [x] better integration into Perspective's build system
  - [x] integration into `yarn setup`
  - [x] linting/autoformatting
- [ ] `to_arrow` feature parity: row paths, implicit index, pkeys
- [ ] Build/test Rust as part of Perspective WASM build and CI
- [ ] Figure out the best way to load Rust's WASM and wrapper JS generated by wasm-pack
- [ ] Integrate `arrow-accessor` into Datagrid and D3FC
- [ ] Explore Rust C FFI and determine how much integration can be done with the C++ codebase


timkpaine:
we should probably add linting/autoformatting to the extent that it matches our other languages

rust/perspective/arrow_accessor/tests/util.rs
```diff
@@ -0,0 +1,332 @@
...
+        Ok(batch) => Box::new(batch),
+        Err(err) => panic!(err)
+    }
+}
+
+// Generate an Arrow record batch with null values in columns, only including
+// a string column if with_string_column is true as the StreamWriter doesn't
+// seem to write dictionary arrays properly - they cannot be parsed in
+// Rust or in pyarrow at the moment.
+pub fn make_arrow_nullable(with_string_column: bool) -> Box<RecordBatch> {
```
nevi-me, line 176: https://github.com/apache/arrow/commit/5ed4695c4d08a60d1dec451182bde33d14b00161 might address this issue. IF you can, may you please check with the latest HEAD. Otherwise this is an issue that we should address before the next Arrow release.

nevi-me (commented):
Hi Perspective developers, thought I'd chime in as I'd be interested in seeing more Rust in great projects like this.

Please feel free to ignore my comments if they are irrelevant

sc1f:
> Hi Perspective developers, thought I'd chime in as I'd be interested in seeing more Rust in great projects like this.

Thank you for the review @nevi-me - exactly the kind of feedback we are looking for! Still definitely a work in progress, but your review helps out a ton.

rust/perspective/arrow_accessor/src/lib.rs
```diff
@@ -0,0 +1,102 @@
...
+        .map(|batch| {
+            // Panic if it can't read the batch
+            Box::new(batch.unwrap())
+        })
+        .collect::<Vec<Box<RecordBatch>>>();
+
+    if let [batch] = &batches[..] {
+        Box::new(ArrowAccessor::new(batch.clone(), schema))
+    } else {
+        panic!("Arrow should only contain a single record batch.")
```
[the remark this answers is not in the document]
sc1f: This is because Perspective exports Arrows as a single record batch, so I figured there was no need for joining multiple batches into a single array etc.

rust/perspective/arrow_accessor/tests/util.rs
sc1f: Will check with latest HEAD - I'll get a minimal pure Rust repro if that doesn't work out.

sc1f: Fixed in latest Git master - thank you for the heads up!

rust/perspective/arrow_accessor/src/accessor.rs
```diff
@@ -0,0 +1,227 @@
...
+    pub column_indices: HashMap<String, usize>,
+    pub data: Option<Array>,
+}
+
+impl ArrowAccessor {
+    // Convert every column in the Arrow record batch into a Javascript
+    // array that matches the output of Perspective's data serialization
+    // methods, i.e. datetime/dates should return Unix timestamps in
+    // milliseconds since epoch, i64/u64 should be coerced to float, etc.
+    pub fn new(batch: Box<RecordBatch>, batch_schema: SchemaRef) -> Self {
```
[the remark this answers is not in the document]
sc1f: Fixed - thanks!

nmichaud:
Independent of the benefits of using Rust within the perspective project (which I totally agree with), I'm curious what the Rust-based arrow data accessor gets you over using Arrow's javascript library? Is it mainly asset size?

sc1f:
> Independent of the benefits of using Rust within the perspective project (which I totally agree with), I'm curious what the Rust-based arrow data accessor gets you over using Arrow's javascript library? Is it mainly asset size?

At the moment, asset size with the rust-based accessor isn't that great - ~900ish kb of .wasm, which is far, far too large for the minimal amount of code. I'm hoping to cut this down in the near future, definitely before getting this anywhere close to production.

I'm not sure what size webpacking the Arrow JS library into Perspective would be - we removed all dependency on Arrow's JS library in #755 and #850, so I don't see us going back and re-adding the JS library. The arrow accessor would be simple to implement in C++ as well, and would just require linking against our custom Emscripten build of Arrow, but at this point the other benefits of using Rust seem to outweigh, and offer at least an opportunity to try out Rust and its WASM integration.

texodus:
Some additional context -
* It's certainly true that the Rust/WASM output as configured in this patch does not output svelte assets.  The time-to-idle for a page though is not so straightforward, since WASM assets with streaming instantiation can load much faster once downloaded than "equivalent" Javascript.  With some tinkering of the build config, I expect the time-to-idle to be roughly the same, with a larger asset that instantiates quicker.  
* Further iteration on this feature should trend towards a single, WASM, Arrow shared library in both the WebWorker _and_ Main thread, which can be downloaded once and copied, which will reduce the overall download size quite a lot.  The next step will be to integrate this work into the WebWorker C++/Emscripten library.
* The Rust and Emscripten Arrow builds we've tested are each several times faster than the TypeScript version, though this is not exactly disqualifying itself.

scripts/build_js.js
```diff
@@ -107,6 +107,28 @@ function compileCPP(packageName) {
     }
 }
 
+// eslint-disable-next-line no-unused-vars
+function compileRust() {
+    const base_dir = path.join(__dirname, "..", "rust", "perspective", "arrow-data-slice");
```
nevi-me, line 112: should "arrow-data-slice" not be "arrow_accessor"?

sc1f: good catch - that was outdated build code which I've replaced with better integration into our existing build system.

sc1f:
> we should probably add linting/autoformatting to the extent that it matches our other languages

@timkpaine - added linting and formatting using `rustfmt`, and I'll probably add [clippy](https://github.com/rust-lang/rust-clippy) as an optional check.

texodus:
I've reorganized this a bit, though it will not currently build until azure is updated to include `rustup` and `wasm-pack` tools it should build fine locally:
* Moved the project to `rust/perspective`, and following the pattern of `language/moduleName` renamed the module to `perspective`.
* This Rust module is now itself an NPM module `@finos/perspective-rust-internal`, which uses `wasm-pack` invoked from it's own NPM scripts' `build`. `rust` top level directory is now a yarn workspace, so any perspective module can link or import this module by name, with a small caveat that we still must perform webpack Voodoo to consecrate loading from another path or origin (the voodoo itself is confined to the `@finos/perspective` module [here](https://github.com/finos/perspective/pull/1261/commits/08e2023695689692bf9d1d70be5ddb4128fd5995#diff-f04c866d761a8acc0c73243c3efd143dd74fbe9faba99059782816578e92b735R29)).
* An example of depending-and-importing this Rust module has been added to `@finos/perspective-viewer-datagrid` in a [completely hackneyed manner](https://github.com/finos/perspective/pull/1261/commits/08e2023695689692bf9d1d70be5ddb4128fd5995#diff-3fb8f22dfe10171f2d143b58a6abfc3e5ecc99c59fdba6d10abffe1430faabeeR22).  This is just a demo of how to invoke this module, and shouldn't be part of the cleaned-up PR obviously ..
* `build_rust.js`, `test_rust.js` and `lint_rust.js` have been removed and replaced with lerna commands.  These may eventually be necessary, but this module as-written does not expose a _Rust API_, just a Javascript one, so for now I think we should use the Javascript build tools e.g. `lerna` & `yarn` rather than custom building.  I did also upgrade `lerna` in the commit to drastically improve error messages.

This module structure e.g. `@finos/perspective-rust-internal` is just temporary, as is the fact that it is currently marked `private`.  We need to decide whether to either publish this module, make the webpack builds of every module that imports them to _inline_ imports to itself, or combine this `wasm-pack` pattern to more complex module builds within `/packages`.

texodus:
I think we can finally close this relic!

The real #1261 was the friends we made along the way!

PR #1266, opened 2020-12-25
Fix: version conflict that makes `npm install` failed

damphat:
Fix: version conflict that makes `npm install` failed
Reason: perspective-webpack-plugin@0.5.6 require depends on perspective@0.4.2

Bug reproduction:
```
dependencies: {
    "@finos/perspective": "^0.5.6",
    "@finos/perspective-webpack-plugin": "^0.5.6",
}
```
```bash
> npm install
npm ERR! Could not resolve dependency:
npm ERR! peer @finos/perspective@"^0.4.2" from @finos/perspective-webpack-plugin@0.5.6
npm ERR! node_modules/@finos/perspective-webpack-plugin
npm ERR!   @finos/perspective-webpack-plugin@"^0.5.6" from the root project
```

### DIFF
--- a/packages/perspective-webpack-plugin/package.json
+++ b/packages/perspective-webpack-plugin/package.json
@@ -27,7 +27,7 @@
         "worker-loader": "^2.0.0"
     },
     "peerDependencies": {
-        "@finos/perspective": "^0.4.2",
+        "@finos/perspective": "^0.5.6",
         "webpack": "^4.35"
     }
 }


texodus (approved):

This fails to be auto-updated by lerna without `--force-publish`, which itself seems like overkill since this package rarely if ever changes.  I'm open to any suggestions on better ways to handle this, a few possibilities I can think of:
* Add a `prod` task for scripted production builds which calls this command, and perhaps eventually publish automatically from github actions or azure.  We've discussed this orthogonally to Webpack integration.
* Remove `peerDependencies` entirely.  Unless we significantly change the file/load architecture, this package should be compatible with all future perspective versions anyway.
* Merge this package with `@finos/perspective`.
* Remove Perspective's custom `.wasm` loading mechanism entirely, and rely on webpack 4's existing wasm support.  I think this is the most promising, but will require a bit of research to familiarize with the latest Emscripten/Webpack best practices.

PR #1271, opened 2021-01-04
Cleanup docs and examples to viewer.load(table)

sc1f:
This PR removes the last traces of `viewer.load(data)` and replaces them with `viewer.load(table)`, most notable in the installation and JS user guide documentation, but also in the script that powers the website's animated Perspective Viewer. 

Additionally, I removed an unused and undocumented API that allows the user to set a "data" attribute on `perspective-viewer` - since the viewer only loads tables, this API (which I don't see any usage of nor documentation about, and the last work done on it seems to be in a commit from [November 2018](https://github.com/finos/perspective/commit/2416733dab4b8cf2cbd7421b68205cce69208824)) can fully be removed.

### DIFF
--- a/docs/md/installation.md
+++ b/docs/md/installation.md
@@ -73,6 +73,7 @@ Once added to your page, you can access the Javascript API through the
 `perspective` symbol:
 
 ```javascript
+// Create Perspective's Web Worker, and then create a Table using your data.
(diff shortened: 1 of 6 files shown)


texodus (approved):

PR #1272, opened 2021-01-04
Validate view config in engine, fix computed column state bugs

sc1f:
This PR makes several bug-fixes on the View creation and computed column features of the Perspective engine.

Most notably, it introduces full validation of view configurations for invalid columns, i.e. a column that is not present in the table. While this is not an issue for most Perspective users, restoring a configuration that has invalid columns, for example, can cause issues that were previously difficult to debug. Now, all configuration options on the view are validated, and if an invalid column is specified, `abort()` is called in Javascript or a `PerspectiveCppException` will be thrown in Python with a specific error message detailing the invalid column name and where it was found.

Additionally, this PR fixes several edge cases with computed column state:

- Previously, one could create computed columns that had the same name as a "table column", i.e. column on the underlying table. Now, this will error in the computed columns UI on `perspective-viewer`, and the engine will `abort()` or throw `PerspectiveCppException` if attempted directly through the API.
- Previously, one could create computed columns that had the same name as another computed column, but retype it (i.e. replacing `a + b as c` with `uppercase(name) as c`). Now, this will error in the computed columns UI, and the engine will `abort()` or throw `PerspectiveCppException`.
- ~~Previously, if multiple views were created at the same time with shared computed columns, deletion of any one of the views would cause the shared column to no longer calculate on subsequent update cycles, as view deletion also removes the view's computed columns from listening to further updates. Now, multiple views can share the same computed columns and as long as one view is in scope containing the shared column, the shared column will continue to correctly update.~~
  - ~~This is/has not been an error on `perspective-viewer`, as the Viewer keeps track of one view at a time and does not simultaneously create multiple views, but it is a reproducible edge case that is now fixed.~~

Finally, more granular benchmarks for computed columns have been added to better gauge performance.

Edit: the additional complexity introduced with patching the shared-dependency issue has been removed from this PR, and the issue will be handled with additional feature/debt relief work that is part of building out the computed columns feature.

### DIFF
--- a/cpp/perspective/src/cpp/emscripten.cpp
+++ b/cpp/perspective/src/cpp/emscripten.cpp
@@ -1431,8 +1431,11 @@ namespace binding {
 
             t_dtype output_column_type = computation.m_return_type;
 
-            // Add the column to the schema.
-            schema->add_column(computed_column_name, output_column_type);
+            // Add the column to the schema if the column does not already
+            // exist in the schema.
+            if (schema->get_colidx_safe(computed_column_name) == -1) {
+                schema->add_column(computed_column_name, output_column_type);
+            }
 
             // Add the computed column to the config.
             auto tp = std::make_tuple(
--- a/cpp/perspective/src/cpp/emscripten.cpp
+++ b/cpp/perspective/src/cpp/emscripten.cpp
@@ -1413,37 +1413,53 @@ namespace binding {
              * columns on the View.
              */
             std::vector<t_dtype> input_types(input_columns.size());
+
+            // If the input columns are invalid, an error will be thrown here.
             for (auto i = 0; i < input_columns.size(); ++i) {
                 input_types[i] = schema->get_dtype(input_columns[i]);
             }
 
             t_computation computation = t_computed_column::get_computation(
                 computed_function_name, input_types);
-            
+        
+            // Throw an exception if the computation is invalid - the UI will
+            // prevent users from saving invalidly-typed computations.
             if (computation.m_name == INVALID_COMPUTED_FUNCTION) {
-                std::cerr 
-                    << "Could not build computed column definition for `" 
-                    << computed_column_name 
-                    << "`" 
-                    << std::endl;
-                continue;
+                // Get actual input types
+                auto valid_input_types = t_computed_column::get_computation_input_types(computed_function_name);
+                std::stringstream ss;
+                ss << "View creation failed: could not build computed column '"
+                << computed_column_name
+                << "' as the input column types are invalid."
+                << std::endl;
+                PSP_COMPLAIN_AND_ABORT(ss.str());
             }
 
             t_dtype output_column_type = computation.m_return_type;
 
-            // Add the column to the schema if the column does not already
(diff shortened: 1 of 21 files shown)


cpp/perspective/src/cpp/computed_column_map.cpp
```diff
@@ -27,7 +37,14 @@ t_computed_column_map::remove_computed_columns(
...
         if (m_computed_columns.count(name)) {
-            m_computed_columns.erase(name);
+            // Don't remove computed columns that are being referred to by
+            // multiple views.
+            if (m_computed_columns_count.at(name) > 1) {
+                m_computed_columns_count[name]--;
+            } else {
+                m_computed_columns.erase(name);
+                m_computed_columns_count.erase(name);
+            }
```
texodus, line 47: The complexity this introduces may not be worth it IMO - e.g., another case that will trivially break is duplicate column name dependencies in separate viewers.  We've discussed removing intermediate columns entirely from engine `process()` - I propose this may be an easier approach at this point than properly modeling all caveats of the intermediate column dependency system as currently written.

cpp/perspective/src/cpp/emscripten.cpp
```diff
@@ -1413,34 +1413,53 @@ namespace binding {
...
                 input_types[i] = schema->get_dtype(input_columns[i]);
             }
 
             t_computation computation = t_computed_column::get_computation(
                 computed_function_name, input_types);
-            
+        
+            // Throw an exception if the computation is invalid - the UI will
+            // prevent users from saving invalidly-typed computations.
             if (computation.m_name == INVALID_COMPUTED_FUNCTION) {
```
texodus, line 1427: `INVALID_COMPUTED_FUNCTION` magic values seems to indicate to me this error should occur in `get_computation()`, since we're already using `ABORT()` for error messages?

cpp/perspective/src/cpp/table.cpp
```diff
@@ -122,16 +137,49 @@ Table::get_computed_schema(
...
-                type = schema.get_dtype(input_column);
+                type = master_table_schema.get_dtype(input_column);
             }
             input_types.push_back(type);
         }
 
+        if (skip) {
+            // this column depends on a column that does not exist, so it
+            // does not need to be typechecked.
+            continue;
```
texodus, line 148: Why though?  With the removal protection also added in this PR - this case is an error, the column data may even be stale if for some reason it exists without its dependents?

texodus (commented):
Let's discuss `m_computed_columns_count` a bit before merging - it is likely worth it to fix this bug, but we need a concrete plan to address the _temporary_ state management problem wrt Computer Column accounting.

cpp/perspective/src/cpp/emscripten.cpp
sc1f: The function returns `INVALID_COMPUTED_FUNCTION` instead of abort so that the calling function can print a verbose error message, but as discussed offline we should get rid of the verbose error messages along with other refactors in a debt relief task.

texodus (approved):

PR #1273, opened 2021-01-04
free() pointer to arrow binary in Python

sc1f:
This PR fixes a small issue in Perspective's arrow loading code in Python - previously, the pointer to the malloc'd section for the arrow binary was never explicitly free'd (as it was in Javascript). This leads to the memory allocated for the binary leaking after the arrow is loaded into Perspective.

With testing using `psutil` and 25 updates using `superstore.arrow`, comparing memory usage with/without free() shows a significant impact:

_Without free()_:
- starting `rss`: 115539968 bytes
- ending `rss`: 302710784 bytes

_With free()_:
- starting `rss`: 113532928 bytes
- ending `rss`: 245538816 bytes

Memory usage grows much more slowly when `free()` is added, demonstrating the leak.

### DIFF
--- a/python/perspective/perspective/src/table.cpp
+++ b/python/perspective/perspective/src/table.cpp
@@ -34,6 +34,7 @@ std::shared_ptr<Table> make_table_py(t_val table, t_data_accessor accessor,
     std::shared_ptr<Table> tbl;
     std::shared_ptr<t_gnode> gnode;
     std::uint32_t offset;
+    void* ptr = nullptr;
 
     // If the Table has already been created, use it
     if (table_initialized) {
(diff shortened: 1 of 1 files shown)


python/perspective/perspective/src/table.cpp
```diff
@@ -193,6 +194,10 @@ std::shared_ptr<Table> make_table_py(t_val table, t_data_accessor accessor,
         _fill_data(data_table, accessor, input_schema, index, offset, limit, is_update);
     }
 
+    if (is_arrow) {
```
timkpaine, line 197: is there any case where `is_arrow && is_delete`?

timkpaine: e.g. should this match above `    if (is_arrow && !is_delete) {`

sc1f: There shouldn't be any - remove only takes a list of pkeys, but this whole conditional block is kind of messy. I'll take a look and fix.

texodus: This logic is incorrect, but so is the fix suggestion - there should be (or will need to be) one case, for `remove()` ... when we get around to fixing this API method generally ...

texodus (approved):

Issue #1242, opened 2020-10-30
Datetime and date `==` filters does not work as expected


sc1f:
### Steps to Reproduce:
 1. Create a table with date or datetime columns
 2. Filter on a datetime/date value using "=="

### Expected Result:

The correct datetime filter being applied to the view.

### Actual Result:

Various issues:

- The input format of the datetime is inconsistent; one cannot filter using the exact datetime string as displayed in the grid, and there are no error messages to help determine the correct format.
- When the datetime is input in a correct format, the filter doesn't work because the `==` match does not work. Using `>` or `<` comparators work okay, but also not as expected as ">" on an exact datetime value (with date, and time) doesn't actually exclude the value it matches.

In a `PerspectiveWidget` with `server=True`, the filtering works as intended:

![Screen Shot 2020-10-29 at 11 52 32 PM](https://user-images.githubusercontent.com/13220267/97658388-8ebf5680-1a42-11eb-8c34-89bff48ed7dd.png)

But not when the `PerspectiveWidget` is in default distributed mode:

![Screen Shot 2020-10-29 at 11 52 38 PM](https://user-images.githubusercontent.com/13220267/97658415-a1d22680-1a42-11eb-8af1-7075f9026951.png)

This leads me to believe this is an issue either in the `make_filter_term` code in `emscripten.cpp` and in the Javascript date parser itself; the former influences how a `Date` in JS is transformed into a Unix timestamp for internal conversion, and it is clear that the correct timestamp _does_ filter correctly in the core engine (as demonstrated by server mode `PerspectiveWidget`), and the latter influences what formats of dates are accepted.

Some suggestions:

- Fix the filter handling in Javascript - the issue should be in `make_filter_term` where it handles the conversion to Unix timestamp for `DTYPE_DATE` and `DTYPE_TIME`
- Make the filter UI more intuitive for datetimes/provide a single, correct format in an error message that the user can use to rewrite their filter config
- Make the JS date parser more lenient in terms of formatting
- More comprehensive viewer tests for filtering, as this seems to be a JS-specific issue

### Environment:

Latest master.

PR #1274, opened 2021-01-04
WIP: datetime filtering fixes

sc1f:
This draft PR will contain fixes for Perspective's datetime filters, specifically the issues mentioned in #1242 (and any others that may come up) - I'm using this draft PR as a place for notetaking/progress-checking and commentary.

Tasks:
- [ ] Fix how the datestring is being converted to timestamp
  - `==` filter compares the Unix timestamps - conversion not working properly?
- [ ] Improve UX for datetime filters
  - Make datestring format more lenient?
  - Add a hint to what formats are supported/make the default format more intuitive?
- [ ] Codify the errors that are happening, so we know exactly what to fix
- [ ] Enable programmatic filters using `Date()` values, not just strings
  - _Implemented already, although needs more tests_

### DIFF
--- a/cpp/perspective/src/cpp/emscripten.cpp
+++ b/cpp/perspective/src/cpp/emscripten.cpp
@@ -61,7 +61,7 @@ namespace binding {
         std::string datetime_string;
 
         if (filter_term.instanceof(t_val::global("Date"))) {
-            datetime_string = filter_term.call<t_val>("toLocaleString").as<std::string>();
+            datetime_string = filter_term.call<t_val>("toISOString").as<std::string>();
         } else {
             datetime_string = filter_term.as<std::string>();
         }
--- a/cpp/perspective/src/cpp/emscripten.cpp
+++ b/cpp/perspective/src/cpp/emscripten.cpp
@@ -57,7 +57,17 @@ namespace binding {
 
     t_val
     is_valid_datetime(t_val filter_term) {
-        return t_val(apachearrow::parseAsArrowTimestamp(filter_term.as<std::string>()) != -1);
+        // Convert `Date()` values to string before parsing
+        std::string datetime_string;
+
+        if (filter_term.instanceof(t_val::global("Date"))) {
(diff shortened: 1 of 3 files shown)


sc1f:
Issue with `==` filters on datetimes:

Because Perspective stores datetimes as Unix timestamps (ms since epoch), it performs the `==` filter by matching timestamp values. Perhaps filter values are being converted to timestamps in local time according to the browser, but Perspective (especially with arrow format data) is storing the timestamp in UTC.

cpp/perspective/src/cpp/emscripten.cpp
```diff
@@ -71,8 +81,27 @@ namespace binding {
...
+        // function in `jsdate_to_t_date` gets the ymd in local time as well
+        // straight from the date object. HOWEVER this issue was present in
+        // 0.4.8 - which leads me to believe that this has been an issue all
+        // along.
+        //
+        // we should maybe have a separate node test process that runs with
+        // TZ="US-Eastern" or something like that to assert the timezone
+        // differences.
+        // ...I think...
+        tm utc_tm = *std::gmtime(&tt);
```
sc1f, line 103: @texodus this change fixed date filtering from:

Screen Shot 2021-01-05 at 9 49 38 PM

to:

Screen Shot 2021-01-05 at 9 47 23 PM

I think this issue was around for a while (earliest version I tested was 0.4.8, far before the date parser rewrite). What I *think* is happening - `Date.getTime()` always uses UTC, so the timestamps stored by Perspective are in UTC, and then re-localized to local time when they are passed into `new Date()` when output. However, this function (and its previous implementation, `jsdate_to_t_date`) pulled the ymd values as local time, which then causes a re-localization on top of an already localized datetime - thus the issue happening in the screenshot above. I would like to verify this with a proper test suite akin to what we have in the Python tests - that will be coming as part of this patch.

What I am unsure about - if datetime values are stored as UTC timestamps and correctly localized, then why does `==` on a datetime still show an offset:

Screen Shot 2021-01-05 at 10 00 10 PM

This behavior has also been around since 0.4.8.

PR #1275, opened 2021-01-06
New Documentation & Project Site

texodus:


sc1f (approved):
Looks good! Pushed up a small change to remove TBB from system dependencies as we don't build it ourselves - new docs format and the ability to see both Python and JS code snippets on the same page is super useful.

PR #1276, opened 2021-01-06
Thread-safe Readers

texodus:
This PR extends the `perspective-python` threading model to support fully thread-safe read operations, in exchange for a
"global" writer lock.  For datasets with many readers and infrequent writes, this can offer significantly better concurrency utilization.

- [x] Implement
- [ ] Measure performance under contention
- [ ] Decide whether to continue to support lock-less `async` mode


texodus:
Closing in favor of #2261

PR #1277, opened 2021-01-06
Order Book Example

texodus:
Adds `order-book` Python example, a non-trivial market simulation using purely Perspective Queries to construct and clear a virtual Order Book constructed live from "Computer" and "Player" orders, with a web frontend Market Tracker, Portfolio Analysis and Order Form built on `<perspective-workspace>`.


timkpaine:
You can't trade against it and its only running 9:30-4, but you can also stream the full IEX order book https://github.com/iexcloud/iexjs/blob/5e3f4ff26c16c3405e1ebd7cc2534edd0afd12de/src/js/streaming/sse.js#L100

bogorman:
Will this example ever be added to the main repo? After watching the presentation it would be great to have a working version.

texodus:
@timkpaine This is going to be the topic of the next Perspective demo - I think this one is too simplistic for IEX anyway, the UI is really designed for a toy/game than a true market picture.

@bogorman Sorry, there's a lot to work on in Perspective and we sometimes lost track.  There are no real blockers, I'm just not completely happy with how this example turned out and I wanted to hack on it for a few days more with an eye towards:

1) Simplifying the code, add tests, docs and such  This example is big and has alot of plumbing, and I'd like it to be informative of "good" perspective usage, so our "official" examples aren't a minefield of good vs half-baked examples.
2) Move some of the abstractions learned through building this example into the core library - specifically, One-Time Queries are super useful and we don't have a good API for them right now in the core library.

I'll put this on the bench for the `1.1.0` release demo for now, and update this thread if for some reason we decide to defer it.

bogorman:
Thanks for the reply. I managed to get the UI up and running but I am having an issue with the server. 

2021-11-30 16:27:12,679 - [Thread-1    ] 31786 ERROR: Exception in callback functools.partial(<function Market._with_thread.<locals>._with_thread.<locals>.<lambda> at 0x7fd402919670>)
Traceback (most recent call last):
  File "/Users/b/opt/anaconda3/lib/python3.8/site-packages/tornado/ioloop.py", line 741, in _run_callback
    ret = callback()
  File "src/server/server.py", line 294, in <lambda>
    self._manager._loop_callback(lambda: q.put(f(self, *args, **kwargs)))
  File "src/server/server.py", line 320, in get_spot
    history = self._get_price_history(sym, end_row=1)
  File "src/server/server.py", line 180, in _get_price_history
    with self._market_table.view(
AttributeError: __enter__
^CTraceback (most recent call last):
  File "src/server/server.py", line 507, in <module>
    loop.start()

I get this immediately after starting it. If you have seen it before then you might let me know what is going on and how to fix otherwise I'll dive into it and see if I can get it working. I'm using a more recent version of perspective and python 3.8 so I'm guessing something broke along the way. I can't see anything obvious though. Thanks

bogorman:
> Thanks for the reply. I managed to get the UI up and running but I am having an issue with the server.
> 
> 2021-11-30 16:27:12,679 - [Thread-1 ] 31786 ERROR: Exception in callback functools.partial(<function Market._with_thread.._with_thread.. at 0x7fd402919670>) Traceback (most recent call last): File "/Users/b/opt/anaconda3/lib/python3.8/site-packages/tornado/ioloop.py", line 741, in _run_callback ret = callback() File "src/server/server.py", line 294, in self._manager._loop_callback(lambda: q.put(f(self, *args, **kwargs))) File "src/server/server.py", line 320, in get_spot history = self._get_price_history(sym, end_row=1) File "src/server/server.py", line 180, in _get_price_history with self._market_table.view( AttributeError: **enter** ^CTraceback (most recent call last): File "src/server/server.py", line 507, in loop.start()
> 
> I get this immediately after starting it. If you have seen it before then you might let me know what is going on and how to fix otherwise I'll dive into it and see if I can get it working. I'm using a more recent version of perspective and python 3.8 so I'm guessing something broke along the way. I can't see anything obvious though. Thanks


I rewrote the way it uses with when its creating the views and that seems to fix that issue. Not sure why. Next issue is due to the fact that I am using a 1.0.2 of perspective and I think this was done with an older one. When its creating the views its complaining about computed_columns which I understand is now an expression. I rewrote the to use the new expressions and I think its correct. 

            # computed_columns=[
            #     {
            #         "column": "outstanding",
            #         "computed_function_name": "-",
            #         "inputs": ["qty", "filled"],
            #     }
            # ],
            expressions=[
                "//outstanding\n\"qty\"-\"filled\""
            ],            
            columns=["outstanding"],

I am getting this error. Does the expression column get added to the schema and am doing the correct thing to name the column or is there some other format I can use. 

perspective.table.libbinding.PerspectiveCppError: Could not find column index for `outstanding` as it does not exist in the schema.

Thanks.

texodus:
Deprecated by #2128

PR #1280, opened 2021-01-07
Parse date locale strings, autocomplete filtering for date/datetime columns

sc1f:
This PR takes over from #1274, and tries to fix some of the issues in #1242. Datetime filters are currently heavily format dependent - though the formats are the same as the date string formats supported by Perspective, this is not well-documented, and more specifically locale strings are not accepted in either the dataset nor the filters, but dates/datetimes are *displayed* as locale strings. This leads to a disconnect in the user experience, as the most obvious date format—the one displayed on the grid—is considered "invalid".

This PR adds in parsing support for US and UK locale strings, as well as enables autocomplete on date/datetime columns, which will display a list of the unique dates/datetimes as locale strings in the filter column. This should improve the user experience of datetime filters.

One unique issue right now pertains to `strptime` in the C STL - when given the US locale string format ("%m/%d/%Y, %I:%M:%S %p"), it parses "12:00:00 AM" as "12:00:00 PM", which is incorrect. I've added in some print statements in the C++ for now to try to figure out where this issue is happening—this is coming from strptime, and not arrow's date parser (which internally uses strptime). This issue does not affect Perspective at present, since the current parser simply does not parse 12-hour US locale strings.

UK locale strings seem to work fine in the same situation though :)

### DIFF
--- a/cpp/perspective/src/cpp/arrow_csv.cpp
+++ b/cpp/perspective/src/cpp/arrow_csv.cpp
@@ -144,6 +144,8 @@ namespace apachearrow {
     std::vector<std::shared_ptr<arrow::TimestampParser>> DATE_PARSERS{
         std::make_shared<CustomISO8601Parser>(),
         arrow::TimestampParser::MakeStrptime("%Y-%m-%d\\D%H:%M:%S.%f"),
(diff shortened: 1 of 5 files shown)


sc1f:
Closing in favor of #1282

PR #1281, opened 2021-01-07
Fix Date/Time formatting in `perspective-viewer` CSV output

### LINKED ISSUE #1265: Dataframe copy / dump to csv from perspective datetime formatting
When I copy a time series dataframe to clipboard or dump the df to csv through perspective, all of the datetimes are in epoch time. Is it possible to copy/dump into a friendlier format? I'm not sure what is best, maybe 'dmmmyy h:mm:ss'?

texodus:
Fixes #1265

### DIFF
--- a/examples/blocks/src/superstore/index.html
+++ b/examples/blocks/src/superstore/index.html
@@ -35,7 +35,7 @@
       
         <script>
 
-            const URL = "/node_modules/superstore-arrow/superstore.arrow";
+            const DATA_URL = "/node_modules/superstore-arrow/superstore.arrow";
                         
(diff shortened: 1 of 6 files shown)


sc1f:
I added a fix - previously, to_csv would only display the date because the `Intl.DateTimeFormatter` was not configured properly. Now, it will display the date and time for datetime columns, and display the date for date columns.

Additionally, I also fixed CSV export from the viewer on a Python server; because the `date_format` API is a little different for Python (takes a fmtstring, and not a locale), I defaulted it to the standard "YYYY-MM-DD HH:MM:SS" format.

PR #1286, opened 2021-01-12
Fix for sorting via `avg` aggregated column.

### LINKED ISSUE #1285: Sorting a column aggregated by `avg` does not sort
## Bug Report

Given at least one `row-pivot` and a sort column with the `avg` aggregate, the sort is not applied.

### Steps to Reproduce:
 
[Draft PR](https://github.com/finos/perspective/pull/1286) with a repro and examples that non-`avg` aggregates sort correctly. [commit](https://github.com/finos/perspective/commit/52f56e0c3b690782608043eb968a9de7dc4bfffc)

### Environment:
JavaScript and Python but only JS tests are in the repro PR.

texodus:
- [X] Isolate with test case
- [x] Fix the bug

lol.

Fixes #1285 .

### DIFF
--- a/packages/perspective/test/js/sort.js
+++ b/packages/perspective/test/js/sort.js
@@ -16,6 +16,92 @@ const data = {
 
 module.exports = perspective => {
     describe("Sorts", function() {
+        describe("With aggregates", function() {
+            describe("aggregates, in a sorted column with nulls", async function() {
+                const data2 = {
+                    w: [3.5, 4.5, null, null, null, null, 1.5, 2.5],
+                    x: [1, 2, 3, 4, 4, 3, 2, 1],
+                    y: ["a", "b", "c", "d", "e", "f", "g", "h"]
+                };
+
+                it("sum", async function() {
+                    var table = perspective.table(data2);
+                    var view = table.view({
+                        columns: ["x", "w"],
+                        row_pivots: ["y"],
+                        aggregates: {
+                            w: "sum",
+                            x: "unique"
+                        },
+                        sort: [["w", "asc"]]
+                    });
+
+                    const json = await view.to_columns();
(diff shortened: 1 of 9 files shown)


packages/perspective/test/js/sort.js
```diff
@@ -14,8 +14,338 @@ const data = {
...
+                        });
+
+                        const json = await view.to_columns();
+                        expect(json).toEqual({
+                            __ROW_PATH__: [[], ["a"], ["b"]],
+                            w: [3, 2, 1],
+                            x: [20, 10, 10]
+                        });
+
+                        // current wrong result:
```
sc1f, line 335: @texodus seems to me the avg here is wrong - it should be `avg(3.5, 4.5, null, null)` and `avg(null, null, 1.5, 2.5)`, so 2 and 1 - the total is right but the individual components are wrong here.

sc1f:
Added some changes that fix #1285 by re-enabling `nan` comparisons while sorting, which were disabled in #689 for performance fixes. After re-enabling `nan` comparisons and running benchmarks on various dataset sizes and sorted contexts, this change does not have a performance impact so it should be okay to re-enable in the interest of correctness.

I've also changed the default error logging from WASM to throw a detailed error message instead of `Abort()`, which will help in debugging (a small improvement from #1289 that I wanted to carry over). I did find another bug with the `avg` aggregate when multiple sorts are applied with nulls, and I've added a (currently failing) test to try to nail down that issue as well.

packages/perspective/test/js/sort.js
sc1f: nvm - it is calculating the avg. of the non-null rows, which is correct.

texodus:
Thanks for the fix!

PR #1289, opened 2021-01-12
Async Table and View Constructor

sc1f:
This PR makes a **breaking** public API change:

The `table()` and `view()` constructors in Javascript and the Python Tornado client are now `async/await`. In Javascript, `perspective.table()` and `table.view()` now return Promises that will resolve to a Table or View, or be rejected with an error. This allows for much clearer error handling during table and view construction, especially in the [client/server](https://perspective.finos.org/docs/md/server.html#clientserver-replicated) model where it is important for the client to know why a table/view could not be constructed and to act on that error information. Additionally, this enables better error reporting from `perspective-viewer` itself, as the viewer can `load()`  a `Promise() => Table` and display the error if the promise is rejected.

Additionally, this PR also improves the error reporting of the WASM binding. Previously, an error message was logged to `console.warning` and an `Error("Abort()"` was thrown:

Screen Shot 2021-01-19 at 2 25 28 PM

This PR simplifies error reporting by throwing an `Error` with the custom error message that was previously just logged, which should simplify the way errors are thrown and make the error condition more clear:

Screen Shot 2021-01-19 at 2 25 46 PM

## Changelog
- async `Table` and `View` constructors in JS, and Python Tornado client
- Refactored `perspective-viewer`, `perspective-viewer-jupyterlab` to use async/await constructors
- Refactored docs, all examples, all tests
- Added error handling & load(Promise) tests
Running benchmarks on this PR, 0.6.0, and 0.5.6 shows that performance is not impacted by this API change:

Screen Shot 2021-01-12 at 5 53 58 PM


texodus (approved):

I fixed up the regressions since `0.6.0`, and reverted the test changes (these are mostly fixed on master anyway and I'm not sure this PR's test fixes do ..)

PR #1215, opened 2020-10-04
Upgrade `pyarrow` to 1.0.1

texodus:
Following up from #1207, this PR configures Python (>=3) to build against `pyarrow>=1.0.1`.

Issue #1157, opened 2020-08-19
Load arrow tables into Perspective without converting to a binary buffer


sc1f:
### Description of Problem:

Perspective expects arrow data to be loaded as an ArrayBuffer in Javascript, and a binary string in Python. This requires, in PyArrow at least, a few lines to convert an Arrow `Table` into binary:

```python
stream = pa.BufferOutputStream()
writer = pa.RecordBatchStreamWriter(stream, arrow_table.schema)
writer.write_table(arrow_table)
writer.close()
perspective_table = perspective.Table(stream.getvalue().to_pybytes())
```

When loading arrow Tables, I expect Perspective to be compatible with Arrow Tables without having to do any conversion. The requirement of `bytes` for an Arrow binary is outlined slightly in the Python user guide, but the conversion process from an Arrow Table -> bytes is unclear.

### Potential Solutions:

Write in an Arrow Table to binary conversion layer solely in the binding layer (using PyArrow or Arrow typescript), which would be simple but incomplete (would require reimplementation in future binding languages) and less performant. If there is a way to transfer a pointer to an arrow table from the binding layer into C++, then we should be able to write the conversion entirely in C++. We already malloc and memcpy the arrow binary from JS/Python into C++, so there might be something already there worth looking into.

timkpaine:
There is a way to transfer the pointer, I discussed this with @wesm a while back and it was eventually implemented to optimize transferring arrow tables between R and python. Let me look up the details of that thread

timkpaine:
Looking at the code it looks simple to add, I will look into it this week and do a PR if I get something useable.

Tim Paine
tim.paine.nyc
908-721-1185

timkpaine:
`unwrap_array` was the original way, but we weren't interested at the time because it required linking against `libarrow_python`. But since we do that anyway it should be ok
https://arrow.apache.org/docs/python/extending.html#_CPPv412unwrap_arrayP8PyObjectPNSt10shared_ptrI5ArrayEE

sc1f:
@timkpaine is there a way to do the same in the WASM? I'm personally ok with only being able to load pyarrow Tables in our python binding but for the sake of completeness it would be helpful in WASM

wesm:
You should be able to use the C interface now. cc @pitrou @nealrichardson

timkpaine:
Non-c api but still doesn't require linking or knowing about pyarrow from the c++ side which was our goal:
[...]

PR #1290, opened 2021-01-12
Statically link `perspective-python` and `libarrow`

sc1f:
This PR builds Apache Arrow as a static library and reconfigures `perspective-python` to no longer link against a version of PyArrow present on the system/installed as part of `perspective-python` dependencies.

Instead, Perspective's C++ library will link against the pre-compiled, minimal arrow static library, and the user is free to use any version of PyArrow they choose to install as long as the Arrow binary format is compatible with the binaries output by Perspective's version of Arrow, which is *v1.0.1* at this time.

At present, Perspective does not `import pyarrow` or use any of Arrow's Python-specific C++ code. If we were to use PyArrow in the Python runtime, however (like in the test suite), Perspective's C++ version of Arrow will continue to work as expected—`perspective-python` will use its own version of Arrow, and `pyarrow` will use the `libarrow.so` that shipped with the PyArrow install. This means further integration with PyArrow is not blocked/degraded, and if anything should be accelerated as we now have a consistent Arrow version across all Perspective runtimes.

In effect, we have taken a "pinned" version of PyArrow that used to depend on external assets we could not fully/consistently control, and turned it into a pinned version of PyArrow that we have full management and control over.

### Benefits

With this change, the complicated and admittedly brittle code we have in place to deal with detecting PyArrow's install location, managing PyArrow versions, copying Arrow DLLs (on Windows), etc. can be removed, as Perspective's build no longer depends on the existence and correct versioning of an asset we have little control over.

Additionally, users can install the newest version of PyArrow in order to gain access to features/fixes without worrying about Perspective compatibility beyond the binary format, which is expected to remain stable and compatible between versions. Finally, the developer experience should be improved for both source builds and configuring the `conda-forge` build, which can unpin its PyArrow dependency.

### Testing

Perspective's WASM build has built and used the minimal version of Arrow since late 2019, and `perspective-python`'s test suite contains a litany of tests that utilize PyArrow to generate/ingest arrow binaries to and from Perspective. This PR updates the Manylinux Dockerfiles and Python test suite to use PyArrow `v2.0.0` (except for Python 2, which continues to use `v0.16.0`). This in effect gives us full coverage over forwards and backwards compatibility between our static Arrow pinned to `v.1.0.1` and the versions of PyArrow used by the tests.

### TODO:

- [x] Remove PyArrow maintenance code/requirements from `setup.py`, `pyproject.toml` etc.
- [ ] Test that Perspective works with the latest version of PyArrow, and update the test suite to use latest Arrow
- [ ] Remove PyArrow custom installs from all Dockerfiles, and regenerate Docker images
- [ ] Clean up CMake code around PyArrow


cmake/arrow/CMakeLists.txt
```diff
@@ -101,6 +101,22 @@ set(ARROW_SRCS
...
+        ${CMAKE_BINARY_DIR}/arrow-src/cpp/src/arrow/tensor/coo_converter.cc
+        ${CMAKE_BINARY_DIR}/arrow-src/cpp/src/arrow/tensor/csf_converter.cc
+        ${CMAKE_BINARY_DIR}/arrow-src/cpp/src/arrow/tensor/csx_converter.cc
+        ${CMAKE_BINARY_DIR}/arrow-src/cpp/src/arrow/util/formatting.cc
+        ${CMAKE_BINARY_DIR}/arrow-src/cpp/src/arrow/util/time.cc
+        ${CMAKE_BINARY_DIR}/arrow-src/cpp/src/arrow/vendored/double-conversion/bignum-dtoa.cc
+        ${CMAKE_BINARY_DIR}/arrow-src/cpp/src/arrow/vendored/double-conversion/fast-dtoa.cc
+        ${CMAKE_BINARY_DIR}/arrow-src/cpp/src/arrow/vendored/double-conversion/fixed-dtoa.cc
+    )
+endif()
```
texodus, line 118: Would be nice if we could move this logic to at least where `ARROW_SRCS` is defined.

sc1f:
@texodus this PR now has the changes from #1215 in it, so once this is ready to merge we can close #1215.

cmake/arrow/CMakeLists.txt
sc1f: I don't think we can conditionally inject in more items - we could use `list(APPEND` but I think this is a clear way to express the additional source files we need for Python.

sc1f:
@texodus tests now run against PyArrow 2.0.0, checked locally and on Azure Mac/Windows jobs. Just need to regen the docker images and should be all clear.

@timkpaine would you be able to pull down this branch and take a look if you have time? I don't have a Windows machine to verify, but the build should work as long as you have flatbuffers installed before build.

timkpaine:
Need to check that an arrow past in from python will be binary compatible with the version we built against, so expose the arrow api version into python and check against whatever attribute is there on the python side, otherwise it will likely just seg fault when the arrow versions are incompatible

texodus:
@timkpaine Thanks for the feedback!  I don't think this is necessarily an issue, since 
* None of the logic of how we use `pyarrow` internally has changed at all, and this PR does not modify any `.py` files.
* The `arrow::Table` load logic used in C++ does version validation (though these don't yet work well in js, see #1289).
* The format is backwards-compatible (?) from `0.16` 
* Adding `pyarrow` to check the version from the Python-side, re-introduces the exact dependency on `pyarrow` that this PR seeks to remove.

I dug up a _very_ old `superstore.arrow` from Perspective `0.1.0` on unpkg [here](https://unpkg.com/browse/@jpmorganchase/perspective-examples@0.1.0/build/superstore.arrow), and it throws an exception with a not super helpful error message but does not segfault:

> Failed to open RecordBatchStreamReader: Invalid: Expected to read 1329865020 metadata bytes, but only read 36639

I concur this needs work, lets aim to improve the quality of our error reporting in a PR which targets this specifically.

cmake/arrow/CMakeLists.txt
texodus: Yes sorry ignore this comment, am dumb

cpp/perspective/CMakeLists.txt
```diff
@@ -435,21 +426,29 @@ psp_build_dep("date" "${PSP_CMAKE_MODULE_PATH}/date.txt.in")
...
-	else()
-		psp_build_message("${Cyan}Found Flatbuffers in ${FLATBUFFERS_INCLUDE_DIR}${ColorReset}")
-		include_directories( ${FLATBUFFERS_INCLUDE_DIR} )
-	endif()
+# FIXME: this is a hack to get Flatbuffers working on Azure Win64 by making the
+# headers accessible. The actual flatc executable is installed using
+# Chocolatey for our Azure Windows job.
+if (WIN32)
+	psp_build_dep("flatbuffers" "${PSP_CMAKE_MODULE_PATH}/flatbuffers.txt.in")
+endif()
```
texodus, line 441: Why do the `flatbuffers` headers need to be available only on Windows?

docs/md/development.md
```diff
@@ -258,13 +250,7 @@ The most common culprits are:
...
 - CMake version too old
 - Boost headers are missing or too old
-- PyArrow not installed prior to installing perspective
-
-Additionally, due to PEP-518 and build isolation, its possible that the version
-of PyArrow that pip uses to build perspective-python is different from the one
-you have installed. To disable this, pass the `--no-build-isolation` flag to
-pip.
-
+- Flatbuffers not installed prior to installing Perspective
```
texodus, line 253: Need a newline, markdown is picky sometimes

texodus:

timkpaine:
> I dug up a _very_ old `superstore.arrow` from Perspective `0.1.0` on unpkg [here](https://unpkg.com/browse/@jpmorganchase/perspective-examples@0.1.0/build/superstore.arrow), and it throws an exception with a not super helpful error message but does not segfault:
> 
> > Failed to open RecordBatchStreamReader: Invalid: Expected to read 1329865020 metadata bytes, but only read 36639
> 
> I concur this needs work, lets aim to improve the quality of our error reporting in a PR which targets this specifically.

So long as it fails in python and not in c++ (e.g. no segfault, user can handle in python) its "fine".

> Adding pyarrow to check the version from the Python-side, re-introduces the exact dependency on pyarrow that this PR seeks to remove.

Note we won't need to re-add `libarrow_python`, we just need to expose the version of arrow that perspective statically linked against into python and we can do the check purely in python at load time. If you're passing in a pyarrow.Table (e.g. `pyarrow in sys.modules`) we can do a version check. [Right now we treat any bytes as arrow](https://github.com/finos/perspective/blob/master/python/perspective/perspective/table/table.py#L52) which should probably have these types of checks anyway. I assume the version is encoded in the table binary in some form as well, so we can even use that

sc1f:
> @sc1f this also seems to break Mac or linux when lib binding looks for libpsp

I think I might have removed the `$ORIGIN` from the rpath setting when I was cutting out the old pyarrow - will take a look and fix.

timkpaine:
I've fixed on my jlab 3 PR

texodus:
@timkpaine `pyarrow` is not imported by `perspective-python`, and `perspective.Table()` doesn't support `pyarrow.Table` as a constructor argument (unless there is some Python coercion coincidence I am unaware of ..).  As I stated above, `arrow::Table` does validation on blobs, as it should since we don't even know what ranges to validate without reading the Arrow docs like savages.

Just to re-emphasize - this PR changes _no python_, it only removes the dependency shared `libarrow` which came packaged with `pyarrow`  (e.g. the non-Python part).  Or as stated in the 3rd paragraph of the PR:

> At present, Perspective does not import pyarrow or use any of Arrow's Python-specific C++ code. If we were to use PyArrow in the Python runtime, however (like in the test suite), Perspective's C++ version of Arrow will continue to work as expected—perspective-python will use its own version of Arrow, and pyarrow will use the libarrow.so that shipped with the PyArrow install. This means further integration with PyArrow is not blocked/degraded, and if anything should be accelerated as we now have a consistent Arrow version across all Perspective runtimes.

timkpaine:
So long as the arrow validation doesn't trigger un catchable errors in python, it should be a straight improvement. 

If you're shoving an arrow into perspective in python, you're probably using pyarrow. Right now we go pyarrow table -> byte string copy -> perspective (2 copies total). We can no longer do pyarrow table -> no copy arrow table -> perspective (1 copy total), but I think it's ok given the better install experience (and we never implemented it anyway).

So the only negative is that if you're shoving an arrow from your version of pyarrow into our version of arrow and it's incompatible, you get an obscure error. We should make it a PerspectiveCPPException or whatever we called it.

texodus:
#1157

timkpaine:
you mean this comment? https://github.com/finos/perspective/issues/1157#issuecomment-676600779

sc1f:
> So the only negative is that if you're shoving an arrow from your version of pyarrow into our version of arrow and it's incompatible, you get an obscure error. We should make it a PerspectiveCPPException or whatever we called it.

IIRC it should be a `PerspectiveCppException` already, although I agree the "Expected to read..." error is a little obscure. We have more detailed error messages when we check the status of our arrow reads during the load, so it reports more verbose messages to tell you which columns failed, whether it could read the recordbatch, etc. It seems like this "Expected to read..." error is coming from the Arrow C++ lib itself, so it would be harder to transform that message.

I saw the fix for rpath on #1294 - I definitely was heavy-handed in removing PyArrow-related code from our CMakeLists, and cut out the "set rpath to $ORIGIN" block as well. Thanks for the fix!

timkpaine:
Yep because we expose `PerspectiveCppException` into python you can `try: except` it, whereas an error out of Arrow C++ is not going to be exceptable. Worse case we just `try:catch` in C++ and in the catch handler bubble it up into python

sc1f:
My recollection of how we set up PyBind is that all exceptions from C++ get raised as `PerspectiveCppException` in Python - let me get the old version of superstore.arrow from @texodus. I tried the oldest version I could find from [May 2019](https://unpkg.com/@jpmorganchase/perspective-examples/build/superstore.arrow) and it worked fine.

Issue #857, opened 2019-12-20
Worker file's runtime location


stevenvachon:
## Support Question
I'm using the webpack plugin, which correctly copies the worker file to my bundle's output directory. My issue is that the worker file can't be found by my server (`localhost:3000`), where the console shows:
```
Uncaught SyntaxError: Unexpected token '<'
http://localhost:3000/perspective.wasm.worker.js
```

The file is actually at `http://localhost:3000/public/plugins/perspective-panel/perspective.wasm.worker.js`.

texodus:
Try setting `publicPath` in the `output` section of your webpack config, ala the [webpack example](https://github.com/finos/perspective/blob/master/examples/webpack/webpack.config.js#L19), if you are not doing so already.  If this doesn't work, I may need some help getting a repro together - a minimal webpack config which replicates the problem would make this an easy fix.

The client code which downloads/instantiates the Web Worker looks for `perspective.wasm.worker.js` and `psp.async.wasm` files in the same [path](https://github.com/finos/perspective/blob/master/packages/perspective-webpack-plugin/src/js/psp-worker-loader.js#L80) where it was itself downloaded (`perspective.js` or `perspective-viewer.js`, or the asset into which it was compiled in your app).  This logic seems to be broken in your app for some reason.

I think in this case both 1) a better error message and 2) a user-configurable download location are warranted - I will add these imminently.   We could also use much better documentation on the plugin and its myriad of options - it actually supports inlining the entire worker and wasm payload into a single asset at the expense of load performance.

stevenvachon:
> it actually supports inlining the entire worker and wasm payload into a single asset at the expense of load performance.

Initial load only?

texodus:
Yes, initial load only, as in time-to-library-ready, not dataset load or processing time.  In this mode, you configure webpack and manually load the `perspective.wasm.worker.js` and `psp.async.wasm` directly, then use the undocumented `perspective.override()` method to override the default download logic.  This increases the cumulative payload size by ~500kb, takes 2-3x as long to load load the library cold, in addition to just wrecking havoc on debuggers and whatnot due to the asset size itself.

[...]

PR #1292, opened 2021-01-15
Add `@finos/perspective-cpp` module

texodus:
This is a big PR so I've added a bit of self-review below to point out the most controversial parts.  The intent was to replace custom build scripts in `/scripts/build_js.js` especially with an NPM module which exports Emscripten-built assets only, modelled on Rust's `wasm-pack`'s architecture.  The general result is a smaller, easier to reason about build, and a decreased reluctance on `webpack` arcana, such that `@finos/perspective-webpack-plugin` can be effectively recreated with just contrib modules:

```javascript
module.exports = {
    rules: [
        {
            test: /perspective\.worker\.js$/,
            loader: "worker-loader"
        },
        {
            test: /perspective\.cpp\.wasm$/,
            type: "javascript/auto",
            loader: "file-loader"
        }
    ]
};
```

Making it simple also makes it easier to customize and harder to introduce bugs like #857, and this new `webpack` build should be easy to set up to run in various cross-origin, chunking, or inlining scenarios as needed in your app's `webpack.config.js`.

The new module added, `@finos/perspective-cpp`, is private but otherwise normal and its `build` task just runs `emscmake` and `emmake`, producing the `.wasm` file in its `dist` folder.  As it is now part of the `yarn workspace` dependency graph, it is set as a `devDependencies` of `@finos/perspective`;  when `@finos/perspective` is later compiled, the resulting `dist` is packaged with `npm pack` and installed within `@finos/perspective` package's own `dist`, making it an "internal" module.

The build script itself is now only a few lines and self contained within `@finos/perspective-cpp` rather than split across directories and scripts.  It can also now be built independently, or to use an e.g. `DEBUG` `@finos/perspective-cpp` with a production `@finos/perspective`, which is useful since `DEBUG` builds can be very slow.  It's also been extended a bit to supprt two Emscripten builds (see self review).

As a side-effect, I made several dependency upgrades to fix various issues:
* `lerna` upgrades fixes error messages being duplicated in the console - as Emscripten now builds within `lerna`, this is a requirement for reading C++ errors.
* `webpack` and co. upgrades fix a number of WebAssembly linking problems and longstanding other issues such as [this](https://github.com/webpack/webpack/issues/11127).  
* `react` and other libraries only used within `/examples` - I went ahead and unified these versions across all packages as well, and removed duplicate dependencies to make this easier to track.


cpp/perspective/CMakeLists.txt
```diff
@@ -586,16 +585,24 @@ message("${BUILD_MESSAGE}\n")
...
+	set_target_properties(perspective_esm PROPERTIES RUNTIME_OUTPUT_DIRECTORY "./esm/")
+	set_target_properties(perspective_esm PROPERTIES OUTPUT_NAME "perspective.cpp")
+
+	# "cjs/perspective.cpp.js" from CMAKE_EXECUTABLE_SYNTAX
+	add_executable(perspective_cjs src/cpp/emscripten.cpp)
+	target_link_libraries(perspective_cjs psp "-s ENVIRONMENT=node")
+	target_compile_definitions(perspective_cjs PRIVATE PSP_ENABLE_WASM=1)
+	set_target_properties(perspective_cjs PROPERTIES COMPILE_FLAGS "-s ENVIRONMENT=node")
+	set_target_properties(perspective_cjs PROPERTIES RUNTIME_OUTPUT_DIRECTORY "./cjs/")
+	set_target_properties(perspective_cjs PROPERTIES OUTPUT_NAME "perspective.cpp")
```
texodus, line 605: Up until now, we've been building Emscripten's mixed JavaScript runtime, which supprots a variety of [states](https://github.com/emscripten-core/emscripten/blob/master/src/settings.js#L546) at the same time.  This is handy for a simple build, but leaves some uncalled references to technically invalid JS in the build which then must be configured in webpack.  Since we want to minimize the configuration required in webpack to the bare minimum, we build 2 Emscripten builds, with `"worker"` and `"node"` modes, as well as stripping ES modules from the node version since Node doesn't support them.

cpp/perspective/build.js
```diff
@@ -0,0 +1,19 @@
...
+const cwd = path.join(process.cwd(), "dist", env);
+
+try {
+    execSync(`mkdirp ${cwd}`, {stdio});
+    process.env.CLICOLOR_FORCE = 1;
+    execSync(`emcmake cmake ${__dirname} -DCMAKE_BUILD_TYPE=${env}`, {cwd, stdio});
+    execSync(`emmake make -j${os.cpus().length}`, {cwd, stdio});
+    execSync(`cpx esm/**/* ../esm`, {cwd, stdio});
+    execSync(`cpx cjs/**/* ../cjs`, {cwd, stdio});
+} catch (e) {
```
texodus, line 16: This is what's left of `scripts/build_js.js`.  In addition to calling `emcmake` and `emmake`, it moves these directors up a level so that `@finos/perspective` later needs not know whether `"release"` or `"debug"` directory has the right files.

cpp/perspective/package.json
```diff
@@ -0,0 +1,19 @@
...
+        "dist/cjs/**/*"
+    ],
+    "scripts": {
+        "build": "yarn emsdk-run node ./build.js",
+        "clean": "rimraf dist/cpp/build",
+        "test": ":",
+        "test:run": ":",
+        "test:build": ":"
+    }
+}
```
texodus, line 19: `@finos/perspective-cpp` packages adds this build cycle to the build yarn workspaces build graph, and wraps up the minimal leftover JavaScript build scripts.  Even though this is private, we use `"files"` param because we will later on use `npm-pack-here` to "deploy" this module within `@finos/perspective/dist`, effectively static-linking modules in JavaScript 💯

sc1f:
I have been playing around with this PR, and the changes look good - I added a small QOL improvement in the setup script that allows you to select packages using a checkbox system instead of a single package at a time, which will be helpful to build `perspective-cpp` and `perspective` at the same time, or for selective builds; for example, when I'm working on the Jupyter plugin I often want to make a change to just `perspective` and `perspective-jupyterlab` to pick it up, which was not possible through the setup script previously.

A GIF of the new setup:

setup2

@texodus let me know what you think; one of the caveats is that selecting "all" or "none" of the items should really print "all" after the selection has been processed, but I don't think there's a way to do that at the moment.

sc1f (approved):
Looks good - tried out locally for a bit and everything works well, as expected.

PR #1294, opened 2021-01-17
JupyterLab update to ^3.0.0, and fix "Open With.." regression

timkpaine:
- Bump supported JupyterLab version from 2. to 3.
- Add `.arrow` renderer (in addition to `.csv`, `.json`, and `.jsonl`)
- `eslint` the typescript


timkpaine:
@texodus did you make web pack changes? we have a regression and are hitting [this error](https://github.com/jupyter-widgets/ipywidgets/issues/2634) again.

timkpaine:
I can only seem to get around it with `// ts-ignore`

timkpaine:
has some issues with the renderers, wait to merge

timkpaine:
looks like this PR broke everything https://github.com/finos/perspective/pull/1231

timkpaine:
All set, load up `csv`, `json`, or now `arrow` from disk!

tmp

packages/perspective-jupyterlab/src/ts/plugin.ts
```diff
@@ -19,16 +19,19 @@ import {PerspectiveView} from "./view";
...
+const EXTENSION_ID = "@finos/perspective-jupyterlab";
+
 /**
  * PerspectiveJupyterPlugin Defines the Jupyterlab plugin, and registers `PerspectiveModel` and `PerspectiveView`
  * to be called on initialization.
  */
 export const PerspectiveJupyterPlugin: IPlugin<Application<Widget>, void> = {
-    id: "@finos/perspective-jupyterlab",
+    id: EXTENSION_ID,
+    // @ts-ignore
```
texodus, line 30: For the record, I, personally, don't care about the use of `@ts-ignore`.  But, I can't pass up on the opportunity to point out, only mostly satirically, that there's a great TypeScript option for disabling type errors called _JavaScript_ that would work wonders here.

packages/perspective-jupyterlab/src/ts/psp_widget.ts
```diff
@@ -18,6 +16,9 @@ import {MIME_TYPE, PSP_CLASS, PSP_CONTAINER_CLASS, PSP_CONTAINER_CLASS_DARK} fro
 
 import {HTMLPerspectiveViewerElement, Pivots, Aggregates, Sort, ComputedColumns, PerspectiveViewerOptions, Filters, Columns} from "@finos/perspective-viewer";
 
+// eslint-disable-next-line @typescript-eslint/no-var-requires
+const perspective = require("@finos/perspective");
```
texodus, line 20: Webpack advises [not mixing `require` and `import`](https://stackoverflow.com/questions/56068126/is-it-possible-to-use-both-require-and-import-together-with-webpack) - is this _required_  because the TypeScript bindings are not correct or something?

packages/perspective-jupyterlab/src/ts/psp_widget.ts
```diff
@@ -152,14 +153,20 @@ export class PerspectiveWidget extends Widget {
      * @param data
      */
     _update(data: TableData): void {
-        this.viewer.table.update(data);
+        if (this.viewer.table === undefined) {
+            this.viewer.load(perspective.worker().table(data));
+        } else {
+            this.viewer.table.update(data);
+        }
```
texodus, line 160: As I mentioned in the #1296, I'd prefer not to replicate this behavior that was just removed from `<perspective-viewer>` itself.  I understand if we want to keep these actual methods on the `PerspectiveViewer` class so we do not need to model virtual `ipywidgets` interfaces for `Table` and `<perspective-viewer>`, but we can definitely remove any proxied code in Python which calls `update()` before `load()` on a `PerspectiveViewer`.

packages/perspective-jupyterlab/src/ts/renderer.ts
```diff
@@ -26,7 +36,7 @@ const baddialog = () => {
 
 export class PerspectiveDocumentWidget extends DocumentWidget<PerspectiveWidget> {
     constructor(options: DocumentWidget.IOptionsOptionalContent<PerspectiveWidget>, type: IPerspectiveDocumentType = "csv") {
-        super({content: new PerspectiveWidget("test"), context: options.context, reveal: options.reveal});
+        super({content: new PerspectiveWidget("Perspective"), context: options.context, reveal: options.reveal});
```
texodus, line 39: This should be unique, no?

texodus (changes requested):
Thanks for the PR!  The new feature is cool but I'd like to see the `PerspectiveViewer` TypeScript API fixed to match the `<perspective-viewer>` API.

Again, a test or at least description of the actual broken case would really help.  It would prevent regressions as #1296 and, respectfully, it is difficult to understand from the info provided if this is fix for a bug, or a personal preference.

packages/perspective-jupyterlab/src/ts/psp_widget.ts
texodus: One possible resolution is to move these specific changes to their own PR to capture the Issue, so we could merge this.

packages/perspective-jupyterlab/src/ts/psp_widget.ts
timkpaine: yes

packages/perspective-jupyterlab/src/ts/plugin.ts
timkpaine: we're bound somewhat by the examples published by JupyterLab, the typescript extension is really the only maintained one it seems.

packages/perspective-jupyterlab/src/ts/renderer.ts
timkpaine: I don't think so, why would it need to be?

packages/perspective-jupyterlab/src/ts/psp_widget.ts
timkpaine: I can also just create the table over [here](https://github.com/finos/perspective/blob/master/packages/perspective-jupyterlab/src/ts/renderer.ts#L50) and leave `PerspectiveWidget` alone. makes no difference to me, though someone should probably check that `client` only mode for the widget is still working

packages/perspective-jupyterlab/src/ts/plugin.ts
texodus: Sounds like what you're saying is, if I were to de-TypeScript this module, you wouldn't object?

packages/perspective-jupyterlab/src/ts/psp_widget.ts
texodus: Ah, so is this what prompted #1296?  If so yes, you should definitely construct the `Table` there.  Also - you should register this table as property of `PerspectiveDocumentWidget` and call its `delete()` method [here](https://github.com/finos/perspective/blob/master/packages/perspective-jupyterlab/src/ts/renderer.ts#L76) - or whatever the equivalent JupyterLab API is.

packages/perspective-jupyterlab/src/ts/plugin.ts
timkpaine: Nope, `ts-ignore` is the final straw for me. It's working perfectly and yet due to the craziness of widgets and jlab 3 it will never compile. If you de-typescript i think you can also delete the perspective typings.

packages/perspective-jupyterlab/src/ts/psp_widget.ts
timkpaine: Yep makes sense, will do it tomorrow

texodus (approved):
Looks good!

PR #1298, opened 2021-01-20
Fix partial updates in Python using dicts

### LINKED ISSUE #1268: Inconsistency with table.update() in Python with Dicts
## Bug Report

First off, thank you for making this great package.

I have noticed a undocumented inconsistency with the `table.update` function in the Python API.
If you format the `update` data as a dict that maps to lists/arrays, it _replaces_ the entry at the index instead of _updating_ the fields.

```python
TABLE.update({"index": ["b"], "num": [0], "static": [0]})  # initial data
TABLE.update({"index": ["b"], "num": [1]})  # this REPLACES the info at index "b" instead of updates
```
in the above code, the second update deletes the `static` value of `index` "b"

Basically, I expected
```
"index": "b", "num": 0, "static": 0   -->   "index": "b", "num": 1, "static": 0
```
but got
```
"index": "b", "num": 0, "static": 0   -->   "index": "b", "num": 1
```
<br>

If one does
```python
BASE_TABLE.update([{"index": "b", "num": i}])  # no list inside the dictionary
```
it works though.

### Steps to Reproduce:

Here is an MVP of the bug that connects to LocalHost

```python

import asyncio
import threading

import tornado
import perspective


def perspective_thread(MANAGER):
    loop = tornado.ioloop.IOLoop()
    MANAGER.set_loop_callback(loop.add_callback)
    loop.start()


class MainHandler(tornado.web.RequestHandler):
    def get(self):
        html = """<!DOCTYPE html>
<html>

<head>
    <meta http-equiv="Content-Type" content="text/html; charset=utf-8">
    <title>{{ table_name }}</title>

    <meta name="viewport"
        content="width=device-width, initial-scale=1, maximum-scale=1, minimum-scale=1, user-scalable=no">
        <script src="https://cdn.jsdelivr.net/npm/@finos/perspective-workspace"></script>
        <script src="https://cdn.jsdelivr.net/npm/@finos/perspective-viewer-datagrid"></script>
        <script src="https://cdn.jsdelivr.net/npm/@finos/perspective-viewer-d3fc"></script>
        <script src="https://cdn.jsdelivr.net/npm/@finos/perspective/dist/umd/perspective.js"></script>

</head>
<style>
    perspective-viewer {
        position: absolute;
        top: 0;
        left: 0;
        right: 0;
        bottom: 0;
    }
</style>

<body>
    <perspective-viewer id="viewer" , plugin="datagrid"></perspective-viewer>
    <script>
        websocket_url = 'ws://localhost:8080/websocket'
        table_name = 'test_table'

        window.addEventListener('WebComponentsReady', async function () {
            const viewer = document.getElementById('viewer');
            const websocket = perspective.websocket(websocket_url);
            const table = websocket.open_table(table_name);
            viewer.load(table);
            viewer.toggleConfig();
        });
    </script>
</body>

</html>"""
        self.write(html)


def make_app(MANAGER,):
    handlers = [
        (r"/", MainHandler,),
        (
            r"/websocket",
            perspective.PerspectiveTornadoHandler,
            {"manager": MANAGER, "check_origin": True},
        ),
    ]
    app = tornado.web.Application(handlers)
    return app


async def main(port):
    MANAGER = perspective.PerspectiveManager()
    columns = {"index": str, "num": int, "static": int}
    BASE_TABLE = perspective.Table(columns, index="index",)
    MANAGER.host_table("test_table", BASE_TABLE)

    thread = threading.Thread(target=perspective_thread, args=(MANAGER,))
    thread.daemon = True
    thread.start()

    app = make_app(MANAGER)
    app.listen(port)

    i = 0
    BASE_TABLE.update({"index": ["b", "c"], "num": [0, 1], "static": [0, 1]})
    while True:
        await asyncio.sleep(1)
        BASE_TABLE.update({"index": ["b"], "num": [i]})
        # BASE_TABLE.update([{"index": "b", "num": i}])
        i += 1


if __name__ == "__main__":
    port = 8080
    asyncio.run(main(port))
```

### Environment:
linux, py38

sc1f:
Fixes #1268 by refactoring the data loading path for dict-shaped datasets.

There are some inconsistencies between how `None` values are handled in JS and in Python. In JS, `null` and `undefined` in a dataset mean different things in a partial update:

```javascript
table = perspective.table({a: [1, 2, 3, 4], b: ["a", "b", "c", "d"], c: [10, 20, 30, 40]}, {index: "a"})
table.update({a: [3], c: [100]}) // "b" is `undefined` here, so values in `b` are not modified.
```

`undefined` resolves to a no-op, so values that are `undefined` will not be modified or overwritten. However, if the value is replaced with null:

```javascript
table.update({a: [3], b: [null], c: [100]}) // "b" is explicitly "null", which will unset the value at "b" for pkey `3` and set it to null.
```

To remedy this, the Python library uses a method called `_has_column`, which ascertains whether a given row contains a given column name. If the column name does not exist, it is a no-op similar to `undefined`, but if the column name exists then it will be overwritten with the new value.

Thus, a row-oriented update of `[{a: undefined}]` that would work as a no-op update in Javascript _does not_ work in Python. In the column-oriented case, #1268 illustrates the issue where a missing column in a columnar dataset would be treated as an _overwrite_, and not a no-op. This PR fixes the behavior by treating missing columns in columnar datasets as no-ops, and has been tested. This behavior is now equivalent between JS and Python.

There will always remain Python-specific idiosyncrasies around partial updates. For example, this update works in JS:

```javascript
table.update({
  a: [2, 3, 4],
  b: [null, undefined, null] // delete, no-op, delete
})
```

but it would not work in Python, because one cannot satisfy the "all columns must have the same # of rows" requirement _AND_ mark a `column[row]` as a no-op, and there is no way to reconcile this behavior:

```python
table.update({
  "a": [2, 3, 4],
  "b": [None, None, None] # delete, delete, delete
})
```

### DIFF
--- a/packages/perspective/test/js/updates.js
+++ b/packages/perspective/test/js/updates.js
@@ -1576,7 +1576,7 @@ module.exports = perspective => {
     });
 
     describe("implicit index", function() {
-        it("should apply single partial update on unindexed table using row id from '__INDEX__'", async function() {
+        it("should partial update on unindexed table", async function() {
             let table = perspective.table(data);
             table.update([
                 {
@@ -1591,6 +1591,73 @@ module.exports = perspective => {
             expected[2]["y"] = "new_string";
 
             expect(result).toEqual(expected);
+            view.delete();
+            table.delete();
+        });
+
+        it("should partial update on unindexed table, column dataset", async function() {
+            let table = perspective.table(data);
+            table.update({
+                __INDEX__: [2],
+                y: ["new_string"]
+            });
+
+            let view = table.view();
+            let result = await view.to_json();
+
+            // does not unset any values
+            expect(result).toEqual([
+                {x: 1, y: "a", z: true},
+                {x: 2, y: "b", z: false},
+                {x: 3, y: "new_string", z: true},
+                {x: 4, y: "d", z: false}
+            ]);
+            view.delete();
+            table.delete();
+        });
+
+        it("should partial update and unset on unindexed table", async function() {
+            let table = perspective.table(data);
+            table.update([
+                {
+                    __INDEX__: 2,
+                    y: "new_string",
+                    z: null
+                }
+            ]);
+
+            let view = table.view();
+            let result = await view.to_json();
+
+            // does not unset any values
+            expect(result).toEqual([
+                {x: 1, y: "a", z: true},
+                {x: 2, y: "b", z: false},
+                {x: 3, y: "new_string", z: null},
+                {x: 4, y: "d", z: false}
+            ]);
+            view.delete();
+            table.delete();
+        });
+
+        it("should partial update and unset on unindexed table, column dataset", async function() {
+            let table = perspective.table(data);
+            table.update({
+                __INDEX__: [0, 2],
+                y: [undefined, "new_string"],
+                z: [null, undefined]
+            });
+
+            let view = table.view();
+            let result = await view.to_json();
+
+            // does not unset any values
+            expect(result).toEqual([
+                {x: 1, y: "a", z: null},
+                {x: 2, y: "b", z: false},
+                {x: 3, y: "new_string", z: true},
+                {x: 4, y: "d", z: false}
+            ]);
             view.delete();
             table.delete();
         });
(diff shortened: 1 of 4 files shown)


texodus (approved):

PR #1306, opened 2021-01-22
Remove cross module boundary imports

LukeSheard:
I was going to raise an issue for this but I thought it was simple enough to propose a fix instead... 

I noticed that when running in a commonjs only environment (i.e. `jest`) that imports of `@finos/perspective-viewer-.*` reach across the module boundary into the `esm` build of `@finos/perspective` even though the `cjs` build of the files is already being used. This causes issues without further transpilation, but will also cause builds to duplicate these modules. 

I fixed this issue locally by simply refactoring imports to use the top level module import which allows the resolution process to resolve the correct relative filenames.

### DIFF
--- a/packages/perspective-viewer-d3fc/src/js/plugin/plugin.js
+++ b/packages/perspective-viewer-d3fc/src/js/plugin/plugin.js
@@ -7,7 +7,7 @@
  *
  */
 
-import {registerPlugin} from "@finos/perspective-viewer/dist/esm/utils.js";
+import {registerPlugin} from "@finos/perspective-viewer";
 import charts from "../charts/charts";
 import "./polyfills/index";
 import "./template";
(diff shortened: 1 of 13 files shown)


texodus:
Thanks for the PR!  Unfortunately this fix does not work;  in addition to causing the WebAssembly assets to be rebuilt and disted to every package (instead of just the core package), this change fails the test suite.

I am closing this for now since I am certain this fix will not work (without rewriting the WAM loader), though I acknowledge that importing this file across project boundaries is hacky.   We had considered potentially adding a new package for this one (<1k) file, but this would introduce API-less internal packages into Perspective's released packages universe.  Feel free to open a new PR if you have an update or a new take (though probably run `yarn test` first).

LukeSheard:
I'm a bit confused as to why the CI is failing for this PR, given that locally it's all passing and all the tests are passing in Azure. There seems to be an issue with the teardown inside the CI that I can't repro or isolate locally. Do you have any ideas where it could come from?

PR #1308, opened 2021-01-22
FIX #1304 changing hypergrid to datagrid in olympics example

### LINKED ISSUE #1304: Olympics example still referring to hypergrid
### Steps to Reproduce:
visit https://bl.ocks.org/texodus/efd4a857aca9a52ab6cddbb6e1f701c9

### Expected Result:
no errors in the console
### Actual Result:
It renders but perspective complains in the console `Unknown plugin "hypergrid"`
### Environment:
windows 10
chrome 87
perspective 0.6.0
### Additional Context:
Non-breaking bug

zepaz:
minor:
changing hypergrid to datagrid as hypergrid is no longer bundled with perspective


examples/blocks/src/custom/index.html
```diff
@@ -77,14 +80,15 @@
                 const name = metadata.value;
                 const span = document.createElement("span")
                 const img = document.createElement("img");
-                img.onload = () => {
-                    img.onload = undefined;
-                    const data = getBase64Image(img);
-                    img.src = data;
-                }
+<!--                img.onload = () => {-->
```
sc1f, line 83: Looks like an HTML comment here and not JS?

zepaz: True, however in testing it was not a blocker, but just to be clear I have removed the whole commented out section.

examples/blocks/src/custom/index.html
```diff
@@ -77,14 +80,9 @@
...
                 const img = document.createElement("img");
-                img.onload = () => {
-                    img.onload = undefined;
-                    const data = getBase64Image(img);
-                    img.src = data;
-                }
                 img.setAttribute("crossorigin", "anonymous")
-                img.setAttribute("src", `http://perspective.finos.org/img/flags/${states[clean_name].toLowerCase()}.png`);
-   
+                img.setAttribute("src", `http://flags.ox3.in/svg/us/${states[clean_name].toLowerCase()}.svg`);
```
texodus, line 84: I think you left some debugging in

examples/blocks/src/custom/index.html
```diff
@@ -106,12 +104,12 @@
                 return obj;
             }, {});
         }
-
         window.addEventListener('DOMContentLoaded', async function () {
             const viewer = document.getElementsByTagName('perspective-viewer')[0];
-            const data = await fetch("/node_modules/superstore-arrow/superstore.arrow");
+            var worker = perspective.worker()
+            const data = await fetch("https://cdn.jsdelivr.net/npm/superstore-arrow/superstore.arrow");
```
texodus, line 110: "

texodus (changes requested):
Thanks for the PR!

You're right, this example should use Datagrid - but there seem to be some other changes in this PR that introduce some new external URLs (yours?) and remaps some local debug URLs to CDNs.  Let's please revert these changes to `index.html`, then it should be good to merge.

examples/blocks/src/custom/index.html
zepaz: To be clear source was not mine, and changed it because hotlinking from finos.org was not allowed so flags were not rendered. Will include the flags as a local resource; the SVG files are in the public domain anyway

examples/blocks/src/custom/index.html
zepaz: Will change back

texodus:
@zepaz I appreciate the effort here but this is literally a 1-line change, and after 2 code reviews there are now **67k** lines of code in this PR including a littany of new changes since the last review that were not requested.  I do not want the SVGs in the `examples/` dir, because they cannot be uploaded to http://bl.ocks.org/ where we host them.  The extra changes to the logic in the example itself is not explained, such as the `` tag cache removal.

Thanks for the heads up on this, I will update this example with the 0.6.1 release.  Closing.

PR #1310, opened 2021-01-22
Reproduce #1312 - broken US date string parsing

sc1f:
In order for US locale strings to be treated as valid by the filter validator, the locale string format must be added to the parser format definition in `arrow_csv.cpp`. This has the side effect of parsing US locale strings as part of the dataset, where I found an issue with strptime in the C STL - when given the US locale string format ("%m/%d/%Y, %I:%M:%S %p"), it parses "12:00:00 AM" as "12:00:00 PM", which is incorrect. 

This seems to be an issue with strptime when compiled in Emscripten; the Python library does not have the same issue. Users should pass in date values as Date() objects, but there can be cases where strings are being passed in and this parsing error occurs. The solution here looks to be to add a custom parser for US locale strings, similar to the custom ISO format parser added by @texodus.

### DIFF
--- a/packages/perspective/test/js/constructors.js
+++ b/packages/perspective/test/js/constructors.js
@@ -982,7 +982,7 @@ module.exports = perspective => {
             table.delete();
         });
 
-        it.skip("Handles datetime strings in US locale string", async function() {
+        it("Handles datetime strings in US locale string", async function() {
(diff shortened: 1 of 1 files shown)


texodus:
Closing for #1312

PR #1311, opened 2021-01-31
Remove webpack as a peerDependency for the webpack plugin

threepointone:
This PR removes webpack as a peerDependency for `@finos/perspective-webpack-plugin`

IMO, a peer dependency (or dependency) should only be included if it's actually required by the source code. (For the same reason, `file-loader` and `worker-loader` _should_ be included because they're implicitly required.) Keeping this peerDependency makes it so that the consuming package has to declare webpack as a dependency, even if they're only passing the plugin to a tool that transitively uses webpack.

### DIFF
--- a/packages/perspective-webpack-plugin/package.json
+++ b/packages/perspective-webpack-plugin/package.json
@@ -24,7 +24,6 @@
         "worker-loader": "^3.0.7"
     },
     "peerDependencies": {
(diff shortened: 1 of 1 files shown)


texodus (approved):
Thanks for the PR!  Fair point, and thanks for the clear explanation of the use case.

PR #1318, opened 2021-02-05
Fix for layout jitter in the Column Selector

texodus:
This is a CSS cleanup/refactor of `@finos/perspective-viewer`'s Column Selector widget, which aims to eliminate "jitter" in the widget's dimensions and layout as columns are activated or the container's size is changed.  The responsive logic which divides this column into "active column list", "inactive column list" and "new column widget" sections does a pretty good job of intelligently using the available space, but has bugs.  In this _Before GIF_, notice that 
 * The "New Column" dialog moves down when opened.
 * When the first column is deselected, the entire panel becomes slightly wider.
 * .. and also the "New Column" editor drops to the bottom of the container (off camera).
 * As each column is deselected, the "deselected" columns list creeps upwards by ~1-2px.
 * When closed, the "New Column" dialog is in a different place in relation to the bottom of the previous panel than when there were no columns deselected.

old

_After GIF_:

new

In this _Before GIF_, when a column is dragged-over a column-accepting input, the input shrinks and drags the rest of the flex container with it.

old2

_After GIF_:

new2

In this _Before GIF_, when the "Column Name" column is dragged into the Active Column list, the list expands by 5px forcing the datagrid to the right:

old3

_After GIF_:

new3


sc1f (approved):
Pulled down locally - looks good!

PR #1320, opened 2021-02-06
Fix missing return in ctx0::notify

sc1f:
This puts back a return statement that was erroneously removed in t_ctx0::notify, which caused filtered contexts to fail when updates were read out. I've added a test for this issue and verified the breakage before the fix was applied.


texodus (approved):
Thanks for the PR and test!

Doubly-so, since I introduced this regression in the first place ...

PR #1331, opened 2021-02-19
Fix #1324: use ~major.minor.patch in PerspectiveWidget versioning

### LINKED ISSUE #1324: `perspective-python` should not use tilde (~) in client version resolution
### Steps to Reproduce:
 1. Implement a `voila` application, but build it in an NPM environment where you have inconsistent (or non-existent) versions of `@finos/perspective-jupyterlab` or other dependencies.
 2. Use a CDN fallbkack, and host the assets instead on a version-qualified server that doesn't support tilde (~) version bounds.

### Expected Result:
Honest, I don't know what a sane developer would expect to happen here but _I_ expect this to fail spectacularly for very obvious reasons.

### Actual Result:
Quite.  We think the tilde is probably unnecessary here though so we'll remove it.

sc1f:
This PR fixes #1324 by using the full version in `_model_module_version` and `_view_module_version`, which should fix the scenario described in the issue.

Before, the version strings were `~{MAJOR}.{MINOR}`, so `~0.5`, `~0.6` etc. Now, the version strings are `~{MAJOR}.{MINOR}.{PATCH}`, so the exact version of Perspective as specified in the root `package.json`: `~0.6.2`, etc. With the tilde this should resolve to `0.6.2 < 0.7.0`.

### DIFF
--- a/python/perspective/perspective/widget/widget.py
+++ b/python/perspective/perspective/widget/widget.py
@@ -21,7 +21,7 @@ from ..core.data import deconstruct_pandas
 from ..core.exception import PerspectiveError
 from ..libpsp import is_libpsp
 from ..viewer import PerspectiveViewer
-from ..core._version import major_minor_version
+from ..core._version import __version__
 
 
 def _type_to_string(t):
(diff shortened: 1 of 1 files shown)


timkpaine:
this introduces a flavor of the same problem, which is that it makes it impossible to upgrade one side without upgrading the other side due to strict semver checking. 

e.g. if you want to fix a bug on the python side, you will also have to rebuild and redeploy javascript, even if its not otherwise modified.

timkpaine (approved):

sc1f:
We already do this though - there is no situation in which the version of perspective-python would != the version of the Javascript packages, since the entire versioning scheme is hooked up to `lerna dist` which updates all lerna-ized packages at The only part that isn't hooked up to `lerna dist` is disting new versions to pypi, which we are admittedly behind on right now.

timkpaine:
I have firsthand experience with a framework in which you could control backend updates, but not frontend updates. The reason we switched to this versioning scheme in the first place was to make it easier for your team to get fixes changes out on the backend without worrying about my team getting the JupyterLab extension updated

texodus:
@timkpaine I appreciate the insight and acknowledge that this change will make it more difficult for version-restricted/role-partitioned environments to upgrade ...  _c'est la vie_.

PR #1335, opened 2021-02-25
fixes #1334, adds py39 now that brew defaults this on osx

### LINKED ISSUE #1334: Python wheel builds failing on catalina for py36 and py38
https://dev.azure.com/finosfoundation/perspective/_build/results?buildId=2147&view=results

We should `pip install -U` wheel [here](https://github.com/finos/perspective/blob/1725f48d6b15fc5927a34524910ab65c47ceef76/scripts/_wheel_python.js#L67) and see if that helps.

Should also add it as a dev dependency [here](https://github.com/finos/perspective/blob/1725f48d6b15fc5927a34524910ab65c47ceef76/python/perspective/setup.py#L77).

timkpaine:


timkpaine:
need to move to main repo to check tags

PR #1336, opened 2021-02-26
Deprecate py27 (linux), add py39 (osx)

### LINKED ISSUE #1334: Python wheel builds failing on catalina for py36 and py38

timkpaine:
Fixes #1334, adds py39 now that brew defaults this on OSX.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -73,11 +73,11 @@ jobs:
 
   strategy:
     matrix:
-      Python27ManyLinux2010:
-        python.version: '2.7'
-        python_flag: '--python2'
-        manylinux_flag: '--manylinux2010'
-        artifact_name: 'cp27-cp27m-manylinux2010_x86_64'
+      # Python27ManyLinux2010:
+      #   python.version: '2.7'
+      #   python_flag: '--python2'
+      #   manylinux_flag: '--manylinux2010'
+      #   artifact_name: 'cp27-cp27m-manylinux2010_x86_64'
       Python37ManyLinux2010:
         python.version: '3.7'
         python_flag: ''
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -244,7 +244,7 @@ jobs:
         artifact_name: 'cp37-cp37m-macosx_10_15_x86_64'
       Python39:
         python.version: '3.9'
-        python_flag: ''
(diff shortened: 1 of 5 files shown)


timkpaine:
builds all passing except the 2 that require docker changes, so looks good

sc1f (approved):
Looks good - will need a rebuild of the Docker images, and a full tagged build run to check if the wheel builds work.

texodus (changes requested):
The `Dockerfile` modifications in this file do not build - it appears as if the method on step 42 for acquiring `nodejs` via `yum` has bit-rotted, which I recognize was not touched by this PR, but nonetheless ...

The experience of discovering this after >1hr of `docker build`, suggests strongly to me we need a better method of building Python, perhaps we build these images exclusively from Azure?  Are these images needed still now that the `pyarrow` dependency has been removed? 

Regardless, two things are true, (1) I cannot merge this PR until the `Dockerfile` builds, and (2) addressing this tech debt must pre-empt any future image dists.

timkpaine:
I've updated and tested building the docker images from scratch. They are necessary as azure does not provide a `manylinux` compatible environment, so in order to dist linux wheels we need these docker images.

timkpaine:
(I've also dropped python2 in these last changes)

texodus:
> (I've also dropped python2 in these last changes)

Nice try - I've refactored these changes to #1374

texodus (approved):

texodus:
Upon further review, the `pypa/manylinux2010_x86_64` container we relied upon is now non-reproducible, as the upper-bound of python2 compatible versions is above the lower-bound of CentOS 6 compatible versions.  This makes python2 effectively deprecated, even if this image was preserved.

Which means that `0.6.2` is the last python2 "compatible" version that will be published - insofar as `perspective-python` can be compatible with a language whose own support has been deprecated.

Issue #1337, opened 2021-02-26
Drop support for python 3.6


timkpaine:
3.6 ends in 10 months and our full CI suite is getting quite large. Additionally, our default coding environment has shifted even further away with the standard installed python version moving recently from python3.7 to python3.9. We should consider dropping support for python3.6.

Python EOL:
https://endoflife.date/python

Full CI:
![Screen Shot 2021-02-26 at 10 54 17](https://user-images.githubusercontent.com/3105306/109323142-22523980-7821-11eb-9a41-fb7e69e1acaa.png)

Python 3.9 upgrade:
https://github.com/Homebrew/homebrew-core/issues/62201

texodus:
As long as 3.6 continues to build and pass the test suite, what harm is there in leaving it?  Don't get me wrong, I am all for deprecating irrelevant technology, but if we're going to deprecate Python versions I think there is another even _older_ version that forms the basis for _substantially_ more technical debt than 3.6 ...

timkpaine:
Discussed offline, we will delay for now in favor of attempting to run complete CI more regularly.

PR #1338, opened 2021-03-02
Add nightly complete builds

timkpaine:
resolve issues raised in https://github.com/finos/perspective/issues/1337


texodus (approved):

I've rebased (to resolve the `emsdk` regression), and also made these into "weeklies" rather than "nightlies" (just temporarily as I validate how this CI load impacts our dev process).

PR #1349, opened 2021-03-15
Fix timezone tests to take DST into account

sc1f:
In our suite of timezone tests, we set the timezone manually and assert that `getTimezoneOffset` is 300 (for EST). However, during Daylight Savings time (which started on Sunday, March 14th) the timezone offset is 240, and our tests did not take that into account. This PR fixes the issue and now should work for both DST and non-DST times.

### DIFF
--- a/packages/perspective/test/js/timezone/timezone.spec.js
+++ b/packages/perspective/test/js/timezone/timezone.spec.js
@@ -64,7 +64,17 @@ const check_datetime = (output, expected) => {
 describe("Timezone Tests", () => {
     beforeAll(() => {
         expect(process.env.TZ).toBe("America/New_York");
-        expect(new Date().getTimezoneOffset()).toBe(300);
+        const now = new Date();
+
+        // offset is 240 during DST and 300 during non-DST, since we manually
+        // set the timezone here just assert the timezone string
+
+        if (now.toString().includes("GMT-0400") || now.toString().includes("Eastern Daylight Time")) {
+            expect(new Date().getTimezoneOffset()).toBe(240);
+        } else {
+            expect(new Date().getTimezoneOffset()).toBe(300);
+        }
+
         console.log("Timezone set to", process.env.TZ);
     });
 


yarn.lock
```diff
@@ -1255,6 +1255,91 @@
     unique-filename "^1.1.1"
     which "^1.3.1"
 
+"@finos/perspective-viewer-d3fc@^0.5.1":
```
timkpaine, line 1258: probably can omit these yarn.lock changes

timkpaine (changes requested):

yarn.lock
sc1f: fixed

timkpaine (approved):

timkpaine:
@sc1f looks like they mightve moved boost again on windows 🤦

sc1f:
@timkpaine PR build successful though, which is strange - let me bump the windows job

sc1f:
@timkpaine https://github.com/actions/virtual-environments/issues/2667 welp - I will make a PR fixing this

texodus:
Honest question - why do we depend on system boost for Windows instead of building the version we want as we do [for wasm](https://github.com/finos/perspective/blob/master/scripts/install_tools.js)?

sc1f:
@texodus we don't have a docker image for windows; I think we may need the built boost and not just the headers, but I will check when I do the fix.

timkpaine:
@texodus if you want to make that script cross-platform...

texodus:
@timkpaine I made this change on the wasm workflow because I measured it was faster overall to build boost from source, as opposed to loading the Docker image, if you include cumulative build time saved from skipping the container.  Ideally this will use Azure's cache to build `boost` once per week or so, but I've yet to get this part to work - regardless the total boost time is ~3min [source](https://dev.azure.com/finosfoundation/perspective/_build/results?buildId=2194&view=logs&j=430750e5-873f-5cdb-f7d1-334fffa69fc8&t=e9a709ee-98dc-5d78-ab76-09541ac9f110) which is longer than downloading the container by ~1min, but significantly faster to build perspective overall for wasm outside of the container.  This won't translate 1:1 with Python but it won't be off by more than a handful of minutes on what is already a 20min Python build on CI, per version.

We may not need to actually build boost completely, but AFAIK `boost` headers are not linkable without the `bootstrap` and `b2` steps [here](https://github.com/boostorg/wiki/wiki/Getting-Started%3A-Overview#installing-boost) [source](https://stackoverflow.com/questions/25158084/header-files-missing-in-boost-build).

texodus:
Thanks for the PR!

PR #1350, opened 2021-03-20
update example webpack config 

f0ster:


f0ster:
I have executed the CLA, can we retrigger the check here?

PR #1351, opened 2021-03-21
install boost via choco

timkpaine:


texodus:
Now that's a good idea

timkpaine:
@texodus this is also how i build on windows, so its closer to real life

texodus (approved):

PR #1353, opened 2021-03-23
Enable editing to mime renderer in JupyterLab

timkpaine:
This is a simple modification that turns on editing for our JupyterLab mime renderer plugin (the thing that lets you right click on a `csv`, `json`, or `arrow` and visualize in perspective). As a result, we now have a pivotable, sortable, filterable, **editable** csv/json/arrow viewer.

tmp

- [x] add mime renderer to jlab readme and usage page
- [x] fix arrow saving

### DIFF
--- a/docs/md/python.md
+++ b/docs/md/python.md
@@ -481,6 +481,10 @@ PerspectiveWidget(table)
 PerspectiveWidget(None)
 ```
 
+## `PerspectiveRenderer`
+
+Perspective also exposes a JS-only `mimerender-extension`. This lets you view `csv`, `json`, and `arrow` files directly from the file browser. You can see this by right clicking one of these files and `Open With->CSVPerspective` (or `JSONPerspective` or `ArrowPerspective`). Perspetive will also install itself as the default handler for opening `.arrow` files.
+
 ## `PerspectiveTornadoHandler`
 
 Perspective ships with a pre-built Tornado handler that makes integration with
--- a/docs/md/python.md
+++ b/docs/md/python.md
@@ -483,7 +483,7 @@ PerspectiveWidget(None)
 
 ## `PerspectiveRenderer`
 
-Perspective also exposes a JS-only `mimerender-extension`. This lets you view `csv`, `json`, and `arrow` files directly from the file browser. You can see this by right clicking one of these files and `Open With->CSVPerspective` (or `JSONPerspective` or `ArrowPerspective`). Perspetive will also install itself as the default handler for opening `.arrow` files.
+Perspective also exposes a JS-only `mimerender-extension`. This lets you view `csv`, `json`, and `arrow` files directly from the file browser. You can see this by right clicking one of these files and `Open With->CSVPerspective` (or `JSONPerspective` or `ArrowPerspective`). Perspective will also install itself as the default handler for opening `.arrow` files.
 
 ## `PerspectiveTornadoHandler`
 
--- a/packages/perspective-jupyterlab/src/ts/renderer.ts
+++ b/packages/perspective-jupyterlab/src/ts/renderer.ts
@@ -92,7 +92,7 @@ export class PerspectiveDocumentWidget extends DocumentWidget<PerspectiveWidget>
 
                 // create a flat view
                 const view = await table.view();
-                view.on_update(() => {
+                view.on_update(async () => {
                     if (this._type === "csv") {
                         const result: string = await view.to_csv();
                         this.context.model.fromString(result);
--- a/packages/perspective-jupyterlab/src/ts/renderer.ts
+++ b/packages/perspective-jupyterlab/src/ts/renderer.ts
@@ -38,9 +38,10 @@ const baddialog = (): void => {
 };
 
 const WORKER = perspective.worker();
+
 export class PerspectiveDocumentWidget extends DocumentWidget<PerspectiveWidget> {
     constructor(options: DocumentWidget.IOptionsOptionalContent<PerspectiveWidget>, type: IPerspectiveDocumentType = "csv") {
-        super({content: new PerspectiveWidget("Perspective"), context: options.context, reveal: options.reveal});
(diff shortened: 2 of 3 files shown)


timkpaine:
Round trip arrow save/load

tmp

packages/perspective-jupyterlab/src/ts/renderer.ts
```diff
@@ -82,21 +82,45 @@ export class PerspectiveDocumentWidget extends DocumentWidget<PerspectiveWidget>
...
+                // construct new table
+                this._psp.viewer.load(this._table);
+            } else {
+                // replace existing table for whatever reason
+                this._psp.replace(this._table);
+            }
+
+            // create a flat view
+            const view = this._table.view();
+            view.on_update(() => {
```
texodus, line 100: Does this work/is actually called in test?  This and other methods are async since #1289 so `this._table` and the return type of `view()` are both `Promise` here, I believe.

texodus:
Also - this is a really cool feature!

packages/perspective-jupyterlab/src/ts/renderer.ts
timkpaine: I'll check again, we have some crazy yarn link stuff to worry about these days

timkpaine:
something to note from our JS docs
Screen Shot 2021-03-23 at 9 46 56 PM

texodus:
Don't understand - it's not "fixed", #1289 implements the async change itself, as well as accompanying docs update.  `0.6.2`, the latest release, does not have this change yet, hence the docs reflect the current release.

timkpaine:
blocked by https://github.com/finos/perspective/pull/1355, will reimplement on top of those changes

texodus:
> blocked by #1355, will reimplement on top of those changes

Converted to Draft - let's use Draft PRs please for blocked or un-mergeable work so I know what needs to be reviewed.

timkpaine:
Oh yeah for sure I didn't realize you could go back to drafts

timkpaine:
@texodus redid this on top of your fixes and its working fine, ready for review

docs/md/python.md
```diff
@@ -481,6 +481,10 @@ PerspectiveWidget(table)
 PerspectiveWidget(None)
 ```
 
+## `PerspectiveRenderer`
+
+Perspective also exposes a JS-only `mimerender-extension`. This lets you view `csv`, `json`, and `arrow` files directly from the file browser. You can see this by right clicking one of these files and `Open With->CSVPerspective` (or `JSONPerspective` or `ArrowPerspective`). Perspetive will also install itself as the default handler for opening `.arrow` files.
```
texodus, line 486: Perspetive

packages/perspective-jupyterlab/src/ts/renderer.ts
```diff
@@ -84,13 +85,39 @@ export class PerspectiveDocumentWidget extends DocumentWidget<PerspectiveWidget>
...
+                        view.to_arrow().then((result: ArrayBuffer) => {
+                            const resultAsB64 = btoa(new Uint8Array(result).reduce((acc, i) => (acc += String.fromCharCode.apply(null, [i])), ""));
+                            this.context.model.fromString(resultAsB64);
+                            this.context.save();
+                        });
+                    } else if (this._type === "json") {
+                        view.to_json().then((result: any) => {
+                            this.context.model.fromJSON(result);
+                            this.context.save();
+                        });
```
texodus, line 111: Prefer `async` over mixing and matching `async` and `Promise` APIs, unless `Promise` is explicitly required (in this case make `on_update()` callback `async`).

packages/perspective-jupyterlab/src/ts/renderer.ts
```diff
@@ -84,13 +85,39 @@ export class PerspectiveDocumentWidget extends DocumentWidget<PerspectiveWidget>
...
+                view.on_update(() => {
+                    if (this._type === "csv") {
+                        view.to_csv().then((result: string) => {
+                            this.context.model.fromString(result);
+                            this.context.save();
+                        });
+                    } else if (this._type === "arrow") {
+                        view.to_arrow().then((result: ArrayBuffer) => {
+                            const resultAsB64 = btoa(new Uint8Array(result).reduce((acc, i) => (acc += String.fromCharCode.apply(null, [i])), ""));
+                            this.context.model.fromString(resultAsB64);
```
texodus, line 104: What does this do - this creates an arrow and not a base64 encoded string in a text file?

texodus (approved):
Thanks for the PR!  Looks good -

* Does the arrow output work?  On first sight it looks suspect, I will test this.
* It may be prohibitively expensive to save on every cell edit, the mime renderer may benefit form a small control panel with "update/reset" and "save/submit" buttons?

docs/md/python.md
texodus: Also is there a call-to-action here?  If this is just an ad for the functionality, maybe roll the paragraph into the `PerspectiveWidget` adjacent sibling?

packages/perspective-jupyterlab/src/ts/renderer.ts
timkpaine: Yep the key is in a different code block where we tell jupyter to open arrow files as binary which means they're exposed in the model as b64 encoded strings https://github.com/finos/perspective/blob/master/packages/perspective-jupyterlab/src/ts/renderer.ts#L181

docs/md/python.md
timkpaine: Mostly an ad, but its different from `PerspectiveWidget` since `PerspectiveWidget` is available in certain contexts (JupyterLab, Voilà, soon nteract and legacy Jupyter) where the Perspective Renderer is not (only available in JupyterLab).

packages/perspective-jupyterlab/src/ts/renderer.ts
timkpaine: done

PR #1354, opened 2021-03-24
Add Expressions Engine using ExprTk

sc1f:
This PR removes the old `computed_columns` API and replaces it with an `expressions` API that uses [ExprTk](https://github.com/ArashPartow/exprtk) for expression parsing and evaluation. This is the first step towards massively enhancing Perspective's functionality for cleaning, transforming, and augmenting data passed into it, and allows for a greater feature set than what was allowed for before.

The use of ExprTk, specifically, enables a _massive_ amount of new features, including:

- Scalars in expressions, which were previously column-to-column only: `"Sales" + (10 + 20 / sqrt(144))`
- Conditionals: `if ("State" == "Texas) 1; else 0;`
- Loops: `var x := 0; for (var i := 0; i < 10; i += 1) { x += i };`
- (Almost) the full range of mathematical and trigonometry functions offered by ExprTk
- Polymorphic functions such as `bucket()`, which combines the numeric and date bucket functions offered in the old API
- An `order(column, a, b, c...)` function, which allows custom sort orders on string columns

This fully overhauls the functionality of the previous API, which used a custom-written parser in Javascript and a clunky, object-based API, with a simple API that takes a list of expression strings, which is now unified across Javascript:

```javascript
const view = await table.view({
  expressions: ['"Sales" + "Profit" / 100', "if (\"State\" == 'Texas') 'okay!'; else 'hmm...'"]
})
```
And Python:
```python
view = table.view({
  expressions: ['"Sales" + "Profit" / 100', "if (\"Order Date\" == \"Ship Date\") 'shipped same day'; else 'not same day'"]
})
```

Examples of the new UI and expressions API in action:

sine_exprtk

errors and help msg

There is still some work left to do before this is ready for a new Perspective release, but merging this large change before the next release allows work to be more effectively parallelized between the additional UI and engine fixes that are required.

### Next steps

- [ ] Update engine semantics to fully store expression columns per-context instead of sharing the underlying `t_gstate` table
- [ ] Improve expression editor UI and move it out of the side panel
- [ ] Add various new, useful functions
- [ ] Rewrite documentation for new API


sc1f:
@ArashPartow ExprTk is an awesome library—I just wanted to say thank you for all the hard work you've put/continue to put into it. It's been awesome using it and integrating it with our API, and I'm excited to be bringing its breadth of functionality into our product.

nmichaud:
Interesting approach using t_scalar as the template arg for exprtk. I didn't think of that, instead typing directly as t_float64 and upcasting when calculating the computed columns in gnode->process(). Are you still thinking of implementing autocomplete on the javascript side?

nmichaud:
It would also be interesting to think about how to use exprtk for custom aggregates - could use the vector processing (section 14 of exprtk docs) and hook it up here - https://github.com/finos/perspective/blob/master/cpp/perspective/src/cpp/sparse_tree.cpp#L1144-L1147 using median as an prototype (which processes a vector of values representing all the values in that sparse tree node)

texodus:
@nmichaud We just started experimenting today with the idea of re-purposing ExprTK work to replace the existing filters.  A nice side effect of this is, without any other changes to the engine or UI, this very cleanly allows you to use filters both to _reduce_ an e.g. bar chart to aggregates satisfying a predicate (as is possible in current perspective), but also alternatively _paint_ the same predicate and chart using the boolean-returning ExprTK column as a column-pivot (_not_ currently possible through the UI or a single table).

We need to look into this further for aggregates - vector processing makes it easy to represent `getDepencencies()` itself via symbol injection, but only if its a numeric type since I do not believe we can replicate the `t_tscalar` template with primitives.  Current aggregate functions, at least, are free to change type between pivot depth levels e.g. `count()`.
In addition to polluting the DSL's API with type-dispatch, I think this may make type-checking somewhat bizarre as we'll need to recursively apply it to each level (e.g. something error messages that apply to specific recursion depths like `Type Error - Cannot use aggregate "concat_ints_to_string()" at pivot depth level 2`).

Regarding vector processing specifically, something I'm very excited about that we can likely do much sooner is implement hard-coded vector-returning functions within the ExprTK DSL itself, such as `window()`, which will allow trivial implementation of pseudo-stateful transforms like moving averages directly within the GUI.

nmichaud:
> but also alternatively _paint_ the same predicate and chart using the boolean-returning ExprTK column as a column-pivot (_not_ currently possible through the UI or a single table).

Not sure I follow what you mean by "paint[ing] the same predicate and chart".

> We need to look into this further for aggregates - vector processing makes it easy to represent `getDepencencies()` itself via symbol injection, but only if its a numeric type since I do not believe we can replicate the `t_tscalar` template with primitives. Current aggregate functions, at least, are free to change type between pivot depth levels e.g. `count()`.
> In addition to polluting the DSL's API with type-dispatch, I think this may make type-checking somewhat bizarre as we'll need to recursively apply it to each level (e.g. something error messages that apply to specific recursion depths like `Type Error - Cannot use aggregate "int_to_string()" at pivot depth level 2, as pivot depth level 1 converted this depth to a string`).

That's true - the exprtk function would be a bit contorted to support returning multiple types - https://github.com/ArashPartow/exprtk/blob/master/readme.txt#L3580-L3591, but it is possible.

> Regarding vector processing specifically, something I'm very excited about that we can likely do much sooner is implement hard-coded vector-returning functions within the ExprTK DSL itself, such as `window()`, which will allow trivial implementation of pseudo-stateful transforms like moving averages directly within the GUI.

Ah yeah, that's a great use case.

sc1f:
@nmichaud this is a good example of "paint[ing] the same predicate and chart":

Screen Shot 2021-03-24 at 10 45 27 PM

Applying a `discount > 0` filter will remove the rows for which the filter returns false, but using the expression here allows us to see `discount > 0` over the totality of the data, which will be useful in a lot of cases. 

In terms of autocomplete - column name autocompletion will be easy to implement, but since this expression language is far more complicated than our previous one, I think we'll need to look at Monaco or some other more managed text editor solution instead of trying to hand-roll our own.

nmichaud:
@sc1f Got it, thanks. I misunderstood what @texodus was saying about using it for filtering. It's even more powerful if you can use it to 'project' selections across different views of the same table: https://twitter.com/naveenma/status/804137377186856963

ArashPartow:
Looks like an awesome use of the library. Do let me know if there's anything I can help with.

nmichaud:
> Looks like an awesome use of the library. Do let me know if there's anything I can help with.

Hi @ArashPartow, is there a description of the Exprtk grammar anywhere (say in ANTLR form)?

ArashPartow:
@nmichaud 

Nothing in BNF per se. In short it's infix expressions from which statements can be constructed. There's some more details about the language structure in SECTION 29 of the readme.

https://github.com/ArashPartow/exprtk/blob/master/readme.txt#L4696

https://github.com/ArashPartow/exprtk/blob/master/readme.txt#L4278

sc1f:
expressions

Added a few new features to the editor - the help panel displays all operators, functions, and control flow operators so that the syntax is clear and discoverable, and error messages are now forwarded from the engine.

nmichaud:
One thing I noticed - the implementation builds the symbol table and reparses the computed expressions on each gnode update cycle (https://github.com/finos/perspective/pull/1354/files#diff-ecb1d4be36447785a2e972405912b5c20743a0502a2527f23bd9f8b5613f6a38R218-R266). While parsing is fast, it's duplicate work which can be done once on definition of the expression and stored on the t_computed_expression. Also once the expression is parsed, it's possible to track all dependent expression variables (see section 16 - https://github.com/ArashPartow/exprtk/blob/master/readme.txt#L2493) and only recompute if any dependent columns are in the flattened update table.

texodus (approved):
There is a lot more work to be done on this feature, but it's already a massive improvement over what's in `v0.8.x` and quite stable;  what remains is mostly UX.  ExprTK is the sole focus of `v0.9.0`, and the next few PRs will refine the integration before this release.

Thanks @nmichaud and @ArashPartow for your feedback, and thanks @sc1f for this _monumental_ PR (nearly 30k lines)!

PR #1371, opened 2021-03-30
Updating example webpack config

f0ster:


texodus:
Thanks for the PR!

I appreciate the fix, but there are a few errors here (dependencies overridden by root, dependencies in wrong section, hard-coded CDN urls, example warnings), and has conflicts, and only the change from `webpack=dev-server` to `webpack serve` are relevent.  In the future though, I can merge fixes much easier if they are focused on a single issue and don't come with other changes I must understand to validate a PR!

f0ster:
@texodus  Thanks for your feedback.
  
Upon encountering the defunct react example, I added the additional node modules required to build and run perspective that were not present in the package manifest. 

The reference to the superstore arrow source file also does not work, hence it the CDN http resource used in the commit.   

Perhaps this merits distinct issues for each case.  The impetus and sum of changes in this PR is to fix the react example and make it functional.

PR #1374, opened 2021-04-02
Remove python2 build support

texodus:
Refactored from #1336.

Deprecates [Python2](https://www.youtube.com/watch?v=IVXJmfd3cmg).  While official support will be removed in `v0.7.0` already, this PR removes the build script and library specializations that remain for supporting python2 builds.


sc1f:
We can remove `six` as part of this PR as well - makes the code a little cleaner.

PR #1375, opened 2021-04-05
Add `call_loop` and `get_table_names`

sc1f:
This PR adds two methods to the `PerspectiveManager` API:

`call_loop(fn, *args, **kwargs)`: calls `fn` on the `loop_callback` of the manager, or raises a `PerspectiveError` if the loop callback is not set.

This allows users to make calls on the Perspective thread without having to deal with the `_loop_callback` property of the manager, which is ostensibly private but (until now) lacked a public API.

`get_table_names()`: Returns a list of the tables hosted on the manager by name.

A common use case is to query the managers' private `_tables` and `_views` dictionaries for users to find out which Tables/Views are hosted on a server, which is especially useful in a server-client context where the consumers of the Perspective server may be different than the author of the server. `get_table_names()` offers a public API that can be called without having to query private properties of `PerspectiveManager`.

### DIFF
--- a/docs/md/python.md
+++ b/docs/md/python.md
@@ -247,10 +247,10 @@ using a websocket API.
 By default, `perspective` will run with a synchronous interface.  Using the
 `PerspectiveManager.set_loop_callback()` method, `perspective` can be configured
 to defer the application of side-effectful calls like `update()` to an event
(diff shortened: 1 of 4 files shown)


texodus (approved):
Looks good!

texodus:
Thanks for the PR!

PR #1355, opened 2021-03-25
Prebuilt JupyterLab extension / Legacy Notebook extension

timkpaine:

PR #1377, opened 2021-04-11
Fix `@finos/perspective-jupyterlab` to work with WebAssembly/Webpack5

texodus:
This was a nightmare and I don't want to talk about it.


timkpaine:
hahahahaha

timkpaine:
I swear I tried my best in https://github.com/finos/perspective/pull/1355

packages/perspective-jupyterlab/src/ts/renderer.ts
```diff
@@ -82,19 +80,14 @@ export class PerspectiveDocumentWidget extends DocumentWidget<PerspectiveWidget>
...
-                        this._psp.viewer.load(this._table);
-                    } else {
-                        // replace existing table for whatever reason
-                        this._psp.replace(this._table);
-                    }
-                });
+
+            if (this._psp.viewer.table === undefined) {
+                // construct new table
+                this._psp.viewer.load(perspective.worker().table(data));
```
sc1f, line 86: This will create a new WebWorker for each widget - the worker should be a property on the DocumentWidget.

rust/perspective-vieux/package.json
```diff
@@ -17,7 +17,7 @@
         "src/**/*"
     ],
     "scripts": {
-        "build:wasm": "wasm-pack build --debug",
+        "build:wasm": "wasm-pack build --target web",
```
sc1f, line 20: We should explain this in the docs - end-users who import Perspective into their own Webpack builds will need to add the `arraybuffer-loader` in order to correctly load perspective.

sc1f (changes requested):

sc1f (approved):
Looks good!

PR #1382, opened 2021-04-14
Adds `yarn repl` to launch a shell inside our docker images

sc1f:
This PR adds `yarn repl`, which will launch a bash shell inside our docker images for better debugging of our Python build when trying to build Linux builds/wheels locally. `yarn repl` can be used with the same options to `yarn build` and `yarn _wheel_python`, i.e. `--manylinux2010`, `--manylinux2014`, etc.


texodus:

texodus (approved):

PR #1384, opened 2021-04-19
Backwards compatibility for table() and view()

sc1f:
#1289 was a breaking change for the Perspective API, as it made Table and View constructors `async` by returning a `Promise` to the Table or View instance instead of the instance itself. Internally, the constructors would wait on the server (across the websocket or a web worker) to return a message that signaled successful Table/View creation. However, when running a Perspective client that is newer than the Perspective server, older versions of the server (i.e. all released versions right now, as the async table/view changes are not in a tagged release yet) will not return the acknowledgement message and thus Table and View constructors will block forever.

To maintain backwards compatibility, this PR adds a simple feature flag implementation on both the client and server, which allows newer versions of Perspective to run against older versions on the server. In the `init` message used to begin client/server communication, the server returns an array of strings that represent enabled features, and the client will use those flags to determine its behaviour.

Currently, only one flag (`wait_for_response`) is implemented - if the flag is present, then the Table and View constructors will wait for an acknowledgement message from the server. If the flag is not present, or if there are no feature flags at all, then the constructors will `resolve()` immediately, thus preserving compatible behaviour between different versions.

### DIFF
--- a/packages/perspective/src/js/api/client.js
+++ b/packages/perspective/src/js/api/client.js
@@ -44,7 +44,7 @@ export class Client {
     }
 
     /**
-     * Process an asynchronous message.
+     * Send a message to the server.
      */
     post(msg, resolve, reject, keep_alive = false) {
         ++this._worker.msg_id;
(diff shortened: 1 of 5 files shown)


texodus (approved):

Issue #1373, opened 2021-03-31
Expose `join` aggregate function to clients


f0ster:
### Description of Problem:
When transforming tables/creating views, a join or list operator to apply to the underlying grouped sets would be helpful for debugging. 

### Potential Solutions:
Support `join` operator on client side table aggregate views.

hemantbaxi:
Hi,

I am interested to give a try on this.

My initial understanding after looking at code is that we need something like SQL join operator. Is that right ?  I am wondering on how is it different from pandas.dataframe.join ? Can that be directly used ?

Secondly,  in problem statement it is mentioned as "join" or "list". I think the two are different operations. Can you please elaborate on the issue, if possible, with an example ?

Thanks,
Hemant

f0ster:
Thanks @hemantbaxi , I added a PR with a possible solution above.  We could have better test cases and perhaps break out the aggregate cases into their own file, cc @texodus 

The idea is to have an aggregate function (join), which takes a list, and returns one element (the list joined/concatenated). 
E.g.,  a simple pythonic join

```
>>> list("cheers")
['c', 'h', 'e', 'e', 'r', 's']
>>> "".join(list("cheers"))
'cheers'
>>> 
```

hemantbaxi:
Thanks Ryan,

I looked at the PR.  If I understood it right, I can apply the two commits on my dev branch. The tests will fail initially, but once I add the python code for "join", these tests should pass.. Right ?  

Thanks,
Hemant

f0ster:
@hemantbaxi No problem.  

The new specs are for the javascript library, but the core aggregate function is exposed to python.

Can you verify the existing tests pass as expected in your local environment , before applying and testing the new commits?

hemantbaxi:
> @hemantbaxi No problem.
> 
> The new specs are for the javascript library, but the core aggregate function is exposed to python.
> 
> Can you verify the existing tests pass as expected in your local environment , before applying and testing the new commits?

I assume that since we can setup perspective in two environments Javascript/Python, each has to build and tested separately. I am able to build successfully and pass all the tests when I run in Python setup. But for Javascript setup, it is failing to build itself.

In Javascript setup, it seems even though @babel/preset-env plugin is installed, webpack is not able to recognize optional-chaining and is giving the below syntax error:

```
 babel src/js --source-maps --out-dir dist/esm
SyntaxError: src/js/api/client.js: Unexpected token (61:83)
  59 |                 this.send(msg);
  60 | 
[...]

PR #1394, opened 2021-04-25
expose join aggregate function for clients #1373

f0ster:
For https://github.com/finos/perspective/issues/1373

### DIFF
--- a/packages/perspective/src/js/config/constants.js
+++ b/packages/perspective/src/js/config/constants.js
@@ -33,6 +33,7 @@ const NUMBER_AGGREGATES = [
     "last by index",
     "last",
     "high",
+    "join",
     "low",
     "mean",
     "median",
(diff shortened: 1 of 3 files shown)


texodus (approved):
Thanks for the PR!  This is great - and a test!

Playing around with this a bit, I recall why we did not enable this aggregate type - it can cause havoc on columns with many distinct string values.  I propose modifying this aggregation to limit the string length to ~280 characters (tweet length), by short circuiting the iteration over the `std::set` which accumulates the unique join-ed values [here](https://github.com/finos/perspective/blob/master/cpp/perspective/src/cpp/sparse_tree.cpp#L1020) when this character limit is reached.  `std::set` is ordered so this should still create deterministic results.

f0ster:
@texodus Thanks for the feedback! 

How does this solution approach look to limit the set space?  

https://github.com/f0ster/perspective/commit/10bcb9f7cc8d257189d8a4c7464e1f479eb2d565

texodus:
Thanks for the update!

I've made a fork of this PR in #1434 which includes these commits, plus a `c_str` limiting calculation, so I'm going to close this PR in favor of that one.  The gist of the change is to avoid allocation of `std::string` per row calculation;  the `t_tscalar` type for strings does not use `std::string`, but a dictionary-encoded character array which saves us calls to malloc in performance sensitive tight loops.  Rather than using a `std::string` and `size()` to calculate this, we just count the characters in the underlying c_str using the start and end indices (which is basically free in comparison).

PR #1395, opened 2021-04-26
Add `get_min_max()` to Perspective API

texodus:
Adds `get_min_max()` function to `View` objects in Python and Javascript, the purpose of which is to provide reasonable value ranges for rendering gradient-maps.  `get_min_max()` takes a column name as an argument, and returns a tuple `[min, max]` of this column's `View`-type.  This new version of min/max calculation works unlike the previous incarnation:
* It is a one-time-query, and does not recalculate unless called.  The new `<perspective-column-style>` component calls this once only on instantiation.
* It does not include parent aggregates, e.g. anything with `row_path` or `column_path` less than their respective maxes, unless ...
* ... in the specific case of Row Pivots, when all children of a row depth are in a `"collapsed"` state, there are no non-parent aggregates, so the next non-empty depth level is used instead.


texodus:
As `@finos/perspective-viewer` now depends on the presence of this call to open the new Column Style menu feature, and thus depends on a `@finos/perspective` or `perspective-python` version which supports this call, this feature should eb `breaking`

Issue #1409, opened 2021-05-02
Cannot find perspective_vieux_bg.wasm


genderev:
### Steps to Reproduce:
 1. I simply followed the example in the [JavaScript User Guide](https://perspective.finos.org/docs/md/js.html).
 2. I saved a simple `index.html` file with the following content:
```
<!DOCTYPE html>
<html lang="en" dir="ltr">
  <head>
    <meta charset="utf-8">
    <title>Perspective Demo</title>
    <script src="https://unpkg.com/@finos/perspective"></script>
    <script src="https://unpkg.com/@finos/perspective-viewer"></script>
    <script src="https://unpkg.com/@finos/perspective-viewer-datagrid"></script>
    <script src="https://unpkg.com/@finos/perspective-viewer-d3fc"></script>
  </head>
  <body>
<script type="text/javascript">
async function main(){
  const worker = perspective.worker();
  const table = await worker.table({A: [1, 2, 3]});
  const view = await table.view({sort: [["A", "desc"]]});
}
main();
</script>
  </body>
</html>
```
 3. I opened this HTML file in a browser that I knew had WebAssembly support.

### Expected Result:
I expected to see a simple table.

### Actual Result:
I received these errors in Chromium:
`unpkg.com/@finos/perspective_vieux_bg.wasm:1`
`Failed to load resource: the server responded with a status of 404 ()`

`index.html:1` 
`Uncaught (in promise) CompileError: WebAssembly.instantiate(): expected magic word 00 61 73 6d, found 43 61 6e 6e @+0`

`unpkg.com/@finos/perspective.cpp.wasm:1` 
`Failed to load resource: the server responded with a status of 404 ()`

### Environment:
OS X (Catalina), using CDN to load Perspective. I don't think my setup is special, just testing out a demo.

### Additional Context:
I tried this demo on a different computer yesterday (also OS X Catalina with Chromium browser). The browser could not find `perspective_vieux_bg.wasm`.

texodus:
The documentation is out-of-date - use the CDN fully-qualified paths as in the [`bl.ocks` examples](https://bl.ocks.org/texodus/45b868833c9f456bd39a51e606412c5d).

```html
<script src="https://unpkg.com/@finos/perspective/dist/umd/perspective.js"></script>
<script src="https://unpkg.com/@finos/perspective-viewer/dist/umd/perspective-viewer.js"></script>
[...]

PR #1410, opened 2021-05-03
Install Boost from JFrog, fix outdated docs from #1409

sc1f:
As of May 1, Bintray has been [sunset](https://jfrog.com/blog/into-the-sunset-bintray-jcenter-gocenter-and-chartcenter/) which causes our `install_tools.js` to fail when trying to install Boost.

This PR removes all Bintray links and replaces them with a link to the Boost artefact in JFrog, from the link provided on the Boost [website](https://www.boost.org/users/history/version_1_71_0.html). Dockerfiles have also been updated to use the JFrog link to prevent any build issues.

Additionally, it fixes the outdated docs mentioned in #1409 by removing Unpkg alias links and replacing them with direct links to the /dist/umd folder.

### DIFF
--- a/docker/docs/Dockerfile
+++ b/docker/docs/Dockerfile
@@ -17,7 +17,7 @@ RUN python3.7 -m pip install numpy pandas
 
 RUN ls -al /usr/local/lib/python3.7/site-packages/numpy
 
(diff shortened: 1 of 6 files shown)


texodus (approved):

PR #1412, opened 2021-05-03
Fix memory leak(s), leak tests, `memory_usage()` wasm heap API

texodus:


packages/perspective/src/js/api/client.js
```diff
@@ -65,6 +65,23 @@ export class Client {
         }
     }
 
+    async memory_usage() {
+        return await new Promise((resolve, reject) => {
+            ++this._worker.msg_id;
```
sc1f, line 70: Seems like the incrementing msg_id, setting a handler etc. logic here is duplicated from `post` - seems like you could probably just call `post({cmd: "memory_usage"}, resolve, reject)` here

PR #1421, opened 2021-05-14
Improve expressions engine

sc1f:
### Changelog

- Multiple views can now create expressions using the same alias, but of different output types.
  * Previously, expression columns were stored on the tables provided by the `gnode` using their alias, so multiple views could overwrite each other's columns as long as they were the same type. 
  * Now, expression columns are stored using the full expression string as typed by the user, and the `View` keeps track of a mapping of expression aliases to expression strings. During view construction, references to an expression alias are removed and replaced with the expression string, and whenever a view API method is called, expression strings are replaced with the expression alias before the result is returned. Thus, multiple views can share the same alias and create columns of arbitrary types using the same alias.
- Small UI fixes to allow for replacing expressions - when an expression is replaced with one of a different type, it is `reset` and removed from columns, filters, sorts etc. - this is a stopgap measure and should be replaced wholly with a better UI for editing and creating expression columns.

### DIFF
--- a/cpp/perspective/src/cpp/computed_expression.cpp
+++ b/cpp/perspective/src/cpp/computed_expression.cpp
@@ -170,23 +170,10 @@ t_computed_expression::compute(
         PSP_COMPLAIN_AND_ABORT(ss.str());
     }
 
-    // create or get output column using m_expression_alias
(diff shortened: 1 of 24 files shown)


sc1f:
As discussed offline with @texodus, needs a few more changes - closing for now.

PR #1431, opened 2021-05-26
Isolate expressions per-context and ensure memory stability

sc1f:
This PR refactors the way that expressions are handled and stored in Perspective - previously, expressions columns were calculated and stored on the main table inside `t_gstate` and its intermediate tables (prev, current, delta, transitions, etc.). By having all expressions on the gnode's tables, expressions were "real" columns to the engine, and it guaranteed that all the features available to real columns (pivot, filter, etc.) were available to expression columns without having to do large refactors to the code that generated aggregates, pivots etc.

However, this meant that expressions were not "isolated" to the `View` they were created in - a user could not create a new expression named `new column` of `float` type, and then replace it immediately with an expression that returns a `string`, as the main table schema had already recorded `new column` as a `float` column and could not overwrite it.

Now, expression columns are stored _per-context_ on data tables that only contain the expression columns belonging to the `View`. This means that each `View` can create expression columns that use the same aliases as other `Views`, and they can be of any type:

```javascript
const v1 = await table.view({
  expressions: [`// new column\n "ask" - "bid"`]
});

const v2 = await table.view({
  expressions: [`// new column\n upper("client")`]
});
```
And now, expression columns are guaranteed to be deleted when `view.delete()` is called - when the context is deleted, the data tables that contain their expressions are cleaned up from the heap.

To implement this, I added a `join` function to `t_data_table` - given two data tables, it creates a new data table that references the columns from the source tables using their shared pointers without copying any of the underlying memory or allocating new data stores on the heap.  When the context receives data from the gnode during `notify`, it is provided a `t_data_table` that contains the "real" columns _joined_ with the expression columns from the context being notified. This means that the context will continue to treat expression columns as "real" columns so there is no loss of functionality.

Some work remains on the UI to allow for duplicate expression aliases to be created - right now, `perspective-viewer` does not allow aliases to be reused. Because #1426 changes the expressions UI "bigly", this PR does not contain any UI changes, and those changes will be added on in a new PR before the Perspective 0.9.0 release.

### Changelog

- Store expressions per-context using the `t_expression_tables` struct
- `t_vocab` for string expressions are now per-context, instead of a static global vocab for all expressions
- Added tests using multiple views with identical expression aliases
- Added tests to assert that there are no memory leaks or excess allocations when views with expressions are created and deleted
- Updated the `View` documentation to reflect the expressions API

### DIFF
--- a/cpp/perspective/CMakeLists.txt
+++ b/cpp/perspective/CMakeLists.txt
@@ -250,6 +250,7 @@ if (PSP_WASM_BUILD)
 			-s DISABLE_EXCEPTION_CATCHING=0 \
 			-s ASSERTIONS=2 \
 			-s DEMANGLE_SUPPORT=1 \
+			-s SAFE_HEAP=1 \
 			")
 	else()
 		set(OPT_FLAGS " \
--- a/cpp/perspective/CMakeLists.txt
+++ b/cpp/perspective/CMakeLists.txt
@@ -501,6 +501,7 @@ set (SOURCE_FILES
 	${PSP_CPP_SRC}/src/cpp/dense_tree_context.cpp
 	${PSP_CPP_SRC}/src/cpp/dense_tree.cpp
 	${PSP_CPP_SRC}/src/cpp/dependency.cpp
+	${PSP_CPP_SRC}/src/cpp/expression_tables.cpp
 	${PSP_CPP_SRC}/src/cpp/extract_aggregate.cpp
 	${PSP_CPP_SRC}/src/cpp/filter.cpp
 	${PSP_CPP_SRC}/src/cpp/flat_traversal.cpp
--- a/cpp/perspective/src/cpp/computed_expression.cpp
+++ b/cpp/perspective/src/cpp/computed_expression.cpp
@@ -128,8 +128,8 @@ t_computed_expression::t_computed_expression(
 
 void
 t_computed_expression::compute(
-    t_data_table* source_table,
-    t_data_table* destination_table,
+    std::shared_ptr<t_data_table> source_table,
+    std::shared_ptr<t_data_table> destination_table,
     std::shared_ptr<t_vocab> vocab) const {
     // TODO: share symtables across pre/re/compute
     exprtk::symbol_table<t_tscalar> sym_table;
--- a/cpp/perspective/src/cpp/computed_expression.cpp
+++ b/cpp/perspective/src/cpp/computed_expression.cpp
@@ -62,17 +62,15 @@ t_computed_expression_parser::LOWER_VALIDATOR_FN = computed_function::lower(null
 computed_function::length
 t_computed_expression_parser::LENGTH_VALIDATOR_FN = computed_function::length(nullptr);
 
-// Register functions in a centralized macro instead of copy-pasting the same
-// lines between compute/recompute
 #define REGISTER_COMPUTE_FUNCTIONS()                                                        \
-    computed_function::day_of_week day_of_week_fn = computed_function::day_of_week(m_expression_vocab);         \
-    computed_function::month_of_year month_of_year_fn = computed_function::month_of_year(m_expression_vocab);   \
-    computed_function::intern intern_fn = computed_function::intern(m_expression_vocab);    \
-    computed_function::concat concat_fn = computed_function::concat(m_expression_vocab);    \
-    computed_function::order order_fn = computed_function::order(m_expression_vocab);       \
-    computed_function::upper upper_fn = computed_function::upper(m_expression_vocab);       \
-    computed_function::lower lower_fn = computed_function::lower(m_expression_vocab);       \
-    computed_function::length length_fn = computed_function::length(m_expression_vocab);    \
+    computed_function::day_of_week day_of_week_fn = computed_function::day_of_week(vocab);         \
+    computed_function::month_of_year month_of_year_fn = computed_function::month_of_year(vocab);   \
+    computed_function::intern intern_fn = computed_function::intern(vocab);    \
+    computed_function::concat concat_fn = computed_function::concat(vocab);    \
+    computed_function::order order_fn = computed_function::order(vocab);       \
+    computed_function::upper upper_fn = computed_function::upper(vocab);       \
+    computed_function::lower lower_fn = computed_function::lower(vocab);       \
+    computed_function::length length_fn = computed_function::length(vocab);    \
     sym_table.add_function("today", computed_function::today);                              \
     sym_table.add_function("now", computed_function::now);                                  \
     sym_table.add_function("bucket", t_computed_expression_parser::BUCKET_FN);              \
@@ -130,7 +128,9 @@ t_computed_expression::t_computed_expression(
 
 void
 t_computed_expression::compute(
-    std::shared_ptr<t_data_table> data_table) const {
+    t_data_table* source_table,
+    t_data_table* destination_table,
+    std::shared_ptr<t_vocab> vocab) const {
     // TODO: share symtables across pre/re/compute
     exprtk::symbol_table<t_tscalar> sym_table;
     sym_table.add_constants();
@@ -148,7 +148,7 @@ t_computed_expression::compute(
     for (t_uindex cidx = 0; cidx < num_input_columns; ++cidx) {
         const std::string& column_id = m_column_ids[cidx].first;
         const std::string& column_name = m_column_ids[cidx].second;
-        columns[column_id] = data_table->get_column(column_name);
+        columns[column_id] = source_table->get_column(column_name);
 
         t_tscalar rval;
         rval.m_type = columns[column_id]->get_dtype();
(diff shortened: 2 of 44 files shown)


cpp/perspective/CMakeLists.txt
```diff
@@ -250,6 +250,7 @@ if (PSP_WASM_BUILD)
 			-s DISABLE_EXCEPTION_CATCHING=0 \
 			-s ASSERTIONS=2 \
 			-s DEMANGLE_SUPPORT=1 \
+			-s SAFE_HEAP=1 \
```
texodus, line 253: IIRC this setting basically makes debug-build Perspective unusable (though it is quite useful for finding memory leaks!)

cpp/perspective/src/cpp/computed_expression.cpp
```diff
@@ -62,17 +62,15 @@ t_computed_expression_parser::LOWER_VALIDATOR_FN = computed_function::lower(null
...
-    computed_function::lower lower_fn = computed_function::lower(m_expression_vocab);       \
-    computed_function::length length_fn = computed_function::length(m_expression_vocab);    \
+    computed_function::day_of_week day_of_week_fn = computed_function::day_of_week(vocab);         \
+    computed_function::month_of_year month_of_year_fn = computed_function::month_of_year(vocab);   \
+    computed_function::intern intern_fn = computed_function::intern(vocab);    \
+    computed_function::concat concat_fn = computed_function::concat(vocab);    \
+    computed_function::order order_fn = computed_function::order(vocab);       \
+    computed_function::upper upper_fn = computed_function::upper(vocab);       \
+    computed_function::lower lower_fn = computed_function::lower(vocab);       \
+    computed_function::length length_fn = computed_function::length(vocab);    \
```
texodus, line 73: Any reason this needs to be a macro and not a function?  The scope-capture of `vocab` specifically is something I'd like to stylistically avoid if possible. (Looks also like `recompute` is deprecated?)

cpp/perspective/src/cpp/context_grouped_pkey.cpp
```diff
@@ -42,9 +42,25 @@ t_ctx_grouped_pkey::init() {
     m_tree = std::make_shared<t_stree>(pivots, m_config.get_aggregates(), m_schema, m_config);
     m_tree->init();
     m_traversal = std::shared_ptr<t_traversal>(new t_traversal(m_tree));
+
```
texodus, line 45: Unrelated to this PR but - we should not compile this as there is no public API (nor tests ...)

cpp/perspective/src/cpp/context_zero.cpp
```diff
@@ -602,6 +595,106 @@ t_ctx0::get_step_delta(t_index bidx, t_index eidx) {
...
+t_ctx0::get_expression_tables() const {
+    return m_expression_tables;
+}
+
+void
+t_ctx0::read_column_from_gstate(
+    const std::string& colname,
+    const std::vector<t_tscalar>& pkeys,
+    std::vector<t_tscalar>& out_data) const {
+    std::shared_ptr<t_data_table> master_table = m_gstate->get_table();
```
texodus, line 681: This can go within the `else` block

cpp/perspective/src/cpp/data_table.cpp
```diff
@@ -607,6 +612,52 @@ t_data_table::borrow(const std::vector<std::string>& columns) const {
...
+    }
+
+    // join both schemas
+    t_schema schema = m_schema;
+    const t_schema& other_schema = other_table->get_schema();
+    std::vector<std::string> other_columns;
+
+    for (const std::string& column : other_schema.m_columns) {
+        // Only append unique columns
+        if (!schema.has_column(column)) {
```
texodus, line 635: The alternative to this seems an error ...

cpp/perspective/src/include/perspective/gnode.h
```diff
@@ -369,8 +356,9 @@ class PERSPECTIVE_EXPORT t_gnode {
     // Output ports stored sequentially in a vector, keyed by the
     // `t_gnode_port` enum.
     std::vector<std::shared_ptr<t_port>> m_oports;
-    std::map<std::string, t_ctx_handle> m_contexts;
+    tsl::ordered_map<std::string, t_ctx_handle> m_contexts;
```
texodus, line 359: ?? `unordered_map` probably?

packages/perspective/src/js/perspective.js
```diff
@@ -1344,7 +1328,18 @@ export default function(Module) {
                 return `${match.substr(0, match.indexOf(full))}'${value}')`;
             });
 
-            validated_expressions.push([expression_alias, expression_string, parsed_expression_string, column_id_map]);
+            const validated = [expression_alias, expression_string, parsed_expression_string, column_id_map];
+
+            // Check if this expression is already in the array, if so then
+            // we need to replace the expression so the last expression tagged
+            // with the alias is the one that is applied to the engine.
+            if (expression_idx_map[expression_alias] !== undefined) {
```
texodus, line 1336: This is impossible without a malformed `config`, no?

python/perspective/perspective/table/_utils.py
```diff
@@ -160,22 +159,15 @@ def _parse_expression_strings(expressions):
 
         alias_match = re.match(ALIAS_REGEX, expression)
```
texodus, line 160: Note to self: `ALIAS_REGEX` is wrong (corrected in JS but not python ..)

packages/perspective/test/js/leaks.spec.js
```diff
@@ -71,4 +97,77 @@ describe("leaks", function() {
...
+
+            await leak_test(async () => {
+                const view = await table.view({
+                    row_pivots: [columns[Math.floor(Math.random() * columns.length)]],
+                    expressions: [expressions[Math.floor(Math.random() * expressions.length)]]
+                });
+                const expression_schema = await view.expression_schema();
+                expect(Object.keys(expression_schema).length).toEqual(1);
+                await view.delete();
+            }, 3000);
```
texodus, line 143: These looks correct on-the-face, but why 3,000?  In theory, these test should run identically with 1 or 1,000,000 - iteration checks are just there for offensively slow operations.

scripts/test_python.js
```diff
@@ -41,11 +41,11 @@ const pytest_client_mode = IS_DOCKER => {
     if (IS_DOCKER) {
         return bash`${docker(IMAGE)} bash -c "cd \
             python/perspective && TZ=UTC ${PYTHON} -m pytest \
-            ${VERBOSE ? "-vv" : ""} --noconftest 
+            ${VERBOSE ? "-vv --full-trace" : ""} --noconftest
```
texodus, line 44: Is this necessary for typical builds or is this debug cruft?  While I'm sure the info is useful in some cases, long stdout make Azure cry.

cpp/perspective/src/include/perspective/gnode.h
```diff
@@ -390,44 +378,56 @@ class PERSPECTIVE_EXPORT t_gnode {
...
+        // Join expression tables on the context with gnode tables and pass those
+        // into the context so there is no distinction between expression and real
+        // columns for the context.
+        std::shared_ptr<t_expression_tables> ctx_expression_tables = ctx->get_expression_tables();
+
+        auto joined_flattened = flattened->join(ctx_expression_tables->m_flattened);
+        auto joined_delta = delta->join(ctx_expression_tables->m_delta);
+        auto joined_prev = prev->join(ctx_expression_tables->m_prev);
+        auto joined_current = current->join(ctx_expression_tables->m_current);
+        auto joined_transitions = transitions->join(ctx_expression_tables->m_transitions);
```
texodus, line 409: Can do this `join` once per context?  The column sets do not change per `gnode.process()` call.

texodus (approved):
Looks good!  Check my review notes for a some debug cruft leftover.

There is a mostly-trivial performance regression here of about 5%, which is IMO acceptable for the massive improvement in code organization, test coverage, inline comments and general good engineering.  We discussed `t_data_table` construction costs at length offline, and I do not suspect these are the cause of this slowdown - but IMO we should measure the impact of this anyway and more generally make some project-wide rules about `shared_ptr` vs. const ref vs. `std::move` just so we're consistent (as this is completely ad-hoc right now :) ).

Screen Shot 2021-05-28 at 9 35 02 PM

cpp/perspective/src/cpp/computed_expression.cpp
sc1f: I don't think this macro additionally captures the vocab - it is basically just inlining this code into where the macro is invoked, and the macro doesn't take `vocab` as a parameter. AFAIK since it's just a text replacement macro, it shouldn't capture `vocab` or have any notion of what `vocab` is during the preprocessor step. I think leaving it as a macro works well for readability in the `compute` function.

scripts/test_python.js
sc1f: This is only for `--debug` - it should have no effect on regular builds or Azure.

packages/perspective/src/js/perspective.js
sc1f: This is possible in the current UI, if one creates an expression column and then replaces it with another one - the viewer API will pass both expressions sequentially: `["// a \n 123", "// a\n upper('abcde')"]`. With all the expression UI/fix PRs merged we can fix this on the viewer side.

cpp/perspective/src/cpp/computed_expression.cpp
texodus: [That's what macro variable capture means](https://en.wikipedia.org/wiki/Hygienic_macro#The_hygiene_problem).  There's little risk of `vocab` being mis-used here as you mention the macro won't compile, but for complexity analysis purposes I think its best if we avoid variable capture (in this case with a macro argument) and try where possible to force scope (with `{}`) in our macros.

PR #1432, opened 2021-05-27
Parse aggregates in column order

sc1f:
This PR fixes a bug where column names that could be parsed as numbers (like the expression `1234`) did not respond correctly to column ordering and would be the first column in a pivoted view:

Screen Shot 2021-05-27 at 3 03 32 PM 2

This was due to the way that aggregates were parsed in the C++, which happened in three separate iterations:

1. Iterate through the `columns` array, and generate default aggregates for all columns _not_ in the aggregate map
2. Iterate through the `aggregate` map (using the order generated by `Object.keys()`), and apply specified aggregates for all aggregate columns _in_ the columns array
3. Generate aggregates for hidden sort columns.

I'm not sure exactly _why_ I wrote the parsing logic the way that I did (in 2019, IIRC), but the issue it causes is simple: by ignore the column order when parsing the aggregates map, it assumes that the aggregates map can be iterated in the same order as columns.

However, in Javascript, in an object with a `string` key that is parsable as an integer (like "1234"), when iterating through the object using `Object.keys()`, the integer-parsable key is automatically moved to the front of the iteration order, and multiple integer-parsable keys are ordered in ascending order. More details [here](https://stackoverflow.com/a/23202095).

An example of this behavior is as follows:

Screen Shot 2021-05-27 at 10 32 28 PM

This behavior does not seem to happen when the keys are parsable as floats ("4.5" and "5.55555", for example, remain in insertion order). Thus, the assumption that the aggregates map is ordered properly is broken, and so goes the column order. The fix is to rewrite the aggregate parsing logic to use the column order while iterating, and access the aggregates map through the column order instead of the order given by `Object.keys()`.

I also removed the unimplemented `and` and `or` filters for boolean columns from the UI - though boolean columns are less common, selecting those would have caused an `abort()`.

### DIFF
--- a/cpp/perspective/src/cpp/view_config.cpp
+++ b/cpp/perspective/src/cpp/view_config.cpp
@@ -200,63 +200,32 @@ t_view_config::get_column_pivot_depth() const {
 // PRIVATE
 void
 t_view_config::fill_aggspecs(std::shared_ptr<t_schema> schema) {
-    /*
-     * Provide aggregates for columns that are shown but NOT specified in 
-     * `m_aggregates`, including expressions that are in the `columns`
-     * array but not the `aggregates` map.
-     */
+    auto max_agg_count = m_columns.size() + m_sort.size();
+    m_aggspecs.reserve(max_agg_count);
+    m_aggregate_names.reserve(max_agg_count);
+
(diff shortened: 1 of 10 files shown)


texodus (approved):

PR #1436, opened 2021-06-01
Fix benchmarks and remove versions pre 0.5.0 from benchmarking

sc1f:
This PR fixes the currently broken benchmark suite:

- Remove versions before 0.5.0 from benchmarks, as the benchmarking results between such large version bounds is uninformative and maintenance of code that calls deprecated APIs makes the benchmarking code unwieldy and hard to read.
- Re-enables expression column tests, and fixes the broken expressions that prevented certain tests from running/created errors
- Fixes benchmark suite errors


texodus (approved):

PR #1437, opened 2021-06-04
Add integer, float, string, date, and datetime conversion functions

sc1f:
This PR adds a set of type conversion functions to the expression language, which will allow users to quickly clean/coerce their data within Perspective if it is inferred as the wrong format:

- `integer(x)`: converts a column or scalar of any type to an integer. In `Perspective.js`, integers are 32-bit, thus values that overflow/underflow will be output as `null`.
- `float(x)`: converts a column or scalar of any type to a float.
- `string(x)`: converts a column or scalar of any type into a string. `date` values are output using the format `YYYY-MM-DD`, and `datetime` values are output using the format `YYYY-MM-DD HH:MM:SS`. Conversion of floats to strings can lose precision.
- `date(year, month, day)`: given a year, month (1-12), and day, create a new `date` value. This allows for parsing of numerical date formats that were previously impossible to cast to a date, such as `YYYYMMDD`. Parsing can be done with a simple expression:

```javascript
var year := floor("x" / 10000); // "x" is a column of numbers in the format YYYYMMDD
var month := floor("x" % 10000 / 100);
var day := floor("x" % 100);
date(year, month, day)
```
- `datetime(timestamp)`: given a POSIX timestamp of milliseconds since epoch, create a new datetime value. This allows for timestamp columns which would usually be inferred as `float` (unless the table is provided with a schema marking the column as a `datetime`) to be converted into datetimes. Because of the addition of scalars, converting from a seconds-since-epoch timestamp to a milliseconds timestamp is simple:

```javascript
datetime("timestamp" * 1000)
```
Additionally, this PR removes a little debug cruft left over from #1431, and edits the Monaco autocomplete definitions to include the proper parameters for `bucket`, `if/else`, and `var`.

### DIFF
--- a/cpp/perspective/CMakeLists.txt
+++ b/cpp/perspective/CMakeLists.txt
@@ -250,7 +250,6 @@ if (PSP_WASM_BUILD)
 			-s DISABLE_EXCEPTION_CATCHING=0 \
 			-s ASSERTIONS=2 \
 			-s DEMANGLE_SUPPORT=1 \
-			-s SAFE_HEAP=1 \
 			")
 	else()
 		set(OPT_FLAGS " \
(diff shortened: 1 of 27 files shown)


texodus (approved):

PR #1441, opened 2021-06-09
Output more metadata on expression errors, fix #1440

### LINKED ISSUE #1440: get_index() method missing from Typescript Table type definition
### Description of Problem:
I am using perspective with react/typescript and am streaming the table updates via websocket. I am currently unable to get the updates to apply to the index I have set (a unique identitifier), instead it keeps adding more rows and thus have duplicates of all the unique identifiers.

I have updated the index.d.ts in my local environment to include:

 get_index(): Promise<string>;

under the Table type definition.

This enables me to obtain the table index and use this code:

const table = await worker.table(server_view, {                index: await server_table.get_index()            });

### Potential Solutions:
If there is no particular reason not to, I propose that the get_index() method is added to the typescript Table type definition file.

If there is a better/preferred way of dealing with this in react/typescript I would be happy to learn that and implement it.

sc1f:
This PR changes the `validate_expressions` API to output an expression error object versus just an error message.

For errors from ExprTk's parser (invalid symbol, mismatched parentheses, etc.), we use ExprTk's `update_error` method to generate a line and column number for the error. Because we pre-parse the expression strings, on certain expressions the line/column positions of a token that errored might not be totally accurate, as the parsed expression string (`COLUMN1 + COLUMN2`, vs. `"Sales" + "Profit"` as typed by the user)  will not match 1:1 in terms of token position to the string the user typed. However, without doing a large amount of parsing and other validation, this is good enough for a generalized idea of where an error might be (especially as it does not output a range, just a line and a column number).

Each error object now contains the following keys:

- `error_message`: a string that describes the error
- `line`: an integer >= 0 that marks the line the error occurred on
- `column`: an integer >= that marks the column position the error occurred on

Type errors and "Column does not exist" errors will always return 0 for both `line` and `column`.

Additionally, this test cleans up some missing definitions in `index.d.ts` as reported by #1440.


texodus (approved):

PR #1445, opened 2021-06-15
Jupyter tests

texodus:


sc1f:
Closing temporarily so that Azure doesn't run 2 builds per commit until we are ready to make this PR mergable.

PR #1447, opened 2021-06-15
Read CSV strings in perspective-python

sc1f:
This PR enables direct loading of CSV strings inside `perspective-python` using the Apache Arrow-based loader used in `perspective.js`. The CSV loader uses the version of `reader.cc` shipped with Apache Arrow, and not the custom `single_threaded_reader.cpp` which is implemented for the WASM binding. During the loading of CSVs, as is the case with loading regular Arrow binaries, the GIL is released by the C++ binding.

### Changelog

- Enables and tests load & update of CSV strings in Python library
- Unifies CMake build of Arrow CSV reader files
- Fixes a bug in Python where parsing a CSV with a null column, such as `"x,y\n1,"` would cause a segfault when trying to read into an invalid `null_bitmap_data` pointer.

### DIFF
--- a/cpp/perspective/src/cpp/arrow_csv.cpp
+++ b/cpp/perspective/src/cpp/arrow_csv.cpp
@@ -12,9 +12,13 @@
 #include <arrow/util/value_parsing.h>
 #include <arrow/io/memory.h>
 
-// This causes build warnings
-// https://github.com/emscripten-core/emscripten/issues/8574
-#include <perspective/vendor/arrow_single_threaded_reader.h>
+#ifdef PSP_ENABLE_WASM
+    // This causes build warnings
+    // https://github.com/emscripten-core/emscripten/issues/8574
+    #include <perspective/vendor/arrow_single_threaded_reader.h>
+#else
+    #include <arrow/csv/reader.h>
+#endif
 
 namespace perspective {
 namespace apachearrow {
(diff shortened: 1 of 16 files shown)


cpp/perspective/src/cpp/arrow_csv.cpp
```diff
@@ -12,9 +12,13 @@
...
 
-// This causes build warnings
-// https://github.com/emscripten-core/emscripten/issues/8574
-#include <perspective/vendor/arrow_single_threaded_reader.h>
+#ifdef PSP_ENABLE_WASM
+    // This causes build warnings
+    // https://github.com/emscripten-core/emscripten/issues/8574
+    #include <perspective/vendor/arrow_single_threaded_reader.h>
+#else
+    #include <arrow/csv/reader.h>
```
sc1f, line 20: @texodus I can enable the `read_options.use_threads = false;` to true a few lines down - I am assuming it should be safe given that the API is exposed by Arrow and we don't do any manual concurrency for this on our side.

texodus: Arrow seems to use [this implementation](https://github.com/apache/arrow/blob/b73bcf0a0ddacd5adc80389b10b8c1b3820ee97a/cpp/src/arrow/util/thread_pool.cc#L395) for populating the default worker pool, which will consume all hardware concurrency without proper env flags.  We should probably figure out how to override these flags in the loaded env, and to respect what was asked for in `set_num_threads()`.

Interesting though, we could easily reuse this static threadpool in place of TBB, looks like [`ParallelFor()`](https://github.com/apache/arrow/blob/b73bcf0a0ddacd5adc80389b10b8c1b3820ee97a/cpp/src/arrow/util/parallel.h#L35) is nearly a dropin replacement for `PSP_PARALLEL_FOR` ..

texodus (approved):

PR #1448, opened 2021-06-15
Static wstring_converter

sc1f:
In `Perspective.js`, we use an `std::wstring_convert` when converting string scalars to Javascript strings. Currently, the converter is initialized per call to `scalar_to_val`, which occurs in a tight loop - by making the converter static we should gain some performance benefits.

TODO: ascertain said performance benefits


sc1f:
Closing - the performance benefits, in fact, did not materialize.

PR #1453, opened 2021-06-17
Build Windows wheel, limit wheel builds to scheduled and tagged builds

sc1f:
This PR adds a wheel build for Windows (tested locally to make sure it installed), and limits wheel generation on Azure Pipelines to tagged and scheduled builds only. This is especially important on Windows builds as the wheel build adds another 15-20 minutes to the entire build. Additionally, the PR also removes the MacOS Python 3.7 build from the general rotation, and limits our per-commit builds to the following:

- WebAssembly
- ManyLinux2010 Python 3.7
- Windows Python 3.7
- MacOS Catalina Python 3.9

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -133,15 +133,18 @@ parameters:
     type: stepList
     default:
       # Build a python wheel for Mac 10.14
-      - bash: yarn _wheel_python --ci $(python_flag) --macos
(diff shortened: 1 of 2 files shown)


timkpaine (approved):
im +1 so long as we check the scheduled builds for wheel problems

texodus:
Thanks for the PR!

PR #1455, opened 2021-06-18
Fixes #1340 - removes dependency on removed ipywidgets API

### LINKED ISSUE #1340: PerspectiveWidget fails to init on ipywidgets >=8.0.0 due to removed API
## Bug Report

On Perspective 0.6.2, the `PerspectiveWidget` is broken because it uses `on_displayed`, a method that was [removed](https://github.com/jupyter-widgets/ipywidgets/pull/2021) in `ipywidgets` 8.0.0. I didn't notice it for the longest time because my local ipywidgets version has been `7.5.0`, and only recently noticed it when installing Perspective on a new laptop from pip.

In `PerspectiveWidget`, we use `on_displayed` to cache updates made in client mode before the widget has been displayed to the user, as the Perspective Javascript engine does not initialize until the widget has been displayed. When we initially built client mode, the impetus was for Perspective to be supported on platforms where building Perspective was difficult, especially as we had not yet set up any infrastructure for building and testing wheels.

Today, with Perspective wheels being tested on CI and disted to PyPi on release, it might be time to remove client mode entirely. The Perspective docs already make no mention of client mode, and the installation docs explicitly state that you should install `perspective-python` first before installing the Jupyterlab plugin.

### Steps to Reproduce:
 1. Install `perspective-python` from pip, which will install 0.6.2
 2. Try to init a `PerspectiveWidget` inside Jupyterlab
 
### Expected Result:
A working `PerspectiveWidget`

### Actual Result:

Screen Shot 2021-03-02 at 12 29 21 PM

There is a quick workaround for now just to stub out the method:

```python
perspective.PerspectiveWidget.on_displayed = lambda x, y: x
```

### Environment:
- `perspective-python` 0.6.2 installed from pip
- `@finos/perspective-jupyterlab-plugin` 0.6.2 installed from NPM
- JupyterLab 3.0.9
- iPyWidgets 8.0.0.a0
- MacOS Catalina

sc1f:
This PR removes the use of `ipywidgets.Widget.on_displayed` from `PerspectiveWidget`, which makes `PerspectiveWidget` compatible with ipywidget >= 8.0.0.

This API was originally used for an esoteric feature: on the widget, in client mode, if the user called `widget.update` before the widget was displayed, the updates were not cached and thus silently dropped. Since the first releases of `perspective-python`, we've tremendously cleaned up our build process and now offer wheels for all platforms, which should make client mode rarely used, if at all. Thus, this feature is slated for removal to maintain our compatibility with future ipywidgets versions.

### DIFF
--- a/python/perspective/perspective/tests/client_mode/test_client_mode.py
+++ b/python/perspective/perspective/tests/client_mode/test_client_mode.py
@@ -330,22 +330,6 @@ class TestClient(object):
         widget.update(data)
         assert hasattr(widget, "table") is False
 
-    def test_widget_client_update_cached(self):
-        import perspective
-        assert perspective.is_libpsp() is False
-        data = {"a": [1, 2]}
-        widget = perspective.PerspectiveWidget(data)
-        mocked_post = partial(mock_post, assert_msg={
-            "cmd": "update",
-            "data": data
-        })
-        widget.post = MethodType(mocked_post, widget)
-        widget.update(data)
-        assert widget._predisplay_update_cache == [data]
-        widget._displayed = True
-        widget.update(data)
-        assert widget._predisplay_update_cache == [data]
-
     def test_widget_client_replace(self):
         import perspective
         assert perspective.is_libpsp() is False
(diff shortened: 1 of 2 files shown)


texodus (approved):

PR #1459, opened 2021-06-24
Spark bar

texodus:
sparkbar


texodus:
w/negative bars
sparkbar2

nmichaud:
Something looks strange in your last example, where you have bars on both the positive and negative sides.

texodus:
@nmichaud Check the column headers for these columns - these are calculated columns (using new ExprTK lang) which filter +/- rows

nmichaud:
Got it - I was thinking the filtering was on the aggregate value.

ArashPartow:
@texodus  for single line conditional statements you could also use the function conditional, eg:

if ("Profit" < 0, "Profit", 0)

if ("Profit" > 0, "Profit", 0)

PR #1460, opened 2021-06-24
Add Jupyterlab tests to CI

texodus:
This PR adds a set of Puppeteer tests for `perspective-jupyterlab`, which assert the basic functionalities of the Jupyterlab widget in a stable and consistent manner.

To run the tests, run ` yarn test --jupyter` with `perspective-jupyterlab` selected during `yarn setup`. By default, the tests use the currently set `$PYTHONPATH` of the shell, which may not resolve to the most up-to-date local build of `perspective-python`. Simply run the test command with `PYTHONPATH` set to your local directory, i.e:

```bash
PYTHONPATH=./python/perspective:$PYTHONPATH yarn test --jupyter
```

PR #1466, opened 2021-06-30
Inlined build for perspective-jupyterlab, improve PerspectiveWidget

### LINKED ISSUE #1454: Perspective 0.7.0 and above does not work in Voila
### Steps to Reproduce:
 1. Open a Jupyter notebook that uses `PerspectiveWidget` in Jupyterlab, where it renders correctly.
 2. Render the same notebook using Voila.

### Expected Result:
Working `PerspectiveWidget` instances.

### Actual Result:
As expected, Voila falls back to `jsdelivr` as a CDN to deliver `perspective-jupyterlab`. Once Voila tries to load the module from CDN, it hits an error:

Screen Shot 2021-06-18 at 1 51 21 PM

### Environment:
Perspective 0.9.0, MacOS Big Sur, Jupyterlab `3.0.14`, Voila `0.2.10`, `@voila-dashboards/jupyterlab-preview v2.0.2`

### Additional Context:

This issue seems related to [https://github.com/finos/perspective/discussions/1452](#1452), but I am not sure of the exact solution at the moment.

sc1f:
This PR changes the `perspective-jupyterlab` build to inline its external dependencies on the main Perspective modules, and makes some long-overdue improvements to `PerspectiveWidget` in order to enhance the user experience in Jupyterlab and Voila. Specifically:

- Fixes #1454, which was due to Voila not being able to access perspective-jupyterlab's external dependencies on the main Perspective modules. While the asset produced (10.6MB as loaded into a Jupyterlab production build) is not ideal, it is guaranteed to work in both Jupyterlab and Voila without additional hard-to-maintain/brittle build cruft.
- Fixes long-running bug where `PerspectiveWidget` had a height of 0 when rendered inside Voila. The height now defaults to 520px, the same as in Jupyterlab.
- The `<perspective-viewer>` config panel now shows by default on all widgets when rendered.
- `worker.terminate()` and `viewer.delete()` is now called whenever the widget output is disposed of, which should make memory usage in Jupyterlab more efficient.
- Fixes #1462: `PerspectiveWidget` now inherits from `ipywidgets.DOMWidget` instead of `ipywidgets.Widget`, which means that `layout` should work out of the box:

widget_layout

- `plugin_config` can now be passed in from Python, which should allow for much finer-grained control over styling of the datagrid and chart plugins. However, because `plugin_config` cannot be read synchronously from the Javascript front-end (as it is not part of the attribute API and there is no async `get` in Javascript), there is currently no way to _read_ the current `plugin_config` from the widget unless it has been set by the user already. However, this is still very useful for developers to "pre-set" styles on `PerspectiveWidget`:

plugin_config

- Modified `WebSocketServer` to strip out semver-like ranges from requests. This is helpful when debugging Voila locally, as I can just set the Voila CDN to `localhost:8080/node_modules/` and all the CDN requests Voila makes will automatically be resolved against my local build.


packages/perspective-jupyterlab/src/config/plugin.config.js
```diff
@@ -27,7 +27,7 @@ module.exports = {
         maxEntrypointSize: 512000,
         maxAssetSize: 512000
     },
-    externals: [/^([a-z0-9]|@(?!finos\/perspective-viewer))/],
+    externals: [/\@jupyter|\@lumino/],
```
timkpaine, line 30: Should this include jupyter widgets external?

timkpaine (approved):

dhirschfeld:
Super keen to get this released so I can get my hands on the voila fixes! :rocket:

texodus:
Thanks for the PR!  A couple post-review notes:

* There is a non-zero chance this change does not actually fix Voila, as we can't fully test the interaction with CDNs and Perspective's multiple-asset build.
* I still do not fully understand much of JupyterLab/Voila packaging, e.g. why `@finos/perspective` needs to be inlined, or conversely why `@juypter-widgets` does not.  This combined with minimal testing (though much improved on the JupyterLab side since #1460) is a recipe for future regressions here.

dhirschfeld:
Unfortunately I can't help with the development or packaging but I'm happy to test the JupyterLab/Voila combo on our deployment and report back any issues...

sc1f:
@dhirschfeld @texodus

I have just tested the 0.10.0 release in a fresh virtualenv with `perspective-python` installed from wheel, `@finos/perspective-jupyterlab` installed from NPM, and everything works. Voila falls back to JsDelivr, and loads the plugin and all externals correctly with or without `--enable_nbextensions=True` set when running Voila.

PR #1476, opened 2021-07-09
Add standard deviation and variance aggregates

sc1f:
This PR adds two new aggregates: `stddev` and `var`, which respectively calculate the standard deviation and variance across each group in a row-pivoted view. Extensive tests have been added, and another incidental bug (previously, `to_arrow()` could not serialize a view with the `mean` aggregate set) has been fixed as well.


examples/workspace-editing-python/package.json
```diff
@@ -7,7 +7,7 @@
         "start": "yarn webpack && yarn start:server",
         "start:client": "webpack-dev-server --open",
         "start:server": "PYTHONPATH=../../python/perspective python3 src/server.py",
-        "webpack": "webpack --colour"
+        "webpack": "webpack --color"
```
timkpaine, line 10: Nice

timkpaine (approved):

examples/workspace-editing-python/package.json
sc1f: 😤🇺🇸

texodus:
Thanks for the PR!

PR #1480, opened 2021-07-16
Add required dependencies for webpack plugin

LukeSheard:
- Resolve loaders relative to the current package location to ensure that the `perspective` version of loaders is picked up. 
- Add required dependencies for webpack plugin, currently when installing into a repository without `string-replace-loader` builds will start failing due to the missing dependency.


texodus (approved):
Looks good!  Thanks for the PR

PR #1516, opened 2021-08-16
Add new aggregates to ViewConfig enum

sc1f:
This fixes an issue where some aggregates (`stddev`, `var`, and `first by index`) are not working in the viewer because they were not added to the `ViewConfig` aggregate in `perspective-vieux`.


texodus (approved):

PR #1518, opened 2021-08-23
Fix count aggregate when last aggregate and partial updates are applied

sc1f:
This PR fixes a long-running issue where a pivoted view with a column showing the `last` aggregate would increase the `count` of another column during partial updates that edited values but did not add or remove rows.

The correct behavior, as applied in this fix, is that the count should increase when a row has been added and decrease when a row was removed, but when a column with `last` was applied to the view, the `count` on another column would increase even when rows were not added or removed, but edited through a partial update. Specifically, the code responsible for calculating the "strand count"—the number of rows inside a pivot group—treats a row as a "new row" if the value has been changed, which means a partial update that only edits existing cells were treated as "new rows".

This issue only manifests when a column with `last`, `high`, or `low` are visible in the pivoted view, as the sparse tree used to calculate contexts adds those columns to the strand column set in order to calculate (to my understanding) the final output value, which is then parsed in `t_stree::build_strand_table_phase_1`, causing the issue.

The fix is relatively simple - the method should check that the row was not a new added or deleted row, not just whether a value has been changed, and this is achieved through checking another value of the `t_value_transition` enum—`VALUE_TRANSITION_NEQ_TT`, which signifies that the value in the row was changed but the row remains valid, which is true for edited values but not for added or removed rows. I've added a good amount of tests to assert correct behavior across different scenarios.

### DIFF
--- a/cpp/perspective/src/cpp/gnode.cpp
+++ b/cpp/perspective/src/cpp/gnode.cpp
@@ -154,7 +154,7 @@ t_gnode::calc_transition(
     bool prev_pkey_eq) {
     t_value_transition trans = VALUE_TRANSITION_EQ_FF;
 
-    if (!row_pre_existed && !cur_valid) {
+    if (!row_pre_existed && !cur_valid && !t_env::backout_invalid_neq_ft()) {
         trans = VALUE_TRANSITION_NEQ_FT;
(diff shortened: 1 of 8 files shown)


texodus (approved):

PR #1519, opened 2021-08-23
Refactor `to_arrow`, fix row deltas for pivoted views

sc1f:
This PR makes refactors to `to_arrow` so that it better matches the behavior of `to_format` by skipping hidden sorts, fixing bugs that occurred when emitting deltas for 1 and 2-sided contexts, which in many situations were completely incorrect. Additionally, the `unique` aggregate now emits `null` for non-unique string values instead of the string `"-"`, which is not semantically correct. A large number of tests have also been added to ensure correctness.


texodus (approved):

Issue #998, opened 2020-04-03
`table.remove(keys)` from node seems to corrupt indexing?


andyferris:
### Steps to Reproduce:
 1. In a node.js backend, we create a new table called `table` and add some rows. We are viewing this in the browser through a seperate frontend app, connecting via websocket.
 2. In the backend, we call `table.remove(keys)` on with one or more keys (a small fraction of the data).
 3. In the browser frontend, using the search functionality to e.g. filter for a particular row (on the value of a primary key, for example) we are shown an incorrect row.

### Expected Result:
The filtered row should obey the filter.

### Actual Result:
Another row is displayed. For example a `==` filter on a primary key displays exactly one row, but it is actually one of the neighboring rows with a different (slightly offset) primary key.

### Environment:
Node 12, Perspective 0.4.6, Google Chrome 80.0.3987.149

### Additional Context:
Reloading the browser window does not fix the problem - so whatever is going on, I suspect it has more to do with the `Table` in node rather than the `View` in the browser.

I haven't yet looked into the implementation of perspective (what's a gnode?)... but given it is so fast, I suspect there are some secondary acceleration indices maintained on the table columns, and the observed behaviour is consistent with these not being updated correctly when a row is removed. E.g. perhaps internally the filter finds the index to a row through the acceleration index, but this is somehow offset to the real data. It doesn't seem to matter which column we filter on - primary key or not (one example had multiple columns that happened to be unique, and filtering on any of them returned the same incorrect row).

Here's an example where `Job.Code` is offset by 2. (I get the same row when I filter on the corresponding `Job.ID` instead).

![image](https://user-images.githubusercontent.com/2974526/78330553-919abb80-75c7-11ea-9c57-eebdd67aee16.png)

sc1f:
Are you able to reproduce this issue when the table and view are all in the same browser process? I took a try at recreating the same scenario using the Python library (which shares the internal C++ with the Javascript library), and set up a quick example where rows were being removed every 500ms:

![filters](https://user-images.githubusercontent.com/13220267/78849272-00ba4a80-79e2-11ea-8e68-907628b46103.gif)

As far as I can tell, the behavior seems correct here when filtering across different columns, including the primary key column ("id"). We don't have solid test coverage around how `remove` interops with other operations, so this definitely seems like a real issue, but I bet it's buried deep within the C++ instead of being specific to the node remote binding.

[...]

PR #1528, opened 2021-09-02
Fix #1505, #998, #1225 - results after remove are correct

### LINKED ISSUE #1225: table.size() incorrect after remove()
### Steps to Reproduce:
 1. Create an indexed Table.
 2. Remove a subset of rows using `table.remove()`
 3. Call `table.size()`

### Expected Result:
`table.size()` returns the size of the table with elements removed.

### Actual Result:
`table.size()` returns the size of the table without removed elements. This is due to remove() nulling out the rows on the `t_gstate` master table without physically removing them, most likely for performance reasons. Data received from `View` objects and `view.num_rows()` are correct, as Views maintain a traversal that will ignore deleted rows.

### LINKED ISSUE #1505: Inconsistent view output under frequent insert&delete
## Bug Report 
When creating new view with new filters under frequent insert and delete, filtered data output could be inconsistent with view config.

### Steps to Reproduce:
 1. Create worker
 2. Create table
 3. Insert and delete table data frequently
 4. Create view without filters
 5. Create view with filters, delete previous view
 6. If it is not reproduced, go back to 4. step

### Expected Result:
I expect to see that the output data of view to be consistent with view config. 

### Actual Result:
When I repeatedly create new view with new config (just filters are different), I eventually encounter some inconsistent output. 

**Screenshot:**
Screen Shot 2021-08-05 at 14 34 46

**Reproducable Example**
This project clearly reveals the bug. But before you open, I want to clarify a few things:

1. There are 3 important section which are aligned horizontally. 
   - PerspectiveViewer
   - view.to_json() output
   - view.get_config() output
2. When you open the project it will automatically create worker, table and view respectively.
3. After table is created, it will start these periodic operations:
   - update(<5rows/500ms), 
   - insert(<5rows/500ms),
   - delete(<5rows/1000ms) .
4. In parallel it will recreate view (with deleting previous one) once in 5000ms with changing just filters in view config. As follows:
   - `{ ...prevConfig, filters: []}`
   - `{ ...prevConfig, filters: [['Column1', 'in', ['C', 'E']]]}`
   - `{ ...prevConfig, filters: []}`
   - `{ ...prevConfig, filters: [['Column1', 'in', ['C', 'E']]]}`
   - and so on...

I hope I didn't confuse. When you open the project, just wait for the error.

[Codesandbox Project](https://codesandbox.io/s/perspective-filtering-error-bcypx?file=/src/App.tsx)

### Environment:
Codesandbox, React Application dev and prod environments.
Browser: Chrome 92.0.4515.131 (Official Build) (x86_64)
Perspective: 0.10.0
OS: macOS Big Sur

sc1f:
This PR fixes #1505, #998, and #1225 by making changes to `t_gstate::get_pkeyed_table`, which is used to "seed" new views with the state of the underlying table. However, if the table had removed rows and a string primary key, it would enter a block of code that rebuilt the index column and did so with an offset of 0 if an empty string ("") was in the primary keys, and 1 otherwise. This caused the internal `psp_pkey` index column to be off-by-one, which does not appear as a problem visually but caused issues when filtering, as Perspective was now returning the wrong rows entirely in the filter.

By throughly testing the logic and removing the offending code block, the issue is fixed and filters after removes and updates should now return the correct rows. This behavior also only materializes when a _new view_ is created after the remove/update, as `get_pkeyed_table` is only used for new views. Thus, we've added a battery of tests asserting the equality of behavior when _a new view is created_ and when we _reuse existing views_. This provides confidence that removing the block of code is, in fact, the correct solution and guarantees correct behavior.

Additionally, I've fixed #1225 by making `table.size()` return the size of the mapping of the gnode state instead of the gnode state master table itself.

A moment to opine; all `t_data_tables` are allocated as a contiguous chunk of memory, and when rows are removed, the physical row is not removed from the master table, as it would cause a re-allocation. Instead, the row is cleared (its values removed) and its primary key removed from `m_mapping`, which is a mapping of primary key scalars to row indices on the master table. The row on the master table is then kept for later updates, which will seek to reuse existing rows that have been removed instead of appending new rows and thus causing more allocations.

This means that`t_gstate.m_table.size()` is always the _totality of all rows appended throughout its lifetime_, and will always be `>=` to the size of `m_mapping`. In contrast, `m_mapping` is a hashmap and primary keys can be evicted at will. Thus, `m_mapping.size()` is always `appended rows - removed rows`, and is the correct `size()` of the table to be provided to the external user-facing API. I've held off on making this change for some time in order to better understand the remove mechanics, and I think this batch of fixes is a good chance to correct that behavior.

Finally I had to make some ESLint changes after rebasing to master, and this PR includes those changes as well to Javascript test files (commenting out some unused imports).

### DIFF
--- a/cpp/perspective/src/cpp/gnode.cpp
+++ b/cpp/perspective/src/cpp/gnode.cpp
@@ -1253,16 +1253,6 @@ t_gnode::clear_output_ports() {
     }
 }
 
-t_data_table*
-t_gnode::_get_pkeyed_table() const {
-    return m_gstate->_get_pkeyed_table();
-}
-
-std::shared_ptr<t_data_table>
-t_gnode::get_pkeyed_table_sptr() const {
-    return m_gstate->get_pkeyed_table();
-}
-
 void
 t_gnode::set_pool_cleanup(std::function<void()> cleanup) {
     m_pool_cleanup = cleanup;
@@ -1281,11 +1271,6 @@ t_gnode::was_updated() const {
 void
 t_gnode::clear_updated() {
     m_was_updated = false;
-}
-
-std::shared_ptr<t_data_table>
-t_gnode::get_sorted_pkeyed_table() const {
-    return m_gstate->get_sorted_pkeyed_table();
 }
 
 std::string
--- a/cpp/perspective/src/cpp/gnode_state.cpp
+++ b/cpp/perspective/src/cpp/gnode_state.cpp
@@ -209,7 +209,8 @@ t_gstate::update_master_table(const t_data_table* flattened) {
                 m_pkcol->set_scalar(master_table_indexes[idx], pkey);
             } break;
             case OP_DELETE: {
-                // Actually erase the specified pkey from the master table here
+                // Erase the pkey from the master table, but this does not
+                // change the size as the row isn't removed, just cleared out.
                 erase(pkey);
             } break;
             default: { PSP_COMPLAIN_AND_ABORT("Unexpected OP"); } break;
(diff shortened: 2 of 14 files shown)


texodus (approved):

The arsenal of tests accompanying this change IMO are great!  This will help a lot to validate futures changes to the engine with confidence.

PR #1552, opened 2021-09-26
Fix python wheel CI

texodus:
@timkpaine

### DIFF
--- a/python/perspective/scripts/test_wheels.sh
+++ b/python/perspective/scripts/test_wheels.sh
@@ -4,7 +4,6 @@
 #
 # This file is part of the Perspective library, distributed under the terms of
 # the Apache License 2.0.  The full license can be found in the LICENSE file.
-set -e
 
 # 
 # Creates a new virtual environment and installs Perspective from a wheel,


timkpaine:
Would be better to remove the line in this file that fails

texodus:
Yes that's exactly what this PR does

timkpaine:
That line turns on "if something in the script fails, fail the whole script". If it's presence causes the script to fail, that implies that something else in the script is bad

texodus:
Yes I understand, the error-ing command is the [failing `grep`](https://github.com/finos/perspective/blob/master/python/perspective/scripts/test_wheels.sh#L90) (e.g. no wheels).  Checking through older builds, It appears these wheels have never worked, you just added `set -e` without testing it against these environments in #1433 

As these failing builds now block CI, I'm reverting the check.

texodus:
I do see [published artifacts for the failing builds](https://dev.azure.com/finosfoundation/perspective/_build/results?buildId=3255&view=artifacts&pathAsName=false&type=publishedArtifacts) though ...

timkpaine:
We really do want to check the wheel installs, but maybe bypass it if no wheels present (or maybe we want to fail if not present, I don't know the original logic).

texodus:
@timkpaine [broken again](https://dev.azure.com/finosfoundation/perspective/_build/results?buildId=3325&view=logs&j=b2d14303-4fa2-58a3-f0aa-67fdcd29dd16&t=27d59515-415e-54f0-b4b9-81d8339ea06e&l=439)

timkpaine:
This is because python formats `rc` as `1.0.0rc` whereas it's being tagged as `1.0.0-rc`

PR #1567, opened 2021-10-01
Fix #1566, remove deprecated flags from WASM debug build

### LINKED ISSUE #1566: Loading remote table from Python breaks the viewer
Screen Shot 2021-10-01 at 1 38 04 PM

sc1f:
This PR fixes server mode Python by fixing the output of `validate_expressions`, which was missing a field in the Python implementation. In the pre-1.0.0 viewer, `expression_alias` was a map of expression aliases to the whole expression string so the user could hover over a column in the UI and see the expression from whence it came. With the "edit expression" button added in 1.0.0, it is no longer necessary to see the whole expression separate from the editor.

Even though the `expression_alias` field is not used in the viewer, it is required and type-checked by Rust and thus threw an unrecoverable error when loading the table from Python. The long-term fix is to move all validation and parsing logic for expressions, along with other logic that is implemented twice in both JS and Python, all into C++.

Additionally, the WASM debug build was outputting deprecated flag warnings, which have been rectified.

### DIFF
--- a/cpp/perspective/CMakeLists.txt
+++ b/cpp/perspective/CMakeLists.txt
@@ -232,9 +232,9 @@ if (PSP_WASM_BUILD)
 	if(CMAKE_BUILD_TYPE_LOWER STREQUAL debug)
 		set(OPT_FLAGS " \
 			-O0 \
-			-g4 \
+			-g3 \
+			-gsource-map \
 			--profiling \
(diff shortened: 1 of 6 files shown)


timkpaine (approved):

texodus (approved):

PR #1579, opened 2021-10-13
Update python classifiers

timkpaine:
Screen Shot 2021-10-13 at 10 24 35 AM


texodus:
Thanks for the PR!

We don't actually support Python 2 anymore, either - but looks good!

PR #1582, opened 2021-10-14
remove deprecated webpack-specific tilde prefix convention

timkpaine:
With the existing webpack-specific tilde convention, other bundlers have difficulty managing perspective's `.less` files so they need to be vendored separately (e.g. with esbuild https://github.com/iam-medvedev/esbuild-plugin-less/issues/15#issuecomment-830181089 and i have had some trouble with rollup as well).

According to [this less-loader doc](https://github.com/webpack-contrib/less-loader#webpack-resolver), the tilde is deprecated as the resolution will fail over to `node_modules` anyway.

### DIFF
--- a/packages/perspective-workspace/src/less/workspace.less
+++ b/packages/perspective-workspace/src/less/workspace.less
@@ -8,7 +8,7 @@
  */
 
 :host {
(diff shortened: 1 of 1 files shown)


texodus:

PR #1583, opened 2021-10-14
Remove python 2 from build and CI scripts

timkpaine:
I don't care about python 2 anymore, and [I was assured that nobody else would after December 2020](https://www.efinancialcareers.com/news/2021/02/jpmorgan-still-has-its-python-2-issue) 😉 🇬🇷 🦉


texodus:
[Looks good!](https://dev.azure.com/finosfoundation/perspective/_build/results?buildId=3421&view=results)

PR #1598, opened 2021-10-28
More regex functions - find, substring, replace

sc1f:
PR on top of #1596, but this one can be merged first to show all the changes - it's in a separate PR to make code review easier.

This PR adds more regex functions:

- `find(string, pattern, output_vector)`: writes the start and end-index (0-indexed) of the first match in string of the first capturing group in pattern. Values can be read from the output vector as follows: `var vec[2]; find("x", '(.*)', vec) ? vec[0] + vec[1] : null`. Returns `true` if a match was found and values were written to the output vector, and `false` otherwise.
- `substring(string, start_idx, length?)`: returns the substring of string from `start_idx` (0-indexed) with the specified `length`. If the length is not provided, returns the substring from start_idx to the end of the string, or null if the string is invalid or indices are invalid.
- `replace(string, pattern, replacer)`: replaces the _first_ match of `pattern` in `string` with `replacer`, and returns the replaced string or the original string if no replacements were made.
- `replace_all(string, pattern, replacer)`: replaces _all_ matches (non-overlapping) of `pattern` in `string` with `replacer`, and returns the replaced string or the original string if no replacements were made.

This PR also renames `fullmatch` from #1596 to `match_all` to remain consistent with the `replace`/`replace_all` API.

With this suite of Regex functions, one can easily clean and convert data from within Perspective - poorly formatted strings, dates that could not be parsed, monetary values that you want to convert to a float but cannot because of other string tokens, etc. can all be cleaned and properly converted from within Perspective's expression API.

### DIFF
--- a/cpp/perspective/src/cpp/computed_expression.cpp
+++ b/cpp/perspective/src/cpp/computed_expression.cpp
@@ -436,11 +436,15 @@ t_computed_function_store::t_computed_function_store(t_expression_vocab& vocab,
     , m_lower_fn(computed_function::lower(vocab, is_type_validator))
     , m_to_string_fn(computed_function::to_string(vocab, is_type_validator))
     , m_match_fn(computed_function::match(regex_mapping))
-    , m_fullmatch_fn(computed_function::fullmatch(regex_mapping))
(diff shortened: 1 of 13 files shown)


texodus (approved):

PR #1611, opened 2021-11-16
Port to `esbuild`

texodus:
Replace `webpack` and `rollup` composite build with `esbuild`, substantially improving the developer experience, as all JavaScript/TypeScript/CSS assets can be built in under 10 seconds, as opposed to 2-3 minutes on webpack.  `@finos/perspective-webpack-plugin` still works and can build these assets via webpack, so this change should really only affect how perspective is built for packaging.

The main potentially-breaking change is that NPM packages have moved from `"main"`/`"module"`/`"browser"`/etc key menagerie to [`"externals"`](https://webpack.js.org/configuration/externals/), which is only supported by `webpack` `^5` and node `12.7.1`.  For `webpack`, while support for version `^4` was already deprecated in `@finos/perspective-webpack-plugin`, perspective could still be built without the plugin using the package's UMD module output (the default);  this is no longer the case, though older webpack versions should still be able to link directly to the UMD filepath within the module, e.g. `@finos/perspective/dist/umd/perspective.js`.  Note also that `externals`, unlike the previous package descriptors, makes other files private by default, so there is some risk of unexpected file resolution issues if a package imports something we did not expect.


examples/workspace/package.json
```diff
@@ -5,7 +5,7 @@
     "description": "An example app built using `@finos/perspective-workspace`.",
     "scripts": {
         "start": "webpack serve --open",
-        "webpack": "webpack --colour"
+        "webpack": "webpack --color"
```
timkpaine, line 8: 🇺🇸

packages/perspective-jupyterlab/package.json
```diff
@@ -39,16 +35,16 @@
         "@finos/perspective-viewer": "^1.0.1",
         "@finos/perspective-viewer-d3fc": "^1.0.1",
         "@finos/perspective-viewer-datagrid": "^1.0.1",
-        "@jupyter-widgets/base": "^3.0.0 || ^4.0.0",
+        "@jupyter-widgets/base": "^4.0.0",
```
timkpaine, line 38: This is a breaking change just to note

packages/perspective-jupyterlab/src/less/index.less
```diff
@@ -1,7 +1,7 @@
-@import "@finos/perspective-viewer/src/less/fonts.less";
-@import "@finos/perspective-viewer/src/themes/material-dense.less";
+// @import "../../../../rust/perspective-viewer/src/less/fonts.less";
+@import "../../../../rust/perspective-viewer/src/themes/material-dense.less";
```
timkpaine, line 2: Can you symlink into node_modules in `postinstall` (like for workspace packages) and avoid a big relative import?

tools/perspective-build/package.json
```diff
@@ -0,0 +1,17 @@
+{
+    "name": "@finos/perspective-build",
+    "version": "1.0.1",
+    "description": "Workspace-wide scripts for Perspective build, mainly esbuild and webpack configuration",
+    "private": true,
```
timkpaine, line 5: Is it worth making public, I have other esbuild + perspective projects I wouldn't mind using these plugins in.

packages/perspective-jupyterlab/package.json
texodus: I need to check - this branch has been under development for several weeks and I don't recall why I removed this, so I'll need to revert and verify.  I suspect it was because `^3.0.0` does not work even in already-released versions - if so, this is only a breaking change by technicality.

packages/perspective-jupyterlab/src/less/index.less
texodus: These are relative because of the introduction of `externals` in package.json, e.g. they are ultimately-inlined shared code between packages, not runtime dependencies.  Docs for these changes is forthcoming.

tools/perspective-build/package.json
texodus: I will add a `@finos/perspective-esbuild-plugin` package for this, but it may not be in this release.  I really want this change to be backwards compatible at least for the next release, as there is little user-facing value in this PR.  As a result, there is _a lot_ of complexity here that ultimately is not necessary, e.g. support for non-module `<script>` imports of the UMD/CDN versions.  I would prefer if this complexity is excised before inviting developers to depend on it in their own builds.

packages/perspective-jupyterlab/package.json
texodus: I remember now - this dep resolved to only `"^4.0.0"` in our `yarn.lock`, hence we actually have no idea if it works for `"^3.0.0"`.  Setting this to `"^3.0.0"` does cause yarn install issues, so I'll likely still remove this, but there's no reason to intermix this change with the other 10k lines of esbuild changes, so I've backed this out for now.

packages/perspective-jupyterlab/src/less/index.less
texodus: Removed this change and made `src` public.

tools/perspective-build/package.json
timkpaine: yep sounds good

PR #1613, opened 2021-11-17
Publish script for Python

texodus:


timkpaine:
Just to note now since we plan to do this later, when we finish the prebuilt extension stuff the python build will be like the jupyterlab tests and require the prior completion of the JS builds before it can be disted (so it can just grab the packed js assets without having to do a complete separate wasm build).

PR #1625, opened 2021-11-22
add threads explicit on mac

timkpaine:


texodus (approved):

PR #1629, opened 2021-11-23
Fix Azure windows wheel step

texodus:
Karmic retribution for not running a full azure build before merging these wheel changes originally ..


timkpaine (approved):

timkpaine:
`# azure pipelines are bad`

texodus:
This branch will be `perspective-python@1.0.3`, since the package itself is identical to what was tagged, but the tagged commit's azure config did not build Windows wheels correctly.

PR #1633, opened 2021-11-28
Add auto-resizing to `<perspective-viewer>`

texodus:
Adds auto-reiszing behavior (on by default) to `<perspective-viewer>`, via an internal [`ResizeHandler`](https://developer.mozilla.org/en-US/docs/Web/API/Window/resize_event) on these elements.  As a result, `notifyResize()` is no longer needed for resizing strictly and will likely eventually be removed as it can lead to duplicate re-renders.  For now, it has been preserved and behavior modified slightly - calling `notifyResize()` when auto-sizing is enabled will log a warning and disable auto-size, unless the `force` bool argument is provided.

Auto size behavior can be disabled, reverting to the old manual resize logic via `notifyResize()` (which will not warn in this case), by calling `setAutoSize(false)`:

```javascript
elem = document.querySelector("perspective-viewer");
elem.setAutoSize(false);
elem.notifyResize(); // No warning
```

### DIFF
--- a/docs/js/index.js
+++ b/docs/js/index.js
@@ -244,7 +244,7 @@ window.addEventListener("DOMContentLoaded", async function() {
                 container.style.display = "flex";
                 await psp1.restore(config_defaults(EXAMPLES[key].config));
                 await psp1.toggleConfig(true);
-                await psp1.notifyResize();
+                await psp1.notifyResize(true);
                 container.style.opacity = 1;
                 container.style.pointerEvents = "";
             });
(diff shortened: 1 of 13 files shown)


timkpaine:
I think the windows build got hung so i force cancelled it, but the webassembly build failed with:
```
Summary of all failing tests
 FAIL  packages/perspective-viewer-datagrid/test/js/superstore.spec.js (9.376s)
  ● superstore.html › pivot › by two rows and two columns

    expect(received).toBe(expected) // Object.is equality

    Expected: "cdaa134913bd6cae25d335ec477d5606"
    Received: "20da969a8500883c68e74abbc6de7ff3"
```

PR #1635, opened 2021-11-28
[Formatting] force line-feed line endings for all text files for easier windows development

timkpaine:
im doing some work on windows for conda and dont want to leak crlf into the code, we only enforce on `.js` files rn

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -3,7 +3,6 @@ docs/**/* linguist-documentation
 python/perspective/notebooks/* linguist-documentation
 python/perspective/docs/* linguist-documentation
 results/*.json linguist-documentation
-*.js text eol=lf
 **/test/**/* linguist-documentation
 **/tests/**/* linguist-documentation
 examples/**/* linguist-documentation
@@ -13,3 +12,5 @@ cmake/** linguist-documentation
 CMakeLists.txt linguist-documentation
 docker/** linguist-documentation
 **/*.sh linguist-documentation
+
+* text=auto eol=lf


timkpaine:
@texodus this is good to go, should basically be a no-op but will be helpful when I work on perspective on windows as I won't leak CRLF into the codebase

timkpaine:
if you want I can do a full `dos2unix` on the codebase to purge any existing CRLF

texodus:
@timkpaine I would prefer a lint hook that fails on CI when CRLF are present anywhere, to just a single PR that just fixes them once.  Unfortunately configuring proper linting for this project is a pain - the linting/formatting story for every language Perspective uses is bad, and bad for entirely different reasons - 

* JavaScript has multiple competing standards for linting and formatting, as well as some cultural obsession with separation-of-concerns such that linting and formatting are performed by different libraries.  Don't get me started on TypeScript ...
* `clang-format` applies arbitrarily different rules depending on what version of clang/OS you use, meaning local dev and CI get different failures (or even just different OSes for the same PR).
* HTML seems to be the formatter no one cares about, except ****ing VSCode which insists on changing every line in the file based on that version's arbitrary standards every time I go to update an example.
* Speaking of VSCode, it has validators for TOML, YAML, JSON - and they all have different hard-coded indentations that the user and file settings don't override (2, 2, 4, respectively).  Plus, exactly none of them correctly validate.
* To top it off, running all of these validation/auto-formatters in sequence takes substantially longer than `esbuild` now.

Sorry, I know that's not what you asked ... I think I've been repressing that for a long time.

texodus (approved):

timkpaine:
@texodus well with the `.gitattributes` change, any time you `git add` a file it should have CRLF swapped for LF, no need for it to even reach `precommit`. But it will only apply from now on

PR #1636, opened 2021-11-28
Upgrade apache arrow to 6.0.2

timkpaine:


timkpaine:
@texodus this is good to go but obviously just a maintenance PR, would be good to run the benchmarks to see if theres any material difference

texodus (approved):

We should regression test the performance of this, but its not going to ultimately prevent us from upgrading, so I'm going to put this in a separate issue and merge this now to prevent it from bitrotting.

texodus:
Closing in favor of #1680

Issue #1648, opened 2021-12-02
Themes moved causing build errors


timkpaine:
## Bug Report

`@finos/perspective^1.0.3`

```
Module not found: Error: Can't resolve resolve '~@finos/perspective-workspace/dist/theme/material.less' in '/Users/...`
```

Looking in the folder:
<img width="390" alt="Screen Shot 2021-12-02 at 15 58 30" src="https://user-images.githubusercontent.com/3105306/144502015-5789d05a-3ac4-407f-ba66-43d08b80a46f.png">

I \*believe\* this should break these two examples:
- [remote-workspace](https://github.com/finos/perspective/blob/e95dc5249542008a7b2cd84d6a61886610e0542a/examples/remote-workspace/src/index.less#L10)
- [workspace-editing-python](https://github.com/finos/perspective/blob/e95dc5249542008a7b2cd84d6a61886610e0542a/examples/workspace-editing-python/src/index.less#L10)

I think we need to copy theme assets in addition to css assets to maintain the same importability:
Add here: https://github.com/finos/perspective/blob/master/packages/perspective-workspace/build.js#L81
To account for removal from web pack step here: https://github.com/finos/perspective/commit/5df1d045c170799ed2726262d51ec780266c5486#diff-9125ba2248bc16eb862b5199a87368613b083e054809ce7500b1091c1e73af38L19

texodus:
We use `externals` now which should allow us to alias this to as many legacy locations as we want, at least for all platforms which recognize `externals` which (I believe) is a strict superset of those which Perspective has plugins (e.g. `webpack@5` as of now).  This annoyingly doesn't work on CDNs though.

In the (near?) future, I'd like to move towards a theme system that allows theming to be controlled within the UI itself, and this will almost certainly be the end of separate JS/CSS assets anyway, at least in the sense that `material.css` is currently a pre-requisite for any app.

PR #1662, opened 2021-12-10
Fix `perspective-workspace` and add `monaco-editor-webpack-plugin` compatibility to `perspective-webpack-plugin`

texodus:
* Add support for multiplexing `@finos/perspective` and `@finos/perspective-viewer` modules to their ESM or UMD builds from the new `esbuild` script, which allows `perspective-workspace` to be subsequently used as a dependency without forcing a build type.
* Fix `perspective-workspace` package metadata to export the internal `lumino` components for integration into Lumino apps.
* Updates to the `perspective-workspace` theme:
    * Left-justified panel headers.
    * Add color level and border between `<perspective-viewer>` background and spacers.
    * Expand Global Filter resize bar to 8px, overlap layout left padding.
    * Remove Global Filter margins entirely.
* Add `monaco` option to `@finos/perspective-webpack-plugin` which allows disabling Monaco compilation, which makes it easier to embed Perspective in an application which uses its own `monaco-editor` and `monaco-editor-webpack-plugin`.
* Adds an example of Perspective + React + Monaco Editor in a single app, which uses Monaco's JSON support to show a live view of a `<perspective-viewer>` config.

    monaco-plugin


timkpaine:
this should take care of https://github.com/finos/perspective/issues/1648? breaking change but will export the themes in src?

texodus:
@timkpaine I hadn't planned on it but fair enough, I said I would.  As mentioned, this is going to get deprecated fully soon anyway, as we extend perspective's theme support to allow UI driven theming, light/dark from user preferences, etc.  I'm also going to remove the `/themes/` references from `/examples/`, so these remappings will only be for integration.

As maybe I failed to mention, this backport will only work from webpack - CDN paths will need to be updated.

PR #1668, opened 2021-12-11
Build windows wheel every time so its tested

timkpaine:


texodus:
[full build](https://dev.azure.com/finosfoundation/perspective/_build/results?buildId=3784&view=results)

I think you meant to merge this with master?  It is currently a PR to merge into #1667

PR #1672, opened 2021-12-12
Build windows wheel every time so its tested

timkpaine:
#1668 against master


timkpaine:
https://dev.azure.com/finosfoundation/perspective/_build/results?buildId=3784&view=results#

timkpaine:
https://dev.azure.com/finosfoundation/perspective/_build/results?buildId=3786&view=results

texodus (approved):

I've squashed these commits so the git history will be legible.

texodus:
What's the reasoning behind a Windows build and a Windows wheel build?  I would like to limit the python builds per commit to 1 per OS if possible, these builds back up when many commits are raised in a short period and lead to multiple-hour delays due to lack of Azure slots.

I can change the publish script to require a passing full-build CI to prevent regressions being published, if thats the intent of testing both of these.

timkpaine:
Windows build followed by windows wheel takes > 60 minutes. We should check windows wheel on every commit, and during release we should run all the windows source builds and windows wheel builds (but can't be in same step cause it will time out).

I think windows builds are actually more available than Mac builds, so the two Mac builds per PR are a bigger issue

texodus:
Seems like the only changes which risk breaking Windows wheel build are C++ dependency changes, changes to the Python build process, and the occasional Azure or conda environment change;  all of which are relatively infrequent and are still verified by running a full build.

Granted, I've done a bad job of remembering to check a full build when approving PRs in these areas.  I'd like to explore wayts to make these exceptional checks befitting their rarity - if we cut 1 Windows build and 1 OSX build, we'll go from 6->4 jobs per PR and 2->3 average concurrent builds capacity.

timkpaine:
I've dropped to:
- 1 manylinux py3.7 + wheel
- 1 macOS 11 py3.9
- 1 windows py3.7 wheel

texodus:
Glorious

PR #1679, opened 2021-12-16
fix wheel build, make extension first to ensure binaries are inlined, then assemble wheel

timkpaine:
fixes #1677

PR #1680, opened 2021-12-20
Upgrade Arrow to `6.0.2`

texodus:
Rebased from #1636

Benchmarks look flat but its worth not letting these changes accumulate until we _need_ to upgrade.

Screen Shot 2021-12-19 at 7 15 01 PM


timkpaine:
Wasm pack stuff started breaking on windows (nothing on our end), it's in progress on #1679

texodus:
We do not need wasm-pack on Windows - just for wasm (obviously).  I'll remove this install step from azure-pipelines

timkpaine:
its separated into separate python and wasm init steps on that branch

PR #1688, opened 2021-12-28
Reduce CI: Turn off branch builds, only build on PRs to master

timkpaine:
reduce ci by only building on PRs to master branch, not building branches independent of PRs, and ommitting some paths

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -1,15 +1,29 @@
 trigger:
+  batch: true
   branches:
     include:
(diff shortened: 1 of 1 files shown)


texodus (approved):

My reading of the [`azure-pipelines.yaml` schema](https://docs.microsoft.com/en-us/azure/devops/pipelines/yaml-schema?view=azure-devops&tabs=schema%2Cparameter-schema) doesn't communicate clearly to me what `batch`, `autoCancel` and `drafts` do, but I suppose we're about to find out.

PR #1714, opened 2022-02-02
remove defunct IEX Cloud example - replace with Polygon.io stock example

timkpaine:
This example doesn't work well, replace it with https://github.com/timkpaine/polygon-io-perspective 
https://polygon.io

<br>

Screen Shot 2022-02-16 at 4 53 34 PM


texodus:

PR #1728, opened 2022-03-01
fix: fire config-update evnt on perspective-viewer

### LINKED ISSUE #1727: Column Styles Updates Do Not Fire `perspective-config-update` Event Correctly
### Steps to Reproduce:
 1. `cd examples; yarn start react-monaco`
 2. Open config panel 
 3. Change a column to be a **Bar** format 
 4. Notice that the config object in the editor on the left does not update

### Expected Result:

Updates to the data table column styles should fire a `perspective-config-update` event on the `perspective-viewer` element.

### Actual Result:

Looks like the `perspective-config-update` event is currently firing on the `perspective-viewer-data-grid` element, not the `perspective-viewer` element.

jkusa:
This PR resolves #1727 by firing the `perspective-config-update` event on `perspective-viewer` element.

This was tested manually. Is there a good place to add a test for this behavior?

### DIFF
--- a/packages/perspective-viewer-datagrid/src/js/plugin_menu.js
+++ b/packages/perspective-viewer-datagrid/src/js/plugin_menu.js
@@ -84,7 +84,7 @@ export function activate_plugin_menu(regularTable, target, column_max) {
         regularTable[PLUGIN_SYMBOL] = regularTable[PLUGIN_SYMBOL] || {};
         regularTable[PLUGIN_SYMBOL][column_name] = config;
         regularTable.draw({preserve_width: true});
-        regularTable.parentElement.dispatchEvent(
+        regularTable.parentElement.parentElement.dispatchEvent(
             new Event("perspective-config-update")
         );
     };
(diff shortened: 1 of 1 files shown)


texodus:

I've added a documented integration test for this in #1730

jkusa:
Thanks @texodus!

I actually thought of a small enhancement to this fix:

```js
const viewer = regularTable.closest("perspective-viewer");
viewer?.dispatchEvent(new Event("perspective-config-update"));
```

It is a little more specific on what it is trying to achieve and resilient to future DOM changes. I can push another PR or you can update it in #1730. Let me know what works best.

texodus:
I wouldn't worry about it unless it really bothers you :smile:

The perspective source is riddled with rigid DOM structure lookups of this sort, sometimes for performance when a known structure is in a tight loop, but often just out of convenience/laziness/"readability".  Personally, I prefer integration tests for behavior we want preserved, as opposed to scrutiny of the "robustness" of a particular implementation.

jkusa:
Sounds good :)

> Personally, I prefer integration tests for behavior we want preserved, as opposed to scrutiny of the "robustness" of a particular implementation.

 Yes 💯 Thanks for adding the test.

PR #1739, opened 2022-03-05
add two new aggtype "last minus first" and "high minus low".

shinny-yangyang:
In some use cases, I need to display "last value minus first value of each group", but I don't find a way to get it. So I add a new aggtype "last minus first" (I copied the code snippet from "last by index" and "first"). 

I also add another similar aggtype "high minus low" (code snippet from "sum"). 

I've done some smoke testing, they seem to work fine.


texodus:
Thanks for the PR @shinny-yangyang!

There are some issues with this, but I've made some fixes and opened a new PR #1742.  This includes your commit, as well as my fixes, rebased on master to avoid a recent CI failure (#1741) - I will add additional comments there covering what I've changed and why.

Closing in favor of #1742

PR #1741, opened 2022-03-06
Disable Windows builds

texodus:
Suspending Windows Python builds until [they can successfully build on CI](https://dev.azure.com/finosfoundation/perspective/_build/results?buildId=4036&view=logs&j=c9b54867-f028-5773-2b8c-07b236940a16).


timkpaine:
could be https://github.com/microsoft/vcpkg/pull/23067

PR #1742, opened 2022-03-06
New aggregates `last minus first` and `high minus low`

texodus:
This PR is a rebased, modified and tested version of #1739.  The major changes:
* The calculation `Xn - X0` for these aggregate types used the double-promoting `operator` implementations for `t_tscalar`, (which is required by ExprTK), however the output column type for these aggregates was not promoted, leading to all 0/`null` values.  This PR adds a new set of type-preserving methods `add_typesafe` and `sub_typesafe` that can be used safely with the new aggregates.
* `first_last_helper` method was being called twice, when internally this method does a column-scan and calculates both values for each pass.  This PR modifies the signature of `first_last_helper` to just return the tuple which can be decomposed by the caller, eliminating the extraneous iteration.  This aggregate should really use the `build_aggregate<>` pattern like `AGGTYPE_HIGH_WATER_MARK` does currently..
* Adds tests, fixed a typo, removed spurious `.gitignore` changes.


shinny-yangyang:
@texodus Thank you very much for your help. I still confused that is it necessary to add every aggtype to t_config::setup() (line 186 of config.cpp)? If so, why AGGTYPE_HIGH_WATERMARK and AGGTYPE_LOW_WATERMARK are not there?

PR #1746, opened 2022-03-08
export `perspective-viewer-datagrid`  plugin customElement and it's typings

zemeolotu:


texodus (changes requested):
Thanks for the PR!  Two issues to address:

* Run `yarn fix`.  There is a [failing CI check for lint](https://dev.azure.com/finosfoundation/perspective/_build/results?buildId=4053&view=logs&j=618c18e8-6013-5164-1629-abf655cc9818&t=1dbb40cb-c7fa-58c0-e6ab-0b58e0a56ee3) (and also a git hook - did you maybe `--no-verify` this?)
* Need to add the declarations for `customElements.get()` and `customElements.whenDefined()` [example for `get()` from `perspective-viewer`](https://github.com/finos/perspective/blob/master/rust/perspective-viewer/src/ts/extensions.ts#L34).  The custom-styles example uses this method anyway, which you'd need to otherwise cast.

texodus (approved):

texodus:
Looks good!

PR #1732, opened 2022-03-02
fix: fire config-update evnt on d3fc updates

jkusa:
This PR resolves #1729 and also fires the `perspective-config-update` event when updates are made to the d3fc legend position.

PR #1748, opened 2022-03-09
Fire `perspective-config-update` event on D3FC legend and axis events

### LINKED ISSUE #1729: Fire `perspective-config-update` on axis switch for multi-axis charts
### Description of Problem:

Currently, it does not seem that the dual axis configuration can be persisted and restored.

### Potential Solutions:

Add dual axis properties in the `plugin_config` section of the config object.

texodus:
Rebased and fixed-up version of #1732.  Fixes #1729.  Changes from #1732 

* For the legend-position event, I've changed this from `drag` to `end`.  For many apps which use `<perspective-viewer>`, the operations which listen for the `perspective-config-update` event may be relatively expensive, e.g. sending an event to a server, so we want to generally conserve these to applicable changes.  With this change, only 1 event is triggered when the drag operation completes, as opposed to 100s while the legend is moved.
* There is a cross-platform instability in `scatter.html` and `treemap.html` I think was introduced in #1722 - this PR does not modify these results (as validated in the reference HTML and PNG output I have from previous runs), so these `results.json` changes are reverted.
* The 2nd test is incorrect - it does not await `.restore()`, then clicks on the parent element of the axis label which has the click event listener.  The result is, the event for the axis switch is not fired, but the `.restore()` methods which is not awaited sometimes does, causing this test to flake.  Fixed by `await`-ing `restore()` and changing the click selector to pick the label itself.
```bash
  ● events.html › perspective-config-update event is fired when series axis is changed

    Node is either not clickable or not an HTMLElement

      59 |                         )
      60 |                     ).asElement();
    > 61 |                     await axisLabel.click(axisLabel);
         |                     ^
      62 |
      63 |                     const count = await page.evaluate(async (viewer) => {
      64 |                         // Await the plugin rendering
```


packages/perspective-viewer-d3fc/src/js/legend/styling/draggableComponent.js
```diff
@@ -39,12 +40,13 @@ export function draggableComponent() {
...
-                pinned = pinNodeToTopRight(node);
-                return;
-            }
+            pinned = isNodeInTopRight(node)
+                ? pinNodeToTopRight(node)
+                : unpinNodeFromTopRight(node, pinned);
+        });
 
-            pinned = unpinNodeFromTopRight(node, pinned);
+        drag.on("end", function (event) {
```
jkusa, line 48: 👍 Thanks for making this update. It did cross my mind that we probably want to fire the event after the user stops updating the position, but wasn't 100% sure how to do that here.

packages/perspective-viewer-d3fc/test/js/integration/events.spec.js
```diff
@@ -0,0 +1,146 @@
...
+            test.capture(
+                "perspective-config-update event is fired when legend position is changed",
+                async (page) => {
+                    // Await the viewer element to exist on the page
+                    const viewer = await page.$("perspective-viewer");
+                    await page.evaluate(async (viewer) => {
+                        // Await the table load
+                        await viewer.getTable();
+
+                        viewer.restore({
```
jkusa, line 90:
```suggestion
                        await viewer.restore({
```
Good catch on the test above. Probably want to do the same here.

texodus: fixed

texodus: Actually this was more nefarious than it appears - turns out that this test had a similar issue, `mouse.down()` was called to initiate the drag in the test, but not `mouse.up()` which fires the drag end event.  The test "worked" by asserting the `perspective-config-update` which was fired by `restore()`.

packages/perspective-viewer-d3fc/test/js/integration/events.spec.js
```diff
@@ -125,6 +125,7 @@ utils.with_server({}, () => {
                     await page.mouse.move(start.x, start.y);
                     await page.mouse.down();
                     await page.mouse.move(target.x, target.y);
+                    await page.mouse.up();
```
texodus, line 128: d'oh

PR #1753, opened 2022-03-10
docs: Fix broken links and an unclosed backtick typo

walker-finlay:


texodus (changes requested):
Thanks for the PR @walker-finlay!

The file you've edited is auto-generated by the TypeScript compiler and thus shouldn't be edited by hand (it will just get reset back when we next release).  These error should be fixed in the [`.ts` source files](https://github.com/finos/perspective/tree/master/rust/perspective-viewer/src/ts) instead, then you can run `yarn docs` from the project root to re-generate the `README.md`.

PR #1756, opened 2022-03-11
Fix datagrid to dist types and update `main` in package.json

zemeolotu:


texodus:
Thanks for the PR!  Discussed offline, we don't want `"main"` pointing to `"esm"` builds.  Closing in favor of #1759

PR #1764, opened 2022-03-14
fix group tree headers resize bug

zemeolotu:
This PR fixes a bug where tree group columns in a pivoted grid did not maintain any resize overrides between draws


texodus:
Thanks for the PR!

I think you've found the location of the bug, but the fix is off - this fix just shifts all column persistence off-by-one to the right, which looks like it fixes the bug on the first click, but shifts all subsequent clicks to the next column.  I've added a fix which memoizes this column headers to `__ROW_PATH__` instead in #1772 - closing this PR in favor of that one.

drift
.

PR #1767, opened 2022-03-14
Make toolbar icons configurable

kevinhinterlong:
Fixes alignment of theme dropdown and scroll lock/edit label

image
image

### DIFF
--- a/packages/perspective-viewer-datagrid/src/less/toolbar.less
+++ b/packages/perspective-viewer-datagrid/src/less/toolbar.less
@@ -36,11 +36,11 @@
 }
 
 #scroll_lock.lock-scroll:before {
-    content: "lock";
+    content: var(--toolbar-scroll-lock-active--content);
 }
 
 #scroll_lock:before {
-    content: "lock_open";
+    content: var(--toolbar-scroll-lock--content);
 }
 
 #select_mode:before {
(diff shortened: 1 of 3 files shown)


packages/perspective-viewer-datagrid/src/less/toolbar.less
```diff
@@ -87,6 +87,8 @@
 
 .button:hover {
     opacity: 1;
+    display: flex;
+    align-items: center;
```
texodus, line 91: What's the context for this?  The children of the elements this class is applied are the same height as the container so this should be inert - and indeed testing this locally against FF+Chrome+Safari and I see no difference?

texodus (commented):
Thanks for the PR!

Can you please share some details on the intent behind the `flex` changes?  These are inert in the current DOM structure, and I don't see any difference in your screenshots or my testing to `1.3.5`, so I expect that you're doing something unexpected (or unintended) with your custom styling.

packages/perspective-viewer-datagrid/src/less/toolbar.less
kevinhinterlong: This was to fix the alignment of the text/selectors in these buttons. Visiting https://perspective.finos.org/ on Firefox I'm able to reproduce this, but I just tested Chrome and Safari and I don't see the issue there

theme selector alignment
scroll lock alignment

jkusa: I am seeing the same as @kevinhinterlong on FF with MacOS.

texodus: Ok, I do see this on FF OSX 78.9, thanks for the clarification

texodus:

PR #1771, opened 2022-03-16
Remove extraneous `monaco-editor` features

texodus:


texodus:
There's no more `monaco-editor` in the project, so this can go

PR #1798, opened 2022-05-05
Support autocomplete for IN filter type

spff:


texodus (approved):

I may play around a bit with the UX in the future (would like to get a pillbox UX), but this is already a vast improvement over the current version.

PR #1807, opened 2022-05-13
Run `yarn fix` on codebase

timkpaine:


texodus:
Thanks for the PR!

These lint fixes are unstable, their outputs are inconsistent (relative to your local dev environment) - hence why they are not enforced by the git hook.  I left them in for occasions when you've made large edits to one of these filetypes (e.g. `.cpp`) and want a quick cleanup, but I don't want to merge pure `fix` applications because they will just get reverted when someone else updates.  I'm open to removing them entirely if this is too confusing.

PR #1808, opened 2022-05-14
Move CI/CD from Azure Pipelines to GitHub Actions

timkpaine:
This PR replaces the Perspective CI/CD infrastructure. We are currently using Azure Pipelines, and this PR replaces that with GitHub actions

# Motivations
Azure pipelines are very similar to GitHub Actions (coming from the same company), both with a number of important distinctions

- Separate interface, login, controls, etc, in Azure Devops instead of natively in GitHub
- More difficult pipeline syntax
- Smaller / fewer community extensions

Additionally, anecdotal evidence suggest that the machines and pipelines used for azure pipelines are slower than for GitHub actions, and Perspective's complex build process (which is becoming more complex due to ongoing JS / Python integration work) leads to long build times.

After some initial exploration, the easier-to-grok syntax and anecdotally faster speed of GitHub Actions was enough to motivate experiments for migrating.

# Requirements

Perspective builds with the following languages and technologies:
- node for scripting / project orchestration
- emscripten emsdk for C++ compilation to webassembly
- C/C++ compilers for python integration
- Python + dependencies
- rust for frontend components via webassembly

Because perspective builds its python assets for deploy on azure, we need to support the following build matrix
- OS X 10.15
- Windows 2019
- Any linux (linux packages built via manylinux docker images which provide python 3.6-3.9)
- Python 3.6-3.9 (mac)
- Python 3.7-3.9 (windows) 

All required technologies, operating systems, compilers, environments are supported for GitHub actions.

# Build assets

Azure and this PR both emit the following wheel assets

- cp36-cp36m-macosx_10_14_x86_64
- cp37-cp37m-macosx_10_15_x86_64
- cp38-cp38-macosx_10_15_x86_64
- cp39-cp39-macosx_10_15_x86_64
- cp36-cp36m-manylinux2010_x86_64
- cp37-cp37m-manylinux2010_x86_64
- cp38-cp38-manylinux2010_x86_64
- cp39-cp39-manylinux2010_x86_64
- cp36-cp36m-manylinux2014_x86_64
- cp37-cp37m-manylinux2014_x86_64
- cp38-cp38-manylinux2014_x86_64
- cp39-cp39-manylinux2014_x86_64
- cp37-cp37m-win_amd64
- cp38-cp38-win_amd64
- cp39-cp39-win_amd64

# Caching
Caching is easier to configure on GitHub Actions vs Azure. However, it has caused some problems before so we may need to revisit it.

- CCache for C++ compilation
- Node cache for node_modules
- Pip cache for python deps
- rust/cargo cache for rust deps
- homebrew cache for mac deps
- vcpkg for windows C++ deps

# Build Flags / Partial Builds
On Azure, we have overridable flags to check if the build is from a git tag (e.g. something starting with `v`), so that we can performa a full build. This flag is overridable so that manual full builds can be run.

This PR provides the following set of flags, which can be specified via:
- commit message
- PR title
- manual workflow run

- `[ci-full]` run a full build across all environments. This is the default when a git tag starts with `v` for releases
- `[ci-skip]` no CI, e.g. for docs or non-build configuration
- `[ci-skip-python]` for pure JS changes, reduces the build time by roughly 50%
- `[ci-include-windows]` include the windows build, increases the build time by roughly 30%
- `[ci-skip-cache]` omits all caching, which has caused problems in the past on azure. Increases the build time by roughly 20%

Helpful scripts have been added to make empty commits with these messages:
```
        "ci:full": "git commit --allow-empty -m \"[ci-full]\"",
        "ci:skip-python": "git commit --allow-empty -m \"[ci-skip-python]\"",
        "ci:include-windows": "git commit --allow-empty -m \"[ci-include-windows]\"",
        "ci:skip-cache": "git commit --allow-empty -m \"[ci-skip-cache]\"",
```

To run a workflow manually, the syntax looks like this (via `gh` command line)
```
gh workflow run "Build Status" --ref tkp/gha  -f ci-full=true -f ci-skip-cache=false  -f ci-skip-python=false -f ci-include-windows=true
```

This runs the `tkp/gha` branch with the provided set of flags

# Tests performed on this PR
- [x] Normal CI build
- [x] Skip CI
- [x] Full CI Build
- [x] Skip Python
- [x] Include Windows
- [x] Skip Cache

See the comments for more information


timkpaine:
Normal CI build passing, time to test the flags

timkpaine:
Skip python

Screen Shot 2022-06-06 at 4 24 46 PM

timkpaine:
Skip CI

Screen Shot 2022-06-06 at 4 40 33 PM

timkpaine:
Include windows:
Screen Shot 2022-06-07 at 12 16 27 AM

timkpaine:
Full:
Screen Shot 2022-06-07 at 10 31 22 AM

timkpaine:
Full assets:
Screen Shot 2022-06-07 at 10 31 59 AM


Grouped contents:
Screen Shot 2022-06-07 at 10 32 32 AM

timkpaine:
Skip Cache:
Screen Shot 2022-06-07 at 10 07 23 PM

timkpaine:
This is code complete from a CI perspective, still need to rework the publish script

timkpaine:
Manually triggered run https://github.com/finos/perspective/actions/runs/2476949750

README.md
```diff
@@ -3,7 +3,7 @@
 
 [![npm](https://img.shields.io/npm/v/@finos/perspective.svg?style=flat)](https://www.npmjs.com/package/@finos/perspective)
 [![PyPI](https://img.shields.io/pypi/v/perspective-python.svg?style=flat)](https://pypi.python.org/pypi/perspective-python)
-[![Build Status](https://dev.azure.com/finosfoundation/perspective/_apis/build/status/finos.perspective?branchName=master)](https://dev.azure.com/finosfoundation/perspective/_build/latest?definitionId=1&branchName=master)
+[![Build Status](https://github.com/finos/perspective/workflows/Build%20Status/badge.svg?branch=master)](https://github.com/finos/perspective/actions?query=workflow%3A%22Build+Status%22)
```
texodus, line 6: GHA may not become the default in this PR - let's split this (and anything else so marked below) in a "move GHA to prod" PR

azure-pipelines.yml
```diff
@@ -1,612 +0,0 @@
-trigger:
```
texodus, line 1: As above, azure needs to stay and work in this PR

docker/python/manylinux2010/Dockerfile
```diff
@@ -6,8 +6,7 @@
 # the Apache License 2.0.  The full license can be found in the LICENSE file.
 #
 
-# https://github.com/pypa/manylinux/issues/428#issuecomment-774729597
-FROM quay.io/pypa/manylinux2010_x86_64:2021-02-06-3d322a5
+FROM quay.io/pypa/manylinux2010_x86_64
```
texodus, line 9: As per #1820 and others, I want to avoid these changes until they're reproducible or removable.

packages/perspective-jupyterlab/test/config/jupyter/jlab_start.js
```diff
@@ -18,7 +18,11 @@ const PACKAGE_ROOT = path.join(__dirname, "..", "..", "..");
  */
 const kill_jlab = () => {
     console.log("-- Cleaning up Jupyterlab process");
-    execute`ps aux | grep -i '[j]upyter-lab --no-browser' | awk '{print $2}' | xargs kill -9 && echo "[perspective-jupyterlab] JupyterLab process terminated"`;
+    try {
+        execute`ps aux | grep -i '[j]upyter-lab --no-browser' | awk '{print $2}' | xargs kill -9 && echo "[perspective-jupyterlab] JupyterLab process terminated"`;
+    } catch {
+        // ignore
+    }
```
texodus, line 25: `execute()` does not throw (`execute_throw()` does, 'nat), this looks like debug code that snuck in

packages/perspective-jupyterlab/test/config/jupyter/jlab_start.js
```diff
@@ -105,6 +109,7 @@ exports.start_jlab = function () {
             return proc;
         });
     } catch (e) {
+        console.error("Error during setup");
```
texodus, line 112: debug code

packages/perspective-jupyterlab/test/config/jupyter/kernels/python3/kernel.json
```diff
@@ -0,0 +1,14 @@
+{
```
texodus, line 1: ???

Also the following two python icons - these are necessary additions to this PR?

packages/perspective-jupyterlab/test/jupyter/utils.js
```diff
@@ -73,7 +73,10 @@ describe.jupyter = (body, {name, root} = {}) => {
     // Remove the automatically generated workspaces directory, as it
     // will try to redirect single-document URLs to the last URL opened.
     beforeEach(remove_jupyter_artifacts);
-    afterAll(remove_jupyter_artifacts);
+    afterAll(() => {
+        console.warn("Normal teardown");
+        remove_jupyter_artifacts();
+    });
```
texodus, line 79: debug code

packages/perspective-jupyterlab/test/jupyter/utils.js
```diff
@@ -109,15 +112,23 @@ module.exports = {
...
+            const button = await page.waitForSelector(
+                "div.jp-Dialog-content > div.jp-Dialog-footer > button.jp-mod-accept",
+                {timeout: 5000}
+            );
+            if (button) {
+                button.click();
+            }
+        } catch {
+            // ignore
+        }
```
texodus, line 126: Why does this revert #1823?  Does this not happen because the notebook kernel does not match the new one added above?

packages/perspective-jupyterlab/test/jupyter/utils.js
```diff
@@ -109,15 +112,23 @@ module.exports = {
...
 
+        // wait for a cell to be active
         await page.waitForSelector(
             '.jp-Notebook-ExecutionIndicator[data-status="idle"]',
-            {timeout: 3000}
+            {timeout: 15000}
```
texodus, line 131: Debug code (I hope ...)

packages/perspective-jupyterlab/test/jupyter/widget.spec.js
```diff
@@ -12,16 +12,20 @@ const utils = require("@finos/perspective-test");
...
     await execute_all_cells(page);
+    console.warn(2);
     const viewer = await page.waitForSelector(
         ".jp-OutputArea-output perspective-viewer",
-        {visible: true, timeout: 30000}
+        {visible: true, timeout: 240000}
     );
+    console.warn(3);
     await viewer.evaluate(async (viewer) => await viewer.flush());
+    console.warn(4);
```
texodus, line 24: Lots of debug code and a 8x timeout inflation ...

packages/perspective-jupyterlab/test/jupyter/widget.spec.js
```diff
@@ -12,16 +12,20 @@ const utils = require("@finos/perspective-test");
...
     return viewer;
 };
 
-jest.setTimeout(30000);
+jest.setTimeout(240000);
```
texodus, line 28: timeout inflation

python/perspective/perspective/tests/table/test_table_arrow.py
```diff
@@ -22,14 +23,22 @@
...
+        if not os.path.exists(SUPERSTORE_ARROW):
+            file_path = SUPERSTORE_ARROW_MINI
+            size = 10
+            width = 19
+        else:
+            file_path = SUPERSTORE_ARROW
+            size = 9994
+            width = 21
+
+        with open(file_path, mode='rb') as file:  # b is important -> binary
```
texodus, line 36: This is because `yarn install` does not run on python builds?  It looks like in the `build.yml` it does though.

I think there are better ways to avoid this - let's either find a way to get `superstore.arrow` in the python build (as we may already and this is debug code again), or just use a small superstore-mini ... though I really would prefer we re-use the asset we published to NPM explicitly for this purpose.

scripts/install_emsdk.js
```diff
@@ -52,6 +52,8 @@ function check() {
     }
 }
 
-if (!check()) {
-    upgrade();
+if (!process.env.PSP_SKIP_EMSDK_INSTALL) {
```
texodus, line 55: I understand you don't want to always auto-install emsdk as a side effect of other commands, but putting this check here disables installation globally including when it is set in `.perspectiverc`.   Can we do this where `install_emsdk` is invoked as a side effect instead (presumably `post_install`?)

I'd like to say no developer is stupid enough to get confused by this, but I just was, for an embarrassingly long time.

scripts/install_tools.js
```diff
@@ -18,16 +18,16 @@ try {
...
+        // console.log("-- Installing Flatbuffers");
+        // execute`mkdir -p /usr/local`;
+        // process.chdir("/usr/local");
+        // execute`git clone --depth 1 --branch v1.12.1 https://github.com/google/flatbuffers.git`;
+        // process.chdir("/usr/local/flatbuffers");
+        // execute`cmake -G "Unix Makefiles"`;
+        // execute`make`;
+        // execute`cp -r /usr/local/flatbuffers/include/flatbuffers /usr/local/include`;
+        // execute`ln -s /usr/local/flatbuffers/flatc /usr/local/bin/flatc`;
+        // execute`chmod +x /usr/local/flatbuffers/flatc`;
```
texodus, line 30: Delete >>> Comment out

scripts/publish_python.js
```diff
@@ -8,125 +8,258 @@
  */
 
 const {execute} = require("./script_utils.js");
-const fs = require("fs");
```
texodus, line 11: This is great but I need this script until we are certain GHA stabilizes (after this PR is merged).  Can we please also move this to the "GHA stabilization PR".

scripts/test_python.js
```diff
@@ -75,17 +68,13 @@ const pytest_client_mode = (IS_DOCKER) => {
 const pytest = (IS_DOCKER) => {
     if (IS_DOCKER) {
         return bash`${docker(IMAGE)} bash -c "cd \
-            python/perspective && TZ=UTC PYTHONMALLOC=debug ${PYTHON} -m pytest \
```
texodus, line 78: I like `PYTHONMALLOC=debug` for tests  - was this remove don purpose?

scripts/test_js.js
```diff
@@ -115,6 +115,7 @@ try {
     if (!IS_JUPYTER) {
         // test:build irrelevant for jupyter tests
         execute`node_modules/.bin/lerna run test:build --stream --scope="@finos/${PACKAGE}"`;
+        return;
```
texodus, line 118: This skips the JS tests???

texodus (changes requested):
Thanks for the PR!

I would like to merge this but it cannot be the default CI in this PR.  I propose this be split into 2 PRs, one with just what is needed to enable GHA for builds, and a second which promotes this to the default;  I've marked some places I believe should go in the latter above.

There is some minor debug/logging code that should be cleaned up or justified, and some test condition changes and "timeout inflation" that looks like leftover debug code I'd like removed or traced (to my knowledge there are no races in the existing jlab test code, I spent quite some time on this issue in #1823 and others).

As discussed, I'll give in on the directives in commit messages due to necessity, on the agreement that they not be used in open PRs unless startlingly necessary, and that they _never_ be exclusively-directive commits (e.g. `--allow-empty` commits).

scripts/install_tools.js
timkpaine: https://github.com/finos/perspective/commit/1ec65f0388e1ce3f8b6f7e67f5142613a0cc5af0

I wasn't sure if we were keeping it around for some reason

packages/perspective-jupyterlab/test/jupyter/utils.js
timkpaine: the notebook kernel specified in our original jupyter tests can fail to match the kernel on the machine, especially if they change the name of the default kernel (as they have recently).

scripts/install_emsdk.js
timkpaine: If you set `PSP_SKIP_EMSDK_INSTALL`, a very specific env var, you will not install `emsdk`. Otherwise it will operate as before.

scripts/install_tools.js
timkpaine: ive deleted this, and all the other commented flatbuffers references.

packages/perspective-jupyterlab/test/jupyter/utils.js
timkpaine: This is max wait time, too short and you get spurious failures, too long and you wait too long before real failures.

timkpaine:
sliced binder changes over to https://github.com/finos/perspective/pull/1855

timkpaine:
flatbuffers changes consolidated in https://github.com/finos/perspective/pull/1820/files

python/perspective/perspective/tests/table/test_table_arrow.py
timkpaine: I don't think python tests should require a developer installation (e.g. `../../../../node_modules` isn't going to work on somebody who just downloads perspective-python) so i've eliminated the node_modules superstore in favor of the local one

scripts/build_python_azure.js
```diff
@@ -0,0 +1,121 @@
+/******************************************************************************
```
timkpaine, line 1: this file has no modifications from current `master` `build_python.js`

scripts/publish_python_azure.js
```diff
@@ -0,0 +1,132 @@
+/******************************************************************************
```
timkpaine, line 1: this file has no modifications from current `master` `publish_python.js`

scripts/install_emsdk.js
texodus: I cited an explicit example of how this fails (env vars injected through `.perspectiverc`), and it is [so easy to fix](https://docs.npmjs.com/cli/v8/commands/npm-install#ignore-scripts)

texodus:
[full build](https://dev.azure.com/finosfoundation/perspective/_build/results?buildId=4516&view=results)

python/perspective/perspective/tests/table/test_table_arrow.py
texodus: Running tests from a e.g. pypi install of `perspective-python` is not a design goal.

timkpaine: It should be, this is how e.g. conda would test that it was built properly, how how one might test the compatibility of their perspective installation with the rest of their install

texodus:
As per discussion offline, I've force pushed this branch with a real commit message.  Additional changes I've made

* Removed `README.md` badge (this is not stable yet and I don't want a scary red warning on our docs site).
* Removed kernel metadata and kernel selection bypass in test utils.  I suspect strongly this was hand-tested against a python environment that is inconsistent with out development/CI structure, as the default kernel name is not apt to change between CI runs ... if this work has merit, it should be able to be justified on its own PR.
* Removed Jupyter test changes.  Further removed condition timeouts entirely.  Despite public outcry, this seems to work just fine.
* Removed OSX and sdist builds from default GHA runs.  sdist specifically is the slowest build of them all, and IMO the lowest risk for failure.
* Revert jupyter widgets version bump - I don't want to guess why timing and/or correctness issues happen on this branch so there should not be _any_ superfluous code changes.
* Removed fake superstore arrow test entirely.  Merits of the sdist testing aside, we need _fewer_ mystery blobs in this repo, not more, and this test is not worth fighting over.

[full gha](https://github.com/finos/perspective/actions/runs/2493093872)
[full azure](https://dev.azure.com/finosfoundation/perspective/_build/results?buildId=4523&view=results)

timkpaine:
Note that the kernel **can** change between CI runs as the valid kernels are specified by the installation of jupyter, not by anything in this package.

PR #1810, opened 2022-05-14
normalize some commands

timkpaine:
Normalizes the commands. There was a mixture of e.g. `lint_js` and `fix:js` so this PR just aliases a few `:` commands to their corresponding `_` variants


texodus:
Thanks for the PR!

1) This was intentional.  `:` in `package.json` indicate task groups and are "internal details" of their top-level task (annoyingly, `package.json` does not support full shell scripts).  To differentiate from these tasks which shouldn't be called from `yarn` directly, we use `_` for simlar tasks that aren't meant to be run as a group, e.g. `build_js` or `build_python`.  I don't want to remove this distinction, there build scripting is already well out-of-hand.
2) This PR adds aliases rather than removing duplicates, which in general does not outweigh the cost of the churn.  I would merge a PR that removes/simplifies these project wide though ..

PR #1811, opened 2022-05-14
prune 2 more python2 references

timkpaine:


timkpaine:
Builds aren't going to run for this since its ignored, so feel free to merge without waiting:
https://github.com/finos/perspective/blob/master/azure-pipelines.yml#L25

texodus (approved):

PR #1813, opened 2022-05-15
Trying a few speedups and caching for CI

timkpaine:


timkpaine:
Closing as the speedup is only marginal.

PR #1820, opened 2022-05-20
remove flatbuffers steps

timkpaine:


texodus:
Thanks for the PR!

I'm going to move this to a Draft because:

1) I am not going to re-build these images, it is annoying
1a) There is a decent chance they will never be built again, pending GHA port.
2) I don't want the master images and master GitHub docker source to diverge.

If we do end up re-building (rather than deleting) these images in the future, I will re-open.

timkpaine:
@texodus adding to a later milestone filed under "remove necessity of maintaining docker images"

PR #1799, opened 2022-05-06
Distribute prebuilt JupyterLab extension / Notebook Classic extension

timkpaine:

PR #1824, opened 2022-05-26
Distribute prebuilt JupyterLab extension / Notebook Classic extension #1799

texodus:


.husky/pre-push
```diff
@@ -1,4 +1,4 @@
 #!/bin/sh
 . "$(dirname "$0")/_/husky.sh"
 
-yarn prepush
+#  yarn prepush
```
texodus, line 4: lol

cpp/perspective/CMakeLists.txt
```diff
@@ -125,6 +125,7 @@ option(CMAKE_BUILD_TYPE "Release/Debug build" RELEASE)
 option(PSP_WASM_BUILD "Build the WebAssembly Project" ON)
 option(PSP_CPP_BUILD "Build the C++ Project" OFF)
 option(PSP_PYTHON_BUILD "Build the Python Bindings" OFF)
+option(PSP_PYTHON_BUILD_OMIT_BINDING "Build the C++ project, omit the python. FOR CI ONLY" OFF)
```
texodus, line 128: What are these `cmake` changes for?  I don't see them used

package.json
```diff
@@ -61,13 +61,13 @@
...
-        "mini-css-extract-plugin": "^1.3.4",
+        "mini-css-extract-plugin": "^1.3.9",
         "minimist": "^1.2.0",
         "mkdirp": "^0.5.1",
         "npm-pack-here": "^1.2.0",
         "npm-run-all": "^4.1.3",
-        "postcss": "^7.0.19",
-        "postcss-loader": "^4.3.0",
+        "postcss": "^8.4.13",
+        "postcss-loader": "^6.2.1",
```
texodus, line 70: These tools have many dependencies themselves and cause downstream headaches for integrators using `webpack`, yet another reason to not introduce `webpack` for this build

packages/perspective-jupyter/package.json
```diff
@@ -19,15 +19,23 @@
...
+        "build:base": "node build.js",
+        "build:lib": "babel src/js/ --source-maps --out-dir dist/esm/",
+        "build:labextension": "jupyter labextension build . --debug",
+        "build:nbextension": "webpack --mode=production -c src/config/webpack.config.js",
+        "clean": "npm-run-all clean:*",
+        "clean:dist": "rimraf dist",
+        "clean:lib": "rimraf lib",
         "clean:screenshots": "rimraf \"screenshots/**/*.@(failed|diff).png\"",
+        "clean:labextension": "rimraf ../../python/perspective/perspective/labextension",
+        "clean:nbextension": "rimraf ../../python/perspective/perspective/nbextension/static",
```
texodus, line 32: These tasks should go in `python/perspective/package.json`, yes?

packages/perspective-jupyter/src/less/index.less
```diff
@@ -15,11 +15,22 @@ div.PSPContainer {
...
 // Widget height for Voila
 .jp-OutputArea-output div.PSPContainer {
     height: 520px;
 }
 
+// Widget height for Jupyter Notebook
+.jupyter-widgets-view div.PSPContainer {
+    height: 520px;
+}
+
```
texodus, line 33: ???

packages/perspective/src/js/perspective.browser.js
```diff
@@ -53,6 +53,10 @@ const _override = /* @__PURE__ */ (function () {
                         }
                     });
 
+                    if (_wasm instanceof ArrayBuffer && !_wasm.buffer) {
+                        _wasm = new Uint8Array(_wasm);
+                    }
+
```
texodus, line 59: I would prefer these here rather than #1800/#1801 definitely, so I'll close these

texodus (commented):
Thanks for the PR @timkpaine, this is a monstrous effort that is much appreciated.  I'm excited for the zero-config `perspective-python`/Jupyter integration future.

A couple of things stick out that I'd like you to elaborate on -

* I understand the reasoning behind renaming `@finos/perspective-jupyterlab` -> `@finos/perspective-jupyter`, but not the impact it will have on existing Jupyter users with the former already installed.  IMO this is _really important_, as the nature of this PR is to simplify installation, we don't want to co-introduce _arbitrary_ regressions.  If you think this is really important to stay, we should add correct upper bounds to the old package and publish it (rather than delete this package),  so that current users of the old package do not suddenly get deprecation errors on update, but rather a clear version-bound error saying that `@finos/perspective-jupyterlab` is incompat with the `perspective-python` bounds.  I believe (?) we can set the NPM deprecation message on that version to point to the new package.
* Please, please, for my own sanity - do not introduce `webpack` back into our build process, especially not to generate a full inline build _twice_.  I _finally_ managed to remove `webpack` entirely in favor of `esbuild`.  Even for a less-frequently-updated package like this, it is disruptive in the extreme, and Jupyter itself already imposes its own build tax on our precious development time.  I need to dig in a bit more (if you want me to), but I see no reason this can't be built in `esbuild` with the existing plugins in worst case;  but what would likely be even easier is to just import the pre-built inline builds of `@finos/perspective` and `@finos/perspective-viewer` directly, then `esbuild` these.  As they are pure JavaScript already, you'll need no special plugins, get an almost identical bundle, at 1000x the build performance.  The only quirk I see is specifying `externals` correctly?

packages/perspective-jupyter/src/less/index.less
timkpaine: Different CSS rules in notebook so this change is symmetric to the lab CSS rules next to it in the same file.

packages/perspective-jupyter/package.json
timkpaine: Either, I think it's better here since this is the file that actually generates those assets. A little weird if you `yarn clean` from this folder and it doesn't actually clean the stuff you want it to clean

cpp/perspective/CMakeLists.txt
timkpaine: I will probably undo these now that the build is so fast, this enables you to split the C++ build between the python version agnostic `libpsp` and the python version dependent `libbinding`

timkpaine:
Now that it all works from one package we should call it `jupyter` but we can call it `jupyterlab` for now until we get to e.g. 2.0

For webpack, the two things are:
- prebuilt extensions rely on webpack module federation, which requires webpack as far as I can tell. I will note that I don't introduce a webpack dependency directly for this (as we use `jupyter labextension build` which has the dep)
- Notebook extensions require `amd` builds which are not supported by esbuild, and the issue discussing this on the esbuild repo has suggestions which did not work

timkpaine:
closing for bad rebase, will move to a separate branch

.husky/pre-push
timkpaine: when you open a new PR you lose [context](https://github.com/finos/perspective/pull/1799/files#r867238799)

PR #1827, opened 2022-05-29
remove separate mock dependency, now part of unittest

timkpaine:
Just a small PR top remove the separate `mock` dependency.


timkpaine:
(no need to rebuild docker images)

texodus:
Thanks for the PR!

Holding for now, same reason as #1820 (don't want `DockerFile` to drift from disted images).

PR #1829, opened 2022-06-01
fix bug in setup.js where extra set of quotes prevents initial build

timkpaine:
Fixes a bug in setup.js where extra set of quotes in the execution command prevents initial build (tries to execute contents of the double quotes as a single executable). 

Before:
Screen Shot 2022-06-01 at 1 54 48 PM

After:
Screen Shot 2022-06-01 at 1 56 37 PM


texodus:

Please save the `bug` tag for user facing issues we want to highlight in the `Fixed` section of the release's `CHANGELOG.md`, and also be sure not to mix [`internal`, `bug` and `enhancement` tags](https://github.com/finos/perspective/blob/master/package.json#L7) or you'll create duplicates

Issue #1831, opened 2022-06-03
Computed Column / aggregation inconsistency between operating systems (race condition?)


timkpaine:
### Steps to Reproduce:

The following code block yields different results on mac and linux, on latest perspective version:

```python
from perspective import Table

t = Table({"id": int, "symbol": str, "valid": bool, "value": int}, index="id")
t.update([{"id": 1, "symbol": "A", "valid": False,  "value": 5},
          {"id": 2, "symbol": "A", "valid": True, "value": 10}])

v = t.view(
    columns=["symbol", "value", "valid_value"],
    expressions=["""//valid_value\nfloat("valid" ? "value" : 0)"""]
)

v_agg = t.view(
    columns=["symbol", "value", "valid_value"],
    group_by=["symbol"],
    aggregates={"symbol": "first", "value": "sum", "valid_value": "sum"},
    expressions=["""//valid_value\nfloat("valid" ? "value" : 0)"""]
)


df_flat = v.to_df()
df_agg = v_agg.to_df()
assert df_flat.iloc[-1]["valid_value"] == df_agg.iloc[-1]["valid_value"]

t.update([{"id": 2, "symbol": "A", "valid": False, "value": 10}])

df_flat = v.to_df()
df_agg = v_agg.to_df()

# Fails on linux
assert df_flat.iloc[-1]["valid_value"] == df_agg.iloc[-1]["valid_value"]
```

### Expected Result:

Before the update, row id `2` has `valid: True`, so the computed column should emit value `10.0` for `valid_value`.
Both the flat view and aggregated view should have a row like this:

|symbol|value|valid_value|
|:-:|:-:|:-:|
|A|10|10.0|

After the update, we set `valid` to `False` for this row, so the `valid_value` should be driven to `0`.

|symbol|value|valid_value|
|:-:|:-:|:-:|
|A|10|0|

### Actual Result:
This works as expected on Mac (tested on big sur 11.6.6 perspective 1.3.5), fails on linux (tested on centos 7.9.2009), tested with python 3.9 in both environments.

On the provided linux environment, the aggregated view has result:

|symbol|value|valid_value|
|:-:|:-:|:-:|
|A|10|10.0|

Furthermore, if you continue to update `value`, the computed `valid_value` will not update.

### Environment:
OS: Mac 11.6.6 / Centos 7.9
Python: 3.9
Perspective: 1.3.5 (latest)

timkpaine:
[...]

PR #1835, opened 2022-06-04
Remove parallel for from data table to fix computed column threading issue in #1831

timkpaine:


cpp/perspective/src/include/perspective/gnode.h
```diff
@@ -208,7 +208,7 @@ class PERSPECTIVE_EXPORT t_gnode {
     std::shared_ptr<t_expression_vocab> get_expression_vocab() const;
     std::shared_ptr<t_regex_mapping> get_expression_regex_mapping() const;
 
-#ifdef PSP_PARALLEL_FOR
+#ifdef PSP_ENABLE_PYTHON
```
timkpaine, line 211: this is unrelated but should be enabled conditional on PYTHON, not PARALLEL_FOR

timkpaine:
@texodus might need your help benchmarking this, if it causes a substantial degradation in performance I can try to break this loop apart to reenable some part in parallel

texodus (changes requested):
It's not clear to my why this change would impact the test provided, so on a hunch I forked just this test to a new branch and it [passes just the same on all Linux builds](https://dev.azure.com/finosfoundation/perspective/_build/results?buildId=4456&view=logs&j=9c1f2f8e-4d43-59a0-c5f5-4a00727b21e2&t=19e3a096-6506-57cc-343b-951c3932e971&l=1558).  What am I missing?

timkpaine:
This test fails on my Linux machine, validated on 2 other Linux machines

timkpaine:
See #1835 for more details

timkpaine:
My guess would be either azure or the manylinux images themselves don't trigger the test case to fail.

timkpaine:
Some additional testing suggests this might've been fixed already via the arrow upgrade https://github.com/finos/perspective/pull/1816

texodus:
So thanks to the accompanying test (which I _do_ want to merge), I can confirm that this issue was resolved by #1793, and is unrelated to Arrow/threading.

PR #1844, opened 2022-06-08
fix bad import in webpack example

timkpaine:
CC https://github.com/finos/perspective/discussions/1842

presumably theres some way to just pull in bar but I figured this fixes it and thats not a super important feature to exhibit in this example.


texodus:
CLosing in favor of #1847

PR #1848, opened 2022-06-10
Expression Docs

brianblakely:
New docs for ExprTK-based expression engine.

NOTES:
1. A number of functions that I was not able to use and have marked with TODOs. Perhaps some guidance can resolve this.
2. Similarly, looping mechanisms that exist in ExprTK and which I could not confirm/deny in Perspective have also been marked with TODOs. Suggestions welcome!

Thank you very much!


docs/md/expressions.md
```diff
@@ -0,0 +1,429 @@
...
+```
+
+<div>
+<perspective-viewer
+  columns='["Customer Name", "Sales", "Profit", "My Column Name"]'
+  expressions='["// My Column Name\n\n0 // REPLACE ALL AND SAVE"]'
+></perspective-viewer>
+</div>
+
+### Strings (Text)
```
texodus, line 139: I appreciate the amount of work this section likely took and there's definitely some good imrpovements to the prose compared to the [auto-generated version I just merged](https://github.com/finos/perspective/blob/master/rust/perspective-viewer/exprtk.md).  However, I'd like the content here to come from the [metadata in the rust src](https://github.com/finos/perspective/blob/master/rust/perspective-viewer/src/rust/exprtk/language.rs#L63), rather than embedded directly in this doc, because:

1) This project is huge and we need to be diligent in auto-generating consistency tasks like this or I will lose my mind.
2) We can assert as a compile-time check that docs exist for new functions that are added or old functions that are changed.
3) I want this information in the intellisense highlights in the actual `monaco` IDE also, and I don't want to have to remember to repeat it in multiple locations.

Any of this content you'd like to move to the metadata (including and especially the examples) would be great

docs/md/expressions.md
```diff
@@ -0,0 +1,429 @@
...
+#### Bucketing by Date
+
+`bucket("Order Date", 'D')` - Produces date from datetime.
+`bucket("Order Date", 'W')` - Produces date from the first day of that week.
+`bucket("Order Date", 'M')` - Produces date from the first day of that month.
+`bucket("Order Date", 'Y')` - Produces date from the first day of that year.
+
+## Conditional Structures
+
+Acting based on conditions is similar to most programming languages.
```
texodus, line 239: Core ExprTK - may want to look at how the ExprTK docs describe this feature also as I think it's a bit clearer, especially if you're coming form an Excel background rather than a Python (or other programming) background.

docs/md/expressions.md
```diff
@@ -0,0 +1,429 @@
...
+<!-- TODO: Could not make this work. -->
+<!-- ### `continue` -->
+
+## Advanced Example
+
+This **[Mandelbrot
+demo](https://bl.ocks.org/texodus/5485f6b630b08d38218822e507f09f21)** calculates
+the popular fractal in a Column Expression and displays it with a heatmap. Edit
+the `color` column to view and manipulate the expression, whose contents are
+below:
```
texodus, line 400: I really like the idea of having more complext PerspExprTK examples in the docs, but this one is both a little complicated and lacking all of the detail needed to actually implement it (e.g. source `Table` structure, X/Y axis calculations).

docs/utils/expr.js
```diff
@@ -0,0 +1,23 @@
+// Quick one-off utility to convert a Column Expression into a single-line
```
texodus, line 1: The easiest way to do this is to click the `Copy` button in the `<perspective-viewer>` status bar, and select `.config.json` option to copy the current `<perspective-viewer>` config to the clipboard as JSON.

Regardless, this file shouldn't be in the PR.

docs/md/expressions.md
```diff
@@ -0,0 +1,429 @@
...
+```html
+// My Column Name
+```
+
+### Editing or Deleting a Custom Column
+
+Once a Custom Column is added, the expression editor can be opened by pressing
+the hamburger button beside the column name. The column may be deleted by
+pressing the X button, but to do so, it must first be removed from the data view
+by unchecking the box beside it.
```
texodus, line 38: I would rephrase this ala "... Only `expressions` which are currently unused may be deleted"?

I would also mention that `expressions` which are unused are not calculated (and therefore don't impact performance at all), and that the `Reset` button does not delete `expressions` unless the `Shift` key is held down when it is clicked.

Maybe this section should also be moved immediately after `### Creating a Custom Column`, as it pertains to the interface rather than the syntax?

texodus (changes requested):
Thanks for the PR @brianblakely!  It's a much-needed addition and much-appreciated contribution!

The content here is good!   It covers the important points and there are some really nice example formulas especially.   I'd like to see the structure updated a bit to make it clearer to read, and easier for us to maintain, namely:

1) Let's rely-on/defer-to the [ExprTK docs](https://github.com/ArashPartow/exprtk/blob/master/readme.txt) as much as possible for core functionality coverage, as it does a more thorough job at no cost to us.  Where we're re-iterating information contained there, I'd prefer these explanations be grouped in a sections entitled something like "ExprTK Core Quickstart" or similar, so there is a clear distinction between this and ...
2) ... A dedicated section for _just_ the non-native extensions we've made to ExprTK, namely the extended scalar types (dates/strings/bools/etc), type inference system, column reference syntax (double-quotes), column alias via comments, and single quote string literals.  Not everything has to be covered right away in this PR, the key is to organize what's already written re: these extensions, separately from the core language quickstart.
3) Please move any additions you want to make to the our added core-library functions themselves to to Rust metadata statics (as per the inline comment re: autogeneration+intellisense). 
4) Try to remove the image assets (that are missing from this PR anyway) - you can replace them with CSS highlights of an embedded `<perspective-viewer>` if you're feeling adventurous, or just remove them.  We are trying to put the project on a diet of debt like this.

Side note - the example expressions are good, I'd love to see a `/examples/blocks` demo of a number of these expressions as applied to `superstore.arrow` so we can make it a `README.md` gallery example highlighting different expressions ... maybe a topic for a future PR?

brianblakely:
Thank you for your feedback, @texodus, will be working on this in the coming days.

texodus:
Been a few weeks on this so I went ahead and made the changes in #1903 - closing this PR

brianblakely:
> I went ahead and made the changes in #1903 - closing this PR

This is greatly appreciated, thank you!

Issue #1833, opened 2022-06-03
`boolean` columns are always editable in `perspective-viewer-datagrid`


timkpaine:
### Steps to Reproduce:
- Construct any `perspective-viewer` with a `datagrid` plugin
- Turn to `Read Only`
- Click a boolean column

### Expected Result:
- `check` won't change to `close` and vice versa

### Actual Result:
toggles

### Environment:
All

PR #1850, opened 2022-06-11
Reimplement `editable` for PerspectiveWidget in JupyterLab to take into account the latest changes

timkpaine:
This PR does 2 things: Fixes some small bugs with boolean values in the perspective datagrid extension (smaller part), and reimplements the `editable` attribute to account for the latest changes in `perspective-viewer`.

# Editability of boolean values in the `datagrid` plugin
CC: https://github.com/finos/perspective/issues/1833

Due to a small bug, boolean columns were always shown as editable.

Hover behavior:
hover

Click behavior:
click

This behavior is fixed to not toggle booleans and not trigger `cursor:pointer` unless the data grid is marked as editable

Hover behavior:
hover

Click behavior:
click

# Editable in PerspectiveWidget
`editable` was added to the `PerspectiveWidget` and `PerspectiveTraitlets` for legacy reasons. Even though the client/server interface existed, the `PerspectiveWidget` plays a dual role as a wrapper for both tables and views, but also the frontend `perspective-viewer`. However, the `perspective-viewer` has no formal client/server interface.

Previously, it was written in pure JS, and a 1-1 wrapper was written in Python in the form of the `PerspectiveTraitlets` class. However, no end-to-end tests were ever written to assert this functionality. As changes were made, most notably the shift to current rust-based `perspective-viewer`, various parts regressed, and any problems had to be manually discovered and fixed.

One of the major regressions was the removal of `editable` as an attribute of `perspective-viewer`, and the inclusion of it in the `plugin_config` for the perspective datagrid. This PR brings the python code in line with the latest rust/JS state by removing `editable` completely from the `PerspectiveTraitlets` and `PerspectiveWidget` class, and reimplementing the required bidirectional syncing to allow for the `plugin_config` based editable flag. This is technically a breaking change as the `editable` attribute is now removed from the various python classes, however this behavior was already non-functional.

## Old Behavior

Construction (non-functional):
constructor

Toggling from JS (non-functional):
js

Toggling from Python (non-functional):
python

## New Behavior
Construction:
constructor

Toggling from JS:
js

Toggling from Python:
python

Working end-to-end:
works

# Tests
Almost more important than these bug fixes, the JupyterLab tests are greatly expanded. Previous tests only checked the load-time behavior of the Perspective Widget:

- loading data
- loading data with an update
- loading data from a `Table` instance
- loading with `settings=False`

These existing tests are expanded with additional tests, including tests which dynamically execute additional python cells, or manipulate the frontend components and assert the syncing of state back to python:

- Loading a widget with editable
- Toggling editability from python (syncs to JS)
- Toggling editability from JS (syncs to Python)
- Toggling `columns`, `filter`, `plugin`, `group_by`, `split_by`, `sort`, and `theme` from Python (syncs to JS)
- Toggling `columns`, `filter`, `plugin`, `group_by`, `split_by`, `sort`, and `theme` from JS (syncs to Python)
- Editing in the frontend (syncs back to Python data)

Helper functions are added to run Python code in a cell after initial notebook run, and to run Python assertions in a cell and track if Python exceptions were raised. This should aid in future extension of these tests.


packages/perspective-jupyterlab/test/jupyter/utils.js
```diff
@@ -112,17 +121,49 @@ module.exports = {
...
         await page.waitForSelector(
             '.jp-Notebook-ExecutionIndicator:not([data-status="idle"])',
-            {timeout: 3000}
+            {timeout: 5000}
         );
 
         await page.waitForSelector(
             '.jp-Notebook-ExecutionIndicator[data-status="idle"]',
-            {timeout: 3000}
+            {timeout: 10000}
```
texodus, line 129: These timeouts are really long and there is no explanation as to why they were added (a total of 45s per test added in this section, 30000, 10000, 5000).  Does that mean the new tests were flakey and this additional tolerance fixes it?  If so, this is a strong indication to me that the conditions we are awaiting are flawed.

Otherwise, if these are just debug code that snuck into the PR, please revert

packages/perspective-jupyterlab/test/jupyter/utils.js
```diff
@@ -112,17 +121,49 @@ module.exports = {
...
+        // after clicking new cell, the document will auto
+        // focus the new cell, so lets grab it
+        const el = await page.evaluateHandle(() => document.activeElement);
+        await el.type(cell_content);
+
+        // now while the element is still focused, click the run cell button
+        await page.click('.jp-Button[data-command="runmenu:run"]');
+
+        // wait a small time
+        await new Promise((res) => setTimeout(res, 2000));
```
texodus, line 159: Ditto - anything that involves "wait an arbitrary time" is pretty much guaranteed to fail eventually.  Certainly there is an appropriate condition to await here ...

packages/perspective-jupyterlab/test/jupyter/widget.spec.js
```diff
@@ -9,30 +9,30 @@
...
+    assert_no_error_in_cell,
+} = require("./utils");
 
 jest.setTimeout(30000);
 
+const getEditable = async (viewer) => {
+    return await viewer.evaluate(async (viewer) => {
+        return viewer.children[0]._is_edit_mode;
+    });
+};
```
texodus, line 24: Why are we checking this internal (and arbitrary_ property instead of `(await viewer.restore()).plugin_config.editable`?

packages/perspective-jupyterlab/test/jupyter/widget.spec.js
```diff
@@ -105,37 +119,351 @@ utils.with_jupyterlab(process.env.__JUPYTERLAB_PORT__, () => {
...
+
+                    // do synchronous
+                    for (let bool of bools) {
+                        bool.click();
+                        // wait a short time
+                        await new Promise((res) => setTimeout(res, 250));
+                    }
+
+                    // wait a small time
+                    await new Promise((res) => setTimeout(res, 1000));
```
texodus, line 428: Same - these are silly also since you can listen for updates on the viewer or view

texodus (commented):
Thanks for the PR!  Great tests coverage and thorough explanation of the issue is much appreciated.

I have a minor gripe, which is that tests that rely on arbitrary timeouts are _doomed_ to fail over enough time.  I understand that these are tiny, few and arguably inconsequential here, but #1823 corrects another set of seemingly-innocuous timeouts which caused _hundreds of hours_ of wasted CI time by only failing ~5% of the runs (but after 45m of build).

Basically, if shrinking these timeouts causes _consistent_ failures, we need to remove them and use discrete await conditions instead.  I'm willing to take your word that these have been tested and/or are well understood, but I'm going to hold off on merging completely until I've verified that the tests are stable in CI through several build cycles.

packages/perspective-jupyterlab/test/jupyter/utils.js
timkpaine: There is no guaranteed condition to wait for here, this code is running arbitrary python code that may or may not mutate any frontend state. If you're not mutating frontend state, you won't have anything for puppeteer to wait on and in my experimentation the idle kernel state indicator is also unreliable.

packages/perspective-jupyterlab/test/jupyter/widget.spec.js
timkpaine: How long do you wait for the changes to update from python to JS? Isn't this just another avenue for arbitrarily long wait times? If tell puppeteer to wait for the viewer to mutate, how long do you wait for it to timeout? If you make it too soon you'll introduce extra flakiness.

packages/perspective-jupyterlab/test/jupyter/widget.spec.js
timkpaine: Sounds reasonable

packages/perspective-jupyterlab/test/jupyter/utils.js
timkpaine: Yes these needed to be extended to avoid spurious failures. This code is basically waiting for the python kernel to process the initial perspective object instantiation, the timing of which is going to vary based on the specs and load of the CI machine. I don't know how you would configure any wait to be long enough to account for these variations without introducing flakiness while being short enough to not extend the build too long during failed builds.

Theoretically these waits should be as long as possible, and we should have jest fast-fail since if you're waiting too long it's highly likely your test is failing.

packages/perspective-jupyterlab/test/jupyter/utils.js
texodus: You can await on the presence of the DOM structure output from the cell, the `Out[#]` cell indicator, or a number of other DOM changes.   `PerspectiveWidget` tests in this file do so with `.jp-OutputArea-output perspective-viewer`.  You may need to use interactive tests to check what the DOM looks like at resolution - but if there truly is "no guaranteed condition to wait for" ... then there is nothing to test.

packages/perspective-jupyterlab/test/jupyter/widget.spec.js
texodus: Like I suggested, you can listen for updates on the `<perspective-viewer>` or `View` for one option.  A much more explicit option would be to await the presence of the `<table>` -> `<td>`  DOM structure for the editable cell with the appropriat checkbox visible?

packages/perspective-jupyterlab/test/jupyter/utils.js
timkpaine: How long do you wait before timing out?

timkpaine: Also which cell's `Out[#]` do you wait on? How do you determine one cell's output from another preexisting cells output? Can you provide an example of puppeteer waiting for specific text content?

timkpaine:
welp nevermind, this is testing pretty stable for me locally but I thought it would've had issues on CI. we can bump the ci a few times to test if we want

PR #1853, opened 2022-06-13
GHA Refactored (proposed)

texodus:
Refactored version of #1808 with "default" aspects moved.


timkpaine:
Note that due to the cleaning up of the build/test scripts for python (which were a mess), python is no longer being tested or packaged on this PR

timkpaine:
lets do direct on #1808

PR #1859, opened 2022-06-15
Remove unused watch commands from viewer-d3fc and viewer-datagrid, remove dependabot, dont trigger full build on merge to master

timkpaine:
As discussed offline, removing the default dependabot upgrading, we will upgrade manually in response to features or bug fixes we want. Also removing the full build from merges to master, we can easily trigger manually and we will still build full on tags. 

Finally, removing some unused watch commands discovered during other development.


texodus (approved):

PR #1866, opened 2022-06-27
Update `emsdk` to `3.1.14`

texodus:
A set of internal changes which make Perspective compatible with Emscripten's `STANDALONE_WASM` output mode, and otherwise reduce the complexity/redundancy of the CI process.
* Update emscripten to `3.1.14`
* Update wasm-opt to `109`
* Remove `main()` from emscripten bindings, switch to `--no-entry` compilation mode.
* Removes double-linking (for node and browser) portion of emscripten build.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -19,7 +19,6 @@ on:
       - docs/
       - examples/
       - python/perspective/README.md
-  pull_request:
   workflow_dispatch:
     inputs:
       ci-full:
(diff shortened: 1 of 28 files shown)


.github/workflows/build.yml
```diff
@@ -1247,6 +1246,9 @@ jobs:
             os: macos-10.15
 
           # Exclude Python 3.7 and 3.8 builds
+          - is-full-run: false
```
timkpaine, line 1249: Wrong spot for this

.github/workflows/build.yml
```diff
@@ -19,7 +19,6 @@ on:
       - docs/
       - examples/
       - python/perspective/README.md
-  pull_request:
```
timkpaine, line 22: Why is this removed?

PR #1867, opened 2022-06-27
Remove Python 3.6 support

texodus:
This PR is a fork of [tkp/py36cleanup](https://github.com/finos/perspective/tree/tkp/py36cleanup), which does not have its own PR but I've reviewed.  I've modified the last PR to remove the Docker changes (as per previous comments).


timkpaine:
one windows test failed on a test that can very rarely be flaky so rerunning.

Issue #1869, opened 2022-06-27
Wrong Label Color


100starnight:
### Steps to Reproduce:
 1. Go to the example bar chart "Quantity & Sales Vs Sub-Category" at [https://perspective.finos.org/](https://perspective.finos.org/)
 2. Click one of the y-axis labels to move it to the right axis
 3. Notice that whichever label you didn't click is gray instead of the color of the data

### Expected Result:
The un-clicked label should stay the same color as its data

### Actual Result:
The label that wasn't clicked turns gray

### Additional Context:
![Screen Shot 2022-06-27 at 4 51 48 PM](https://user-images.githubusercontent.com/55207313/176042372-225a7c99-6c53-4747-81b6-ee9690186c6c.png)

PR #1871, opened 2022-06-29
Fix wrong label color bug

100starnight:
Resolves Issue #1869 

Keep labels the same color as their data, but hide the switch-axis arrow if the label is un-moveable

Screen Shot 2022-06-30 at 11 43 34 AM

### DIFF
--- a/packages/perspective-viewer-d3fc/src/js/axis/splitterLabels.js
+++ b/packages/perspective-viewer-d3fc/src/js/axis/splitterLabels.js
@@ -24,16 +24,16 @@ export const splitterLabels = (settings) => {
             .dataJoin("span", "splitter-label")
             .key((d) => d);
 
-        const disabled = !alt && labels.length === 1;
+        const unMoveable = !alt && labels.length === 1;
         const coloured = color && settings.splitValues.length === 0;
         labelDataJoin(selection, labels)
-            .classed("disabled", disabled)
+            .classed("unMoveable", unMoveable)
             .text((d) => d.name)
             .style("color", (d) =>
                 coloured ? withoutOpacity(color(d.name)) : undefined
             )
             .on("click", (event, d) => {
-                if (disabled) return;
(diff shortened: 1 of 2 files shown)


100starnight:
> To make sure your pull request is accepted successfully, we ask all our open source contributors to sign a [Contributor License Agreement](https://finosfoundation.atlassian.net/wiki/spaces/FINOS/pages/75530375/Legal+Requirements#LegalRequirements-ContributorLicenseAgreement); having reviewed our contributor list, we require a CLA for the following people : (@100starnight)

Does anyone know why the finos-cla-bot doesn't recognize me? I signed the CLA [here](https://github.com/finos/open-developer-platform/pull/317); is there something else I have to do?

texodus:
@100starnight Thanks for the PR and the effort is appreciated, but this PR does not actually fix the issue reported.  It looks like you've done 2 things:
1. Renamed the `disabled` variable and classname to `unMoveable`.  This would be inert normally, but you missed several references to `disabled` class in the `.less` assets, mainly [this one](https://github.com/finos/perspective/blob/master/packages/perspective-viewer-d3fc/src/less/chart.less#L297), which is the actual cause of the style issue (disabled controls default to `greyscale()` filter).  So while this change technically makes the label the correct color, it does so by just breaking the code that explicitly made it grey, leaving the vestigial broken code behind and introducing a few other CSS quirks (e.g. the hover color indicator now triggers on disabled labels).  The correct solution is to just remove the `greyscale()` rule.
2. You've changed the default opacity of the arrow indicator under one of the cases of the renamed "disabled"/"unMoveable".  I don't know what the intention behind this change was but it does not impact the label color.

I've made the real fix in #1883

As a side note, while we don't publish a style guide, it's considered good PR etiquette to try to follow the implicit style when refactoring - `unMoveable` stands alone in this project as a camelCase CSS class identifier.

PR #1893, opened 2022-07-05
Remove custom docker images

texodus:
Removes the `/docker` directory and all usage of it.

* `perspective-viewer` tests do not need this, they have not in a while - I think this is a regression from Azure
* `_wheel_python` now uses images [from PYPA](https://github.com/pypa/manylinux)


timkpaine:
This PR breaks the `publish_python` script due to wheel differences:
https://github.com/finos/perspective/blob/master/scripts/publish_python.js#L69

texodus:
No, it doesn't - please test before you comment.

PR #1907, opened 2022-07-11
Fix Python sdist CI build 

texodus:
[fail](https://github.com/finos/perspective/runs/7283232392?check_suite_focus=true)

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1466,7 +1466,7 @@ jobs:
     ################
     # Python
     - name: Install python dependencies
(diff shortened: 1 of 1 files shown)


texodus:
[still fail](https://github.com/finos/perspective/runs/7286540688?check_suite_focus=true)

timkpaine:
https://github.com/finos/perspective/blob/0e3e3d691c10888e0eb95313ed84a292e30efa3b/python/perspective/setup.cfg#L20 Try adding the pytest command line arg for this

timkpaine (approved):
Obviously we want to get away from enumerating deps in multiple places (which I think we can do by installing <wheel>[develop]) but good for now

PR #1915, opened 2022-07-25
Fix publish issues

texodus:
* Pin `JupyterLab` in CI to 3.4.3, as a recent release has broken the perspective master build.
* Fixes the [example link to binder](http://beta.mybinder.org/v2/gh/finos/perspective/master?urlpath=lab/tree/examples/jupyter-notebooks), which currently installs the latest `perspective-python` and `@finos/perspective-jupyterlab` packages at the time of docker container construction.  These may not always be in sync, so I've added a check which makes sure the JS matches the Python library's version.
* Fixes various broken cases in the `scripts/publish_*` scripts encountered during publishing `1.6.0`
* Fix `@finos/perspective-jupyterlab` to use `@finos/perspective-viewer-openlayers` UMD build, as jupyterlab's build seems to have trouble resolving `ol` for some reason.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -304,7 +304,7 @@ jobs:
     ################
     # Python
     - name: Install python dependencies
-      run: python -m pip install --upgrade "jupyterlab>=3.2" "tornado==6.1"
+      run: python -m pip install --upgrade "jupyterlab==3.4.3" "tornado==6.1"
 
     ################
     # Linux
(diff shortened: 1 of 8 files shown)


timkpaine:
fyi related pins 
https://github.com/finos/perspective/blob/ecd36ac64118250123a560ac0b69e506c3e80fb6/python/perspective/setup.py#L65
https://github.com/finos/perspective/blob/ecd36ac64118250123a560ac0b69e506c3e80fb6/python/perspective/setup.py#L69

texodus:
@timkpaine This PR was _not_ made under the assumption that either the random version bounds in `build.yml`, or the bounds specified in `setup.py`, are correct;  they clearly aren't - a release of a dependency within these bounds broken our `master` branch.  Resolving this break is the only thing this PR attempts to address.

For future work in this area, I'll add that it is also IMO a major _code smell_ that I needed to change ***six*** instances of the `jupyterlab` version bound.

PR #1917, opened 2022-07-25
Cleanup up python requirements installation in CI, remove some dangling azure references, refresh python README

timkpaine:
- removing bespoke package installs/runs in the github actions script (python dependencies and pytest configuration)
   -  fixes the very correct complaints in https://github.com/finos/perspective/pull/1915#issuecomment-1194667739 namely, dependencies are derived from a single source of truth that is the `setup.py`
   - dependencies can still break our CI process, but these should be legitimate breakages of compatibility where a dependency has triggered an incompatibility at **run time**, rather than some stale hardcoding of a version of dependency in a CI script at **ci time**
- Remove python readme as it is trivial for it to become stale, instead pull main README during python build/packaging
- Remove azure references from package.json, remove azure build script
- add python 3.10 pypi classifier
- remove `.artifactignore` azure config file
- add `tornado` pin < 6.2
- add `jupyterlab` pin < 4
- add conda installation instructions to python docs
- cleanup some typos and unclear stuff in python docs
- redo of https://github.com/finos/perspective/pull/1827 now that the docker images are gone (cherry pick from #1925)
- [MacOS 10.15 has been long deprecated by GItHub](https://github.com/actions/virtual-environments/issues/5583) and our CI is failing currently due to brownout.


timkpaine:
@texodus should trigger a full build so we can verify that `grep` works on windows 😅

python/perspective/setup.py
```diff
@@ -62,9 +62,11 @@ def get_version(file, name="__version__"):
 
 requires_aiohttp = ["aiohttp"]
 
+requires_jupyter = ["jupyterlab>=3.2,<4"]
```
texodus, line 65: As I mentioned in #1915 - [JupyterLab `3.4.4`](https://github.com/jupyterlab/jupyterlab/releases/tag/v3.4.4) (just released) [doesn't work](https://github.com/finos/perspective/runs/7491578729?check_suite_focus=true) and I haven't fixed it, so these bounds will fail on a fresh install right now.

README.md
```diff
@@ -3,6 +3,7 @@
 
 [![npm](https://img.shields.io/npm/v/@finos/perspective.svg?style=flat)](https://www.npmjs.com/package/@finos/perspective)
 [![PyPI](https://img.shields.io/pypi/v/perspective-python.svg?style=flat)](https://pypi.python.org/pypi/perspective-python)
+![Conda](https://img.shields.io/conda/v/conda-forge/perspective)
```
texodus, line 6: 👎

No more badges please.

python/perspective/README.md
```diff
@@ -2,42 +2,50 @@
 
 Python APIs for [perspective](https://github.com/finos/perspective) front end
 
-[![npm](https://img.shields.io/npm/v/@finos/perspective.svg?style=flat-square)](https://www.npmjs.com/package/@finos/perspective)
```
texodus, line 5: I realize that this is out-of-date, but it is worth asking _why_ it is out-of-date.  Implied rules for project maintenance are not scalable to a project of this size.

If this is to be a duplicate of the project readme, can we add this copy to `.gitignore` and duplicate it as part of the python build please - or replace it altogether with a trivial readme that just links to the project `README.md` on GitHub?

python/perspective/setup.py
timkpaine: You fixed the incompatibility here https://github.com/finos/perspective/pull/1915/files and this PR is currently passing tests with jupyterlab 3.4.4

README.md
timkpaine: Booo! More badges! I'll update the docs instead to mention conda, probably worth it as e.g. Mac arm and python 3.10 builds are already available there while not yet available as wheels on pypi

python/perspective/README.md
timkpaine: I'm good with that, it has been a slightly different variant of the main README for a while now, not sure why it passed review but deleting and copying on build is better

README.md
timkpaine: fixed

python/perspective/README.md
timkpaine: fixed

texodus (approved):

texodus:
Adding #1924 also

Issue #1039, opened 2020-05-08
Overlapping of chart labels and tick values 


aniruddhamurali:
### Description of Problem:
Sometimes, when displaying the candlestick and OHLC charts, the y-axis tick values and label overlap, making the charts a little hard to read. I found this example in the Conceptual Overview tab on the Perspective docs page:

<img width="787" alt="Screen Shot 2020-05-07 at 9 26 28 PM" src="https://user-images.githubusercontent.com/35414964/81363084-76ccd280-90b0-11ea-9c69-de1cc08fbd40.png">

### Potential Solutions:
Y-axis label should be positioned such that there is consistent space between the y-axis values and the label. Chart may need to be scaled accordingly to make the chart display properly.

texodus:
Thanks for the PR!

Yeah this is annoying.  This also occurs on the X-axis, especially with `date` or `datetime` axes when the major ticks are spaced too closely.  Should be a pretty easy fix though!

arteevraina:
@texodus Can I work on this issue?

rgrupesh:
@texodus 

This is CSS issue, right? Can I work on this?

texodus:
Not a CSS issue - labels are calculated by D3FC in JavaScript.

You don't need to ask for permission to work on an issue - just go for it.

rgrupesh:
@texodus 

I trying to setup development environment on my local machine(Ubuntu 20.04). I did everything as per [developer guideline](https://perspective.finos.org/docs/md/development.html#building).

The thing is I have only `boost` dependencies(`flatbuffers` is missing) in `/usr/include`.

As per:

![Screenshot from 2021-09-10 18-02-00](https://user-images.githubusercontent.com/54068868/132852666-23d824b3-1e73-4fe8-a849-8e55521e2558.png)

texodus:
@rgrupesh Please, this is not Facebook or Blackboard;  GitHub Issues are for tracking and note aggregation for specific bugs and features, I've had to delete a few of your comments that are off-topic and obfuscate this important function.

Please use [Discussions](https://github.com/finos/perspective/discussions) for questions, not Issues.

Ranganathan-J:
use label rotation and dyamic tick spacing for this issue. let me explain dynamic tick spacing means it automatically adjusts the y-axis based on zoom requirements. label rotation means the numbers on the y-axis should be in horizontal.

texodus:
[...]

PR #1931, opened 2022-08-03
Format numbers to avoid common label overlaps

100starnight:
## Description
Long tick labels tend to overlap the axis labels (issue #1039); this PR updates the value formatter to avoid the most common instances of collision.

## Photos
axisTickLabelOverlapBug
axisTickLabelOverlapFix


raymondfx (approved):

jkusa (approved):

100starnight:
Hey @texodus I see the tests were completed successfully; just wanted to check in and see if this pr is just waiting for reviews or if there's another approach I should consider...?

texodus:
@100starnight I appreciate the attention to this issue but this is not the correct fix.  The axis label container width (in this case for `linearAxis` types in the Y position) are not being calculated correctly (or at all - I don't know offhand).  Your fix just shortens the text generated within this container in.a specific case.  While this removes overlap in this specific case, the container is still the wrong width, so different data-extents/fonts/etc still will overlap.  Ergo, the PR "fixes" some cases and "breaks" others;  this is common symptom of fixes that wallpaper-over rendering issues with CSS.

A correct fix will make sure the axis container is the minimum bounding rectangle for the rendered labels.   I suspect the we'll need to implement something like the [bounds calculation for `ordinalAxis`](https://github.com/finos/perspective/blob/master/packages/perspective-viewer-d3fc/src/js/axis/ordinalAxis.js#L249).

100starnight:
Hey @texodus! While I agree that this update doesn't address all of the bug cases, I do think it's a good quick win for a very common bad UI experience (as reported by our users & others who've tried to fix the bug before). Would you consider merging this change while I keep looking into it?

Honestly, even independent of the overlap issue, I think that re-evaluating the ticks' current level of precision is worthwhile - setting SI_MIN to 100,000 seems fairly reasonable and it's more readable than our current state

100starnight:
@texodus thoughts on above?

PR #1943, opened 2022-08-22
Per-component CSS model

texodus:


python/perspective/setup.py
```diff
@@ -52,7 +52,7 @@ def get_version(file, name="__version__"):
 version = get_version(os.path.join(here, "perspective", "core", "_version.py"))
 
 requires = [
-    "ipywidgets>=7.5.1",
+    "ipywidgets>=7.5.1,<8",
```
timkpaine, line 55: This is good, it's going to take some digging to get 8 fully in order. Might want to pin on the JS side as well

PR #1958, opened 2022-08-29
build!: assume `emcc` already in `PATH` shouldn't source `emsdk_env.sh`

ErichDonGubler:
This allows us to avoid making assumptions on platforms where `emscripen` is more-or-less installed
correctly for our use, like with Chocolatey on Windows.

Signed off by @texodus while pairing.

### DIFF
--- a/package.json
+++ b/package.json
@@ -86,6 +86,7 @@
         "webpack": "^5.60.0",
         "webpack-cli": "^4.3.1",
         "webpack-dev-server": "^3.11.2",
+        "which": "^2.0.2",
         "worker-loader": "^3.0.7"
     },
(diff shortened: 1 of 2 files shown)


ErichDonGubler:
Not sure how to get the CLA check to go green; I've [followed the instructions](https://github.com/finos/open-developer-platform/pull/329), AFAICT.

texodus:
Thanks @ErichDonGubler!

The test failure is due to our misunderstanding of the `which` module - it throws when the symbol is not found, rather than returning false-y.  I would fix this, but I also want to re-evaluate adding this dependency, so closing for now until the build issues comes up again.

PR #1964, opened 2022-09-06
Minor Build fixes

sinistersnare:
Hi,

I have been working on a Containerfile (Dockerfile) for this project, and to build it, I had to make these changes.

* The addition of the `package.json` is to appease Yarn 3
* The addition of the `typescript` dev dependency in the rust subproject was needed during the `yarn build` phase.
* And the `let_chains` feature was needed to build the Rust code.

I'm not sure how the project builds now without these changes, so if they are unneeded, let me know!

Thanks!
Davis


timkpaine:
Typescript is already a top level dev dependency:
https://github.com/finos/perspective/blob/master/package.json#L84

We use yarn 1.* for everything.

The rust change I don't know about, @texodus can comment. 

There is a developer guide on the website: [here](https://perspective.finos.org/docs/development).

sinistersnare:
> Typescript is already a top level dev dependency:
https://github.com/finos/perspective/blob/master/package.json#L84

That is what surprised me. For some reason, when i run without this dev-dependency, `yarn build` fails, stating that `yarn tsc` can't be found (for when it runs the `build.js` in `rust/perspective-viewer`).

> We use yarn 1.* for everything.

Sure, but supporting Yarn 3 is as easy as adding this file. I think its worth it to keep it. However, if you still want, I can definitely remove it in favor of staying on Yarn 1.

> There is a developer guide on the website: [here](https://perspective.finos.org/docs/development).

The developer guide is a bit out of date. For example, it says flatbuffers needs to be installed prior, but that actually caused a pretty strange error. Instead, the `yarn` commands auto-install flatbuffers nowadays.

timkpaine:
> That is what surprised me. For some reason, when i run without this dev-dependency, `yarn build` fails, stating that `yarn tsc` can't be found (for when it runs the `build.js` in `rust/perspective-viewer`).

you ran `yarn` then `yarn build` at the top level of the project and it failed?

> Sure, but supporting Yarn 3 is as easy as adding this file. I think its worth it to keep it. However, if you still want, I can definitely remove it in favor of staying on Yarn 1.

I don't have an opinion, though that folder is internal-only so would likely be better replaced by just moving it inside instead of a [separate file](https://github.com/finos/perspective/blob/master/docs/package.json#L27).
 
> The developer guide is a bit out of date. For example, it says flatbuffers needs to be installed prior, but that actually caused a pretty strange error. Instead, the `yarn` commands auto-install flatbuffers nowadays.

Docs were updated ~2 months ago, we don't require flatbuffers as we don't need need the `flatc` executable anymore. But otherwise not much has changed installation-wise, far better to reference the developer docs than to try to reverse-engineer something.

sinistersnare:
> you ran `yarn` then `yarn build` at the top level of the project and it failed?

Ah, I did not run `yarn` only. That's not mentioned in the developer guide, so I forgot about that.

Just tested it. Running `yarn`, `yarn install`, `yarn build` still fails without this line added to the dev-dependencies. I do not know much about JS-ecosystem subproject isolation, but in my docker container, when I move up to `rust/perspective-viewer` and try `yarn tsc --version` it fails, unless this dev dependency is listed.

> ... so would likely be better replaced by just moving it inside instead of a [separate file](https://github.com/finos/perspective/blob/master/docs/package.json#L27).

Interesting idea. Would you like me to make that change, or should we save that for another discussion on specifics?

> we don't require flatbuffers

I think the problem is that `arrow` has a dependency on flatbuffers, and my installed flatbuffers was a different version than it should have been. This caused the error. So because I installed them it caused a compilation failure in the C++ build.

> far better to reference the developer docs than to try to reverse-engineer something.

Definitely true! I did start with the commands from the developer-guide, but I was still getting some errors, so eventually I got to here. Im hoping that this work (that is, working on a Dockerfile), can strengthen the developer-guide even more, giving people an idea of what exactly is required to build the project.

timkpaine:
> Just tested it. Running `yarn`, `yarn install`, `yarn build` still fails without this line added to the dev-dependencies. I do not know much about JS-ecosystem subproject isolation, but in my docker container, when I move up to `rust/perspective-viewer` and try `yarn tsc --version` it fails, unless this dev dependency is listed.

This is not how the project is built, you will be running `yarn build` from the top level only.

sinistersnare:
Yeah, `yarn build` at the top level itself goes to that directory and runs `yarn tsc --project tsconfig.json`. And without this dev dependency, it fails. Sorry for not being specific, I was commenting on what the build process is doing (from my perspective, again I don't know it so well)

texodus:
Thanks for the PR @sinistersnare!  Unfortunately I don't think any of these fixes apply.

From a read of the thread here, the issue seems to be -
* Unclear developer documentation regarding `workspace` structure, dev dependencies and the intended usage.
* Wrong dependencies -
    *  Rust  you need 1.64 (perhaps not documented - I didn't check).
    *  Yarn - we are not planning to support yarn 3.  This is not as easy as adding a file (and this is the source of your typescript dev-dependency error also).

There's an unresolved mystery around your `flatbuffers` experience, but this is clearly already far down a rabbithole of unrelated knock-on effects as above.  As @timkpaine mentioned, we don't require `flatbuffers` anymore (docs are out-of-date).  However, we also do not require `arrow`, so not clear how a local copy of _either_ could cause the error you describe.  Please open an issue around this specifically if you reproduce it with a compatible Yarn and otherwise mystery-free build?

PR #1984, opened 2022-10-03
Bump dependencies when installing and building python, drop support for `manylinux2010` 

timkpaine:
- adds the `-U` flag to the python requirements installation, and adds `pip` as a dev dependency. 
- drops the build of `manylinux2010` wheels, deprecated August 2022


texodus:
[full build](https://github.com/finos/perspective/actions/runs/3177600183)

texodus (approved):

PR #1991, opened 2022-10-08
accidentally copying license as readme when disting python

timkpaine:
I made a copy-pasta typo when writing the script, so the python README is actually the LICENSE. this fixes that.

Screen Shot 2022-10-08 at 16 05 46


100starnight (approved):

texodus (approved):

PR #2011, opened 2022-11-16
Add python 3.10 to build matrix

timkpaine:
Fixes https://github.com/finos/perspective/issues/1682
Picks up a fix for a change to a dependency used in tests
Big build: https://github.com/finos/perspective/actions/runs/3482649366
`gh workflow run "Build Status" --ref tkp/3.10  -f ci-full=true -f ci-skip-cache=false  -f ci-skip-python=false -f ci-include-windows=true`

Note: perspective is already built and deployed across linux/Mac/Mac arm/windows for python 3.7-3.10 via conda-forge https://github.com/conda-forge/perspective-feedstock

PR #2013, opened 2022-11-20
Re-add secondary dependency as primary dependency to fix python tests

timkpaine:
extracted from https://github.com/finos/perspective/pull/2011 which is segfaulting in 3.10

xref: https://github.com/encode/starlette/issues/1943, https://github.com/encode/starlette/issues/1945

https://www.starlette.io/#dependencies

`[httpx](https://www.python-httpx.org/) - Required if you want to use the TestClient.`

### DIFF
--- a/python/perspective/pyproject.toml
+++ b/python/perspective/pyproject.toml
@@ -1,2 +1,2 @@
 [build-system]
-requires = ["setuptools", "wheel", "numpy>=1.13.1"]
+requires = ["setuptools", "wheel", "numpy>=1.13.1,<2"]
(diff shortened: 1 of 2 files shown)


texodus:
Does this go in `require_starlette` optional dependency?

Can we please also add version bounds to `starlette` itself?

timkpaine:
> Does this go in `require_starlette` optional dependency?
> 
> Can we please also add version bounds to `starlette` itself?

from the docs:

`Required if you want to use the TestClient.`

So no, really should only go in the develop dependencies as its not a runtime dep.

texodus (approved):

PR #2045, opened 2022-12-12
Provide better tooltip offset in relation to points.

brochington:
Currently tooltips (labels) of point data have the possibility of overlapping the actual point area. This PR is an attempt to provide a more calculated offset based on the area of the point.


brochington:
My math is off on this. Will resubmit when I have a better solution.

texodus:
@brochington the Label fix #2035 has similar positioning logic we could steal

PR #2047, opened 2022-12-13
Feature/improved point label position (redux of #2045)

brochington:
Currently tooltips (labels) of point data have the possibility of overlapping the actual point area. This PR is an attempt to provide a more calculated offset based on the area of the point.

redux of [this PR](https://github.com/finos/perspective/pull/2045)

### DIFF
--- a/packages/perspective-viewer-d3fc/src/js/series/pointSeriesCanvas.js
+++ b/packages/perspective-viewer-d3fc/src/js/series/pointSeriesCanvas.js
@@ -12,6 +12,9 @@ import { groupFromKey } from "./seriesKey";
 import { fromDomain } from "./seriesSymbols";
 import { toValue } from "../tooltip/selectionData";
 
+const LABEL_PADDING = 5;
+const LABEL_COSINE = 0.7071067811865476; // cos(45 * Math.PI / 180)
+
 export function pointSeriesCanvas(
     settings,
     seriesKey,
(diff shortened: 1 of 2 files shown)


packages/perspective-viewer-d3fc/src/js/series/pointSeriesCanvas.js
```diff
@@ -43,11 +46,25 @@ export function pointSeriesCanvas(
             context.font = settings.textStyles.font;
             const { type } = settings.mainValues.find((x) => x.name === label);
             const value = toValue(type, d.row[label]);
-            const offset = size
-                ? Math.round(scale_factor * size(d.size)) * (15 / 1000) + 10
-                : 10;
 
-            context.fillText(value, offset, 4);
+            let magnitude = 10;
```
texodus, line 50: ... and we should do them all :)

packages/perspective-viewer-d3fc/src/js/series/pointSeriesCanvas.js
```diff
@@ -43,11 +46,25 @@ export function pointSeriesCanvas(
...
+
+            if (size) {
+                // `size(d.size)` is area (A) of circle.
+                // A = pi * r^2
+                // r = sqrt(A / pi)
+                const radius = Math.sqrt(
+                    (scale_factor * size(d.size)) / Math.PI
+                );
```
texodus, line 58: This was news to me - I had assumed that the `.size()` method meant radius, but according to the docs you are indeed correct, [this is actually _area_](https://github.com/d3fc/d3fc/blob/master/packages/d3fc-series/README.md#seriesPoint_size)!

texodus (commented):
Thanks for the PR!

I really like the approach taken here and I think it looks good with bubble charts.  The spacing between the text and bubble edge in this mode is made very consistent and looks great!

Old:
Screenshot 2022-12-28 at 10 50 14 AM
New:
Screenshot 2022-12-28 at 10 50 25 AM

However, subjectively I find this 45deg angle offset less readable generally than the x-offset version in the old charts, especially without bubbling the next can get lost when the scatter area is crowded.

Old:
Screenshot 2022-12-28 at 10 54 42 AM
New:
Screenshot 2022-12-28 at 10 54 56 AM

I propose we keep the radius calculation for the x-offset, but remove the 45deg angle (e.g. `LABEL_COSINE` is 1 and no y-offset is applied).  This will get us consistent spacing on the x-axis without the (IMO) confusing label placement when data is clustered.

I've done this change in #2067, which I've also rebased onto master and removed the "cargo.lock" commits.  Please [prefer rebase to git merge master and be sure to squash commits](https://github.com/finos/perspective/blob/master/CONTRIBUTING.md) for open Perspective PRs - for a project of this size and regression coverage, it is of great help to have as clean a change history as possible.

PR #2048, opened 2022-12-14
Draft: Additional View methods for faster accessing of row and column data.

brochington:
Currently much of the data exchange between  the perspective data store and charting code happens via calls to `to_json`. This blanket serialization approach might be better served with more precise data accessing. This is especially true when fetching columnar data, where passing a dict and index array can be much faster.


timkpaine:
Also note that any changes like this should also have symmetric changes in python

PR #2054, opened 2022-12-16
Move examples to `gh-pages` site

### LINKED ISSUE #2049: bl.ocks.org down, examples not loading
### Steps to Reproduce:
 1. Click on any example link in the Github readme that points to bl.ocks.org (Heroku).

### Expected Result:
Example app is running on Heroku

### Actual Result:
We get the following error message:

> Application error
> An error occurred in the application and your page could not be served. If you are the application owner, [check your logs for details](https://devcenter.heroku.com/articles/logging#view-logs). You can do this from the Heroku CLI with the command
> `heroku logs --tail`

image

texodus:
Fixes #2049


docs/src/pages/block.js
```diff
@@ -0,0 +1,110 @@
...
+function getQueryVariable(variable) {
+    const query = window.location.search.substring(1);
+    const vars = query.split("&");
+    for (let i = 0; i < vars.length; i++) {
+        const pair = vars[i].split("=");
+        if (decodeURIComponent(pair[0]) == variable) {
+            return decodeURIComponent(pair[1]);
+        }
+    }
+}
```
sinistersnare, line 19: We can use [URL.searchParams](https://developer.mozilla.org/en-US/docs/Web/API/URL/searchParams) instead, it does URI decoding for us too AIUI.

```js
function getQueryVariable(variable) {
  const params = (new URL(document.location)).searchParams;
  return params.get(variable); // returns null if not found.
}
```

texodus: 👍

docs/src/pages/examples.js
```diff
@@ -0,0 +1,112 @@
...
+                        url="https://sc1f.github.io/pudgy-penguin-perspective/"
+                        item="nft"
+                    ></ExampleTD>
+                    <ExampleTD
+                        key="nypd ccrb"
+                        img="https://texodus.github.io/nypd-ccrb/preview.png"
+                        url="https://texodus.github.io/nypd-ccrb/"
+                        item="nypd ccrb"
+                    ></ExampleTD>
+                </tr> */}
```
sinistersnare, line 84: Intentional comment-out?

PR #2062, opened 2022-12-21
Fix background color of plugin select dropdown in Material Dark theme.

brochington:
Current there is a missing custom property (--select--background) in the Material Dark theme. This PR adds a default.


texodus:
Thanks for the PR!

Closing as this has issue bit-rotted as of #2110.

PR #2072, opened 2022-12-29
Add macOS arm wheels to build matrix

timkpaine:
Testing here:
https://github.com/finos/perspective/actions/runs/3804830237

`gh workflow run "Build Status" --ref tkp/armmac -f ci-include-windows=false -f ci-skip-python=false -f ci-full=true`


texodus (approved):
Thanks for the PR! It looks like the full run worked and I can install it on an M1, so I presume this is good to go!

PR #2092, opened 2023-01-25
Update charting tests to cover more HTML

sinistersnare:
Before, D3FC tests only covered a single SVG, now they cover all SVGs in the chart and also Canvases!

(shoutout to @brochington who helped)


texodus (approved):

texodus (approved):
Thanks for the PR!  Looks good (reviewed offline)

Issue #2091, opened 2023-01-25
DOMException CustomElementRegistry dupe on esbuild example


davisford:
### Steps to Reproduce:
 1.  Check out finos/perspective repo 
 2. Copy examples/esbuild-example to tmp/ (Doing this b/c whole repo has higher-level package-lock.json which screws up yarn install paths
 4. cd /tmp/esbuild-example
 5. yarn install
 6. yarn start
 7.  open http://localhost:8080

Console Output:

```sh
➜ yarn start
yarn run v1.22.19
$ yarn build && http-server dist
$ node build.js
Starting up http-server, serving dist
Available on:
  http://127.0.0.1:8080
  http://192.168.25.68:8080
Hit CTRL-C to stop the server
[2023-01-25T16:53:25.933Z] "GET /" "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/109.0.0.0 Safari/537.36"
(node:12630) [DEP0066] DeprecationWarning: OutgoingMessage.prototype._headers is deprecated
(Use `node --trace-deprecation ...` to show where the warning was created)
[2023-01-25T16:53:25.971Z] "GET /perspective_viewer_bg.wasm" "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/109.0.0.0 Safari/537.36"
[2023-01-25T16:53:25.972Z] "GET /editor.worker.js" "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/109.0.0.0 Safari/537.36"
[2023-01-25T16:53:25.973Z] "GET /perspective_viewer_bg.wasm" Error (404): "Not found"
[2023-01-25T16:53:25.973Z] "GET /editor.worker.js" Error (404): "Not found"
```

### Expected Result:
Loads a datagrid at http://localhost:8080

### Actual Result:
Fails to load several resources it wants, and has duplicate registrations in browser's CustomElementRegistry which cause the whole thing to fail.

![Screen Shot 2023-01-25 at 11 59 11 AM](https://user-images.githubusercontent.com/186677/214629388-18a3a42e-0b2f-4160-8c05-735014e6bb73.png)

### Environment:

Mac OS Monterey, M1 chip, no virtual environment.

```
[...]

PR #2095, opened 2023-01-30
Fixing broken example for esbuild 

davisford:
Please refer to this issue for detail https://github.com/finos/perspective/issues/2091

It appears the name of the .wasm file changed and the example never got updated.

### DIFF
--- a/examples/esbuild-example/src/index.html
+++ b/examples/esbuild-example/src/index.html
@@ -7,7 +7,7 @@
         <title>Perspective `esbuild` Example</title>
 
         <link rel="preload" href="perspective.cpp.wasm" as="fetch" type="application/wasm" crossorigin="anonymous">
-        <link rel="preload" href="perspective_viewer_bg.wasm" as="fetch" type="application/wasm" crossorigin="anonymous">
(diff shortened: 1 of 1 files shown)


davisford:
Thank you `finos-cla-bot` -- I'm not keen on signing off on a bunch of captchas and legal, etc. things.  Take or leave the PR @finos team.  You have a great product here.  Just trying to give back in small gestures.

texodus:
Thanks @davisford - apologies about the CLA, I concur it is obnoxious and unnecessary.

You are correct that this file was renamed (and `editor.worker.js` no longer exists either) - however this preload link should not cause errors (and does not at least on my local Chrome).  I'm honestly not sure what's going on here just yet, let me investigate a bit before I merge this to make sure I understand how this can fail, we may prefer to remove the preload hints entirely.

texodus:
@davisford I took a look at this and I am certain (1) this was not the cause of your issue, its just a warning, and (2) this only fixes 1/2 warnings.  As it is not a complete fix (nor an actual issue), I'm closing in favor of #2112

PR #2117, opened 2023-02-23
Remove references to 3.7, default to python 3.10 if not specified, migrate from flake8 to ruff

timkpaine:
- remove a few remaining references to python `3.7`
- default python version to 3.10 if not otherwise specified (should not happen), was `3.7`
- clean up `.gitignore`
- migrate more things from `setup.cfg` to `pyproject.toml` in anticipation of future changes
- remove usage of `flake8`, switch to the much faster [`ruff`](https://github.com/charliermarsh/ruff)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -90,6 +90,7 @@ flycheck_*.el
 *.out
 *.app
 
+# Build artifacts
 build
 CMakeCache.txt
 CMakeFiles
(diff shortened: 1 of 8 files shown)


texodus:
@timkpaine `ruff` is fine but this PR does not seem to work as I expect.  The `lint_python` command no longer seems to flag obvious errors, e.g. this snippet from a simple extra-whitespace issue I injected:

On this branch:

```bash
> yarn lint_python
yarn run v1.22.19
$ node scripts/lint_python.js
Python 3.9.6
✨  Done in 0.87s.
```

On master:

```bash
> yarn lint_python
yarn run v1.22.19
$ node scripts/lint_python.js
Python 3.9.6
perspective/handlers/common.py:31:24: E222 multiple spaces after operator
perspective/handlers/common.py:31:25: BLK100 Black would make changes.

Command failed: cd perspective/python/perspective && python3 -m flake8 perspective bench setup.py
error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```

I think we need `ruff` to provide a similar level of validation _and_ formatting that `black`+`flake8` currently provide.

timkpaine:
we were using `flake8-black` extension for this, this is what make `flake8` detect things that `black` would change and raise them as lint errors (even though they're not technically lint errors). We can restore this behavior if it is desired, this is something akin to `prettier` vs `eslint` and flagging formatting "errors" vs actual static analysis problems.

texodus:
I disagree, please look at my example above again.  The error missed is [E2222 multiple spaces after operator](https://www.flake8rules.com/rules/E222.html), which _is_ a lint error from `flake8` (and a `black` error).  It doesn't seem like _any_ lint errors are flagged in this PR.

texodus:
I just tried `F401` unused-import and `F402` import-shadowed-by-loop-var, the first two supported lints on the ruff docs, and neither are triggered (`flake8` reports both).  Also this is valid `ruff` formatting apparently (try it):

```python
def         
    gen_name(    size      =10
            , chars=       string.ascii_uppercase
        +     string.    digits
    ):
    return ""     .join          (
            random            .choice(
            chars)     for     x in       range(
        size)
    )
```

I think "too new for us" is an understatement.  One sure-fire way to be blazingly-fast is to no-op :)

timkpaine:
> Also this is valid `ruff` formatting apparently (try it):

I get:
```
error: Failed to parse perspective/tmp.py: invalid syntax. Got unexpected token Newline at line 1 column 13
perspective/tmp.py:1:14: E999 SyntaxError: invalid syntax. Got unexpected token Newline
Found 1 error.
```

PR #2118, opened 2023-02-23
fix icon filenames for windows

timkpaine:
https://github.com/finos/perspective/pull/2111 introduces great new icons, but `:` can't be in a filename on windows


texodus (approved):

PR #2123, opened 2023-02-28
Implement `title` field in `PerspectiveWidget`

tomjakubowski:
### Works:
* can set title from a notebook `widget.title = "Hello world"`
### Doesn't work
* setting the view's title from the widget doesn't update python state, so doesn't persist across refreshes


texodus:
Looks like synch doesn't work (as per @tomjakubowski), deferring this to #2131

texodus (approved):

PR #2127, opened 2023-03-03
Distribute prebuilt JupyterLab extension / Notebook Classic extension (2023 edition)

tomjakubowski:


tomjakubowski:
I think I signed the CLA over here, but maybe not https://github.com/finos/open-developer-platform/pull/365

.github/workflows/build.yml
```diff
@@ -208,6 +208,11 @@ jobs:
             ##########
             # Caches #
             ##########
+
+            # needed for act
```
timkpaine, line 212: I would avoid these

scripts/nb_link.js
```diff
@@ -0,0 +1,46 @@
...
+        "extension",
+        "finos-perspective-nbextension.json"
+    )
+);
+
+(async function () {
+    try {
+        execute`mkdir -p ~/.jupyter/nbextensions/@finos/`;
+        execute`mkdir -p ~/.jupyter/nbconfig/notebook.d/`;
+        execute`ln -s ${nb_path} ~/.jupyter/nbextensions/@finos/perspective-jupyterlab`;
```
timkpaine, line 39: this is a Tim special, I symlink everything in `jupyter --paths` to `~/.jupyter`, but this is not the default installation behavior. We should note this somewhere in the contributor docs

.github/workflows/build.yml
```diff
@@ -307,7 +312,7 @@ jobs:
             ################
             # Python
             - name: Install python dependencies
-              run: yarn _requires_python
+              run: pip install jupyter_packaging && yarn _requires_python
```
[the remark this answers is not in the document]
timkpaine: make sure to pin, most of the stuff we use is deprecated and will be removed after `1.0`

.github/workflows/build.yml
tomjakubowski: `act` compatibility would be really nice to have for sanity-checking changes to CI - it's hard maintaining a state clean enough locally to replicate CI issues, and pushing to github actions is sloooow

that said it's likely that just giving in and using one of the bigger images would work here

.github/workflows/build.yml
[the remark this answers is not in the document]
tomjakubowski: there's a chicken-.egg problem here: setup.py has to be able to run before dependencies are installed, but setup.py imports items from jupyter_packaging, which is one of its dependencies

the common fix for that seems to be to try/except the jupyter_packaging import and provide dummy values for eg cmdclasses -- however, this change also depends on jupyter_packaging to parse perspective-jupyterlab's version number.   I could just revert that part + so that setup.py continues to own version parsing.

.github/workflows/build.yml
tomjakubowski: using one of the fat images seems to work. my ~/.actrc

```
--container-architecture linux/amd64
-P ubuntu-20.04=catthehacker/ubuntu:full-20.04
```

.github/workflows/build.yml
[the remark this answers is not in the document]
tomjakubowski: note to self: pin the version installed in _wheel_python.js also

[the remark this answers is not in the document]
tomjakubowski: got this working well enough in setup.py

scripts/nb_link.js
tomjakubowski: I have a pretty plain setup and `~/.jupyter` is listed in my `jupyter --paths` (after `share/jupyter` in my venv root)… so linking this into `~/.jupyter` should make it available to jupyter, I think? I haven't actually tested the nbextension yet

python/perspective/pyproject.toml
```diff
@@ -1,2 +1,10 @@
 [build-system]
-requires = ["setuptools", "wheel", "numpy>=1.13.1,<2"]
+requires = [
+    "setuptools",
+    "wheel",
+    "numpy>=1.13.1,<2",
+    "jupyter-packaging==0.12.0",
```
timkpaine, line 6: should match `setup.py`

python/perspective/pyproject.toml
```diff
@@ -1,2 +1,10 @@
...
+]
+
```
timkpaine, line 8: `build-backend = "jupyter_packaging.build_api"`

timkpaine (changes requested):

timkpaine:
moving to https://github.com/finos/perspective/pull/2136 so its a branch on main repo

PR #2128, opened 2023-03-05
New demo `market`

texodus:
Adds a new Perspective examples `market`, a JS only reimagining of #1277.  This multi-process example implements order clearing, expiration and pricing as Perspective queries.

[Demo here!](https://prospectiveco.github.io/tm-redesign-demo/)

Other fixes in this branch:

* Adds `min` and `max` aggregate functions.  Unlike `low` and `high` aggregates, these are _always_ recalculated, which means they track the range for the current dataset (as opposed to all data entries to ever exist in the dataset, even if they've been updated or removed).
* Fixed bug which printed the string "null" over null label points in "X/Y Scatter" charts.
* Fixed Candlestick and OHLC charts to allow <4 columns, where the missing fields are omitted (in the case of high/low) or synthesized (as the next open in the case of close).
* Fix charts to only grow extents on `update()`, which leads to a less janky visual experience when the dataset changes range quickly.  The range is reset on _any_ interaction.
* Fix `load()` to not overwrite default `column` on 2nd+ calls if they have already been set.
* Fix drag & draop implementation on iOS (remove `mobile-drag-drop` polyfill and adapt native HTML5 drag/drop behavior to work on iOS).
* Add `query_columns()` method to `Table`, which emulates `view()` + `to_columns()` + `delete()` in a single event loop dispatch.

[Screenshot 2023-03-05 at 11 19 02 AM](https://prospectiveco.github.io/tm-redesign-demo/)


examples/blocks/src/market/README.md
```diff
@@ -0,0 +1,7 @@
+A simple market simulation
+
+-   State - The simulation is implemented as queries into a Persoective `Table`
```
timkpaine, line 3: Typo

PR #2145, opened 2023-03-20
run all lints and fixes during commit preparation, alphabetize and reorder/prune lint commands

timkpaine:
https://github.com/finos/perspective/pull/2136#discussion_r1141498027

This PR redoes the linting to enforce all linted file types, to avoid issues like the above where a routing change involves excess noise due to trying to autoformat. The first commit is substantive, all subsequent commits are running auto formatting.


cpp/perspective/src/cpp/sparse_tree.cpp
```diff
@@ -1303,12 +1303,13 @@ t_stree::update_agg_table(t_uindex nidx, t_agg_update_info& info,
...
                 old_value.set(dst_scalar);
                 // new_value.set(std::max(dst_scalar, src_scalar));
-                // if (is_expr || old_value.is_nan() || new_value.is_nan() || new_value > old_value) {
-                    auto pkeys = get_pkeys(nidx);
-                    std::vector<double> values;
-                    read_column_from_gstate(gstate, expression_master_table,
-                        spec.get_dependencies()[0].name(), pkeys, values, true);
-                    new_value.set(*std::max_element(values.begin(), values.end()));
+                // if (is_expr || old_value.is_nan() || new_value.is_nan() ||
+                // new_value > old_value) {
```
hibnite, line 1307: An alternative would be to uncomment the `{}`, each on a line by itself. This would allow the indentation to remain, with the commented-out `if` clause.  To enable, one would just need to un-comment the `if` clause.

The only downside is that it wouldn't follow the hanging-bracket style.

PR #2152, opened 2023-03-25
Improve `group_by` performance by inlining `boost::hash_range`

texodus:
Under profile, calls to `strlen` and `strcmp` feature prominently, as Perspective does not currently share `t_vocab` between table and view instances, these are currently re-interned as `char*`.  This should _definitely_ be fixed and should represent a significant performance improvement (AFAIK only the `join` aggregate can yield a not-yet-interned string result).

In the meantime, this single `boost::hash_range` call dominates view creation, as the index maps are built in this step.  First, it iterates the characters to calculate the range length, then _again_ to calculate the hash.  Replacing this with a single-pass hash calculate dedicated to C strings halves this computation time, as well as removes an unnecessary call to boost.

untitled (4)

untitled (3)

### DIFF
--- a/cpp/perspective/src/cpp/vocab.cpp
+++ b/cpp/perspective/src/cpp/vocab.cpp
@@ -73,9 +73,9 @@ t_vocab::get_interned(const char* s) {
     t_sidxmap::iterator iter = m_map.find(s);
 
     t_uindex idx, bidx, eidx;
-    t_uindex len = strlen(s) + 1;
 
     if (iter == m_map.end()) {
+        t_uindex len = strlen(s) + 1;
         idx = genidx();
 
         bidx = m_vlendata->size();
--- a/cpp/perspective/src/include/perspective/base.h
+++ b/cpp/perspective/src/include/perspective/base.h
@@ -442,7 +442,11 @@ struct t_cchar_umap_cmp
 struct t_cchar_umap_hash {
     inline t_uindex
     operator()(const char* s) const {
-        return boost::hash_range(s, s + std::strlen(s));
+        size_t h = 5381;
+        int c;
+        while ((c = *s++))
+            h = ((h << 5) + h) + c;
+        return h;
     }
 };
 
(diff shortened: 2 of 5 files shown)


cpp/perspective/src/cpp/vocab.cpp
```diff
@@ -73,9 +73,9 @@ t_vocab::get_interned(const char* s) {
     t_sidxmap::iterator iter = m_map.find(s);
 
     t_uindex idx, bidx, eidx;
-    t_uindex len = strlen(s) + 1;
 
     if (iter == m_map.end()) {
+        t_uindex len = strlen(s) + 1;
```
tomjakubowski, line 78: Even better would be updating this function's signature to accept a `size_t s_len` and use that here.  The only caller has a `std::string`, and can call `std::string::size()` (which is cheap, constant time) instead of this function making a potentially expensive strlen() call

tomjakubowski, line 78: could save this strlen() for callers which already know the string's length (e.g. anything with a `std::string`) by providing an overload which takes a `size_t s_len`

sinistersnare:
The URL in the `perspective-bench` `package.json` is wrong.

It is

> https://github.com/finos/perspective/packages/perspective-bench

And should be 

> https://github.com/finos/perspective/tools/perspective-bench

texodus:
@sinistersnare I believe it should actually be [https://github.com/finos/perspective/tree/master/tools/perspective-bench](https://github.com/finos/perspective/tree/master/tools/perspective-bench), but not something this PR introduced either way.

cpp/perspective/src/cpp/vocab.cpp
texodus: Good point - but for the records, in this instance `t_vocab` is only used for values that can be `t_scalar`, which does not include `std::string`.

texodus:
Closing, these results are not reproducible on master

PR #2157, opened 2023-03-28
gracefully handle websocket errors on python side, fixes #2132

### LINKED ISSUE #2132: StreamClosedError logged when disconnecting/reconnecting
### Steps to Reproduce:

1. run the `python-tornado-streaming` example
2. refresh the webpage very quickly (like press CTRL+R very fast)

I found it is easier to reproduce if you reduce the size of the data set, from 2500 to 25 and reduce the `def data_source()` update size as well.

### Expected Result:

it should work fine

### Actual Result:

I can see this error in stdout:

```
2023-03-09 14:52:02,172 - asyncio - ERROR - Task exception was never retrieved
future: <Task finished name='Task-1860' coro=<WebSocketProtocol13.write_message.<locals>.wrapper() done, defined at /opt/pysetup/.venv/lib/python3.10/site-packages/tornado/websocket.py:1090> exception=WebSocketClosedE
rror()>
Traceback (most recent call last):
  File "/opt/pysetup/.venv/lib/python3.10/site-packages/tornado/websocket.py", line 1092, in wrapper
    await fut
tornado.iostream.StreamClosedError: Stream is closed

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/opt/pysetup/.venv/lib/python3.10/site-packages/tornado/websocket.py", line 1094, in wrapper
    raise WebSocketClosedError()
tornado.websocket.WebSocketClosedError
```

### Environment:

- perspective-python==1.9.4 
- Python 3.10.10
- macOs (but also on docker with `python:3.10.10-slim-bullseye`)

### Additional Context:

This doesn't impact server or the client. But it's an issue for us because we operate perspective in production and ERROR message trigger an alert which is hard to filter.

I believe it comes from a something wrong in how handler/torando.py calls write_message in https://github.com/finos/perspective/blob/master/python/perspective/perspective/handlers/tornado.py#L50

timkpaine:
Disconnect-related errors were handled for `aiohttp` and `starlette` when waiting for the client to send a message, this PR adds that same support to tornado but also handles disconnect when trying to send outbound messages. 

**Note** While testing, I noticed the route to the themes is incorrect, this is particularly noticeable after the recent theme changes

### DIFF
--- a/python/perspective/perspective/handlers/aiohttp.py
+++ b/python/perspective/perspective/handlers/aiohttp.py
@@ -6,7 +6,7 @@
 # the Apache License 2.0.  The full license can be found in the LICENSE file.
 #
 
-from aiohttp import web, WSMsgType
+from aiohttp import web, WSMsgType, WebSocketError
 
 from .common import PerspectiveHandlerBase
 
@@ -48,10 +48,14 @@ class PerspectiveAIOHTTPHandler(PerspectiveHandlerBase):
             self.on_close()
 
     async def write_message(self, message: str, binary: bool = False) -> None:
-        if binary:
-            await self._ws.send_bytes(message)
-        else:
-            await self._ws.send_str(message)
+        try:
+            if binary:
+                await self._ws.send_bytes(message)
+            else:
+                await self._ws.send_str(message)
+        except WebSocketError:
+            # Ignore error
+            ...
 
     # Use common docstring
     __init__.__doc__ = PerspectiveHandlerBase.__init__.__doc__
--- a/examples/python-tornado-streaming/index.html
+++ b/examples/python-tornado-streaming/index.html
@@ -23,7 +23,7 @@
         <link
             rel="stylesheet"
             crossorigin="anonymous"
-            href="/node_modules/@finos/perspective-viewer/dist/css/themes.css"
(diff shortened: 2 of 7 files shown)


python/perspective/perspective/handlers/aiohttp.py
```diff
@@ -48,10 +48,14 @@ async def run(self) -> None:
...
-        if binary:
-            await self._ws.send_bytes(message)
-        else:
-            await self._ws.send_str(message)
+        try:
+            if binary:
+                await self._ws.send_bytes(message)
+            else:
+                await self._ws.send_str(message)
+        except WebSocketError:
```
texodus, line 56: I don't know enough about the ecosystem of possible exceptions here, but I have spent countless hours of my life chasing bugs that would've been immediately obvious had they not been swallowed by a blanket catch block. At minimum, some comments about what to expect in these cases and why would be super helpful - but I propose we generally try to handle specific exceptions instead, which would also communicate intent here.

texodus (approved):

python/perspective/perspective/handlers/aiohttp.py
texodus: Maybe also in this case merits a `debug` log, now that I think about it.

PR #2164, opened 2023-04-04
Add build for python 3.11

tomjakubowski:
Add support for python 3.11

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -777,12 +777,15 @@ jobs:
                     - is-full-run: false
                       os: macos-11
 
-                    # Exclude Python 3.8, 3.10 builds
+                    # Exclude Python 3.8, 3.10, 3.11 builds
                     - is-full-run: false
                       python-version: 3.8
 
                     - is-full-run: false
                       python-version: "3.10"
+
+                    - is-full-run: false
+                      python-version: 3.11
 
             # NOTE: don't fail fast as sometimes npm blocks the burst of fetches from GHA
             fail-fast: false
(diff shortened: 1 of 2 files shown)


timkpaine:
https://github.com/conda-forge/perspective-feedstock/blob/main/recipe/meta.yaml#L37

tomjakubowski:
Oh, is this going to break something on Windows?

timkpaine:
> Oh, is this going to break something on Windows?

it might, needs to be tested on all platforms with full builds

texodus (changes requested):
Thanks for the PR!

Let's go ahead and actually do 3.11 builds in GitHub Actions - no point in merging this separately.

texodus:
lol I meant on Monday :)

[full build here](https://github.com/finos/perspective/blob/master/scripts/publish.js#L61)

We'll also need to update the [publish scripts](https://github.com/finos/perspective/blob/master/scripts/publish.js#L61) for these packages to actually be disted.

texodus:
This looks like it's dying on Windows due to [some link error with boost](https://github.com/finos/perspective/actions/runs/4841715483/jobs/8628280629#step:22:924)

For posterity when this log is flushed, here's the first missing file:

> D:\a\perspective\perspective\python\perspective\dist\src\include\perspective/step_delta.h(22,10): fatal  error C1083: Cannot open include file: 'boost/variant/apply_visitor.hpp': No such file or directory (compiling source file D:\a\perspective\perspective\python\perspective\dist\src\cpp\binding.cpp) [D:\a\perspective\perspective\python\perspective\build\temp.win-amd64-cpython-38\Release\psp.vcxproj]

texodus:
Cherry-picked pybind updates to #2234

Issue #1127, opened 2020-07-17
Datagrid gets transposed (axis are swapped) when plugin is changed from "Grid" to "Heatmap"


pgagarinov:
#### 1. 
```
import pandas as pd
import numpy as np
from perspective import PerspectiveWidget
from datetime import timedelta,date
arrays = {'A':['bar', 'bar', 'bar', 'bar', 'baz', 'baz', 'baz', 'baz', 'foo', 'foo', 'foo', 'foo', 'qux', 'qux', 'qux', 'qux'],
          'B':['one', 'one', 'two', 'two', 'one', 'one', 'two', 'two', 'one', 'one', 'two', 'two', 'one', 'one', 'two', 'two'],
          'C':['X', 'Y', 'X', 'Y', 'X', 'Y', 'X', 'Y', 'X', 'Y', 'X', 'Y', 'X', 'Y', 'X', 'Y'],
          'D':np.arange(date(2020,6,1),date(2020,6,17),timedelta(days=1),dtype=date)}

df = pd.DataFrame(arrays)
df.set_index(['D'])
```
<img width="165" alt="image" src="https://user-images.githubusercontent.com/4868370/87778905-96ad0f00-c834-11ea-8946-54ac3393d917.png">

#### 2. 
```
psp = PerspectiveWidget(df, row_pivots=['D'], column_pivots=['A','B'],aggregates={'C':'count'},columns=['C'])
psp
```
<img width="607" alt="image" src="https://user-images.githubusercontent.com/4868370/87778999-c52aea00-c834-11ea-9086-2230769bd528.png">
So far so good...

#### 3. 
Let us change "Grid" to "Heatmap" on UI (by hand):

<img width="1382" alt="image" src="https://user-images.githubusercontent.com/4868370/87779122-f6a3b580-c834-11ea-8f46-4d816d94f2a3.png">

We see that the axis are swapped while the column names in "groupBy" and "splitBy" remained the same.
This is incorrect.

### Expected Result:
<img width="1077" alt="image" src="https://user-images.githubusercontent.com/4868370/87779318-4aae9a00-c835-11ea-9a9b-dfb992e98283.png">

### Environment:
Perspective 0.5.1

timkpaine:
I suspect this is a duplicate of #1122, if you turn it into a flattened grid you should see the columns are all messed up to begin with.

pgagarinov:
[...]

PR #2186, opened 2023-04-13
fix "Data" to "Heatmap" axes transposing-1127

zxy994:
#1127 Datagrid gets transposed (axis are swapped) when plugin is changed from "Grid" to "Heatmap" #1127

[https://github.com/finos/perspective/issues/1127]

BEFORE:

https://user-images.githubusercontent.com/38087267/231719273-e1d65e1b-f7d3-4bae-b517-8d51200530aa.mp4

AFTER:

https://user-images.githubusercontent.com/38087267/231719332-166b59bb-da4c-4294-a1d3-fcdb5244e7d1.mp4

### DIFF
--- a/packages/perspective-viewer-d3fc/src/js/charts/heatmap.js
+++ b/packages/perspective-viewer-d3fc/src/js/charts/heatmap.js
@@ -18,6 +18,52 @@ import { colorRangeLegend } from "../legend/colorRangeLegend";
 import zoomableChart from "../zoom/zoomableChart";
 import nearbyTip from "../tooltip/nearbyTip";
 
+export const invertAxesFlag = false;
+
+function invertAxes(flag, settings, data) {
+    let xAxis = axisFactory(settings)
+        .excludeType(AXIS_TYPES.linear)
+        .settingName("crossValues")
+        .valueName("crossValue")(data);
+
+    let yAxis = axisFactory(settings)
+        .excludeType(AXIS_TYPES.linear)
+        .settingName("splitValues")
+        .valueName("mainValue")
+        .modifyDomain((d) => {
+            let is_number = !isNaN(d[0]);
+            return is_number ? d.reverse() : d;
+        })
+        .orient("vertical")(data);
+
+    if (flag === true) {
+        xAxis = axisFactory(settings)
+            .excludeType(AXIS_TYPES.linear)
+            .settingName("splitValues")
+            .valueName("crossValue")
+            .modifyDomain((d) => {
+                let is_number = !isNaN(d[0]);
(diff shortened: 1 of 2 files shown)


mcleo-d:
Hi @maoo and @texodus - Would it be possible to review this PR and kick off the GitHub Action below to help unblock and merge @zxy994's PR. Many thanks!

https://github.com/finos/perspective/actions/runs/4687813968

Screenshot 2023-04-18 at 11 00 31

texodus:
@zxy994 @mcleo-d This PR is much appreciated! 

... but the issue it stems from is out-of date. I [posted a comment in the original issue](https://github.com/finos/perspective/issues/1127#issuecomment-1505278556) explaining the reasoning here , #2113 is the correct fix which pre-dates this PR.

I apologize, this was extremely sloppy book-keeping on my end:

* I should've updated the issue sooner.
* #2113 I merged didn't close the issue or contain a description (I'll rectify this).

We're committed to making this situation better - clean up the backlog, publish a roadmap, create a chatroom community.

Back to the PR - @zxy994 I unfortunately [can't merge it](https://github.com/finos/perspective/issues/1127#issuecomment-1505278556) as-is, but if you're willing to push this change a bit further, I would like to preserve the functionality for another feature we're about to contribute from [Prospective Co.](https://www.prospective.co/) - Chart Config UI. While this feature won't be ready for a bit longer, if I can have this PR off-by-default, I can merge it now and re-enable it when there is a configuration screen implemented for the chart type. I would accept either of 2 forms of this -

1. Easy mode - feature flag. Just put a conditional around this functionality which defaults to `false`.  We will write a new Issue describing the next steps and re-enable this later.
2. Hard mode - `plugin_config`. Perspective plugins have their own custom persistence layer represented by the `plugin_config` config key.  We could add a custom settings for heatmaps (e.g. `invert_axis`) that is `false` by default and allows adventurous users to enable this functionality through the [Plugin API](https://github.com/finos/perspective/blob/master/rust/perspective-viewer/src/ts/plugin.ts#L161). This is the same general idea as a feature flag, but involves wiring the state change through a few additional files.

If this is out of scope, we can reconsider this PR when we have a configuration screen to support it.

texodus (changes requested):
As per comment - we need (at least) feature-flagged off.

zxy994:
@texodus Thanks for the review! I'll proceed to adding a conditional around this functionality which defaults to false.

texodus:
Apologies, we're going to be taking this component in a different direction and this PR will be immediately squashed by this work, so closing (its non functional regardless)

zxy994:
@texodus Sure!

PR #2187, opened 2023-04-13
pass through websocket connection arguments in python websocket clients, normalize max message size and ensure set for aiohttp

### LINKED ISSUE #2185: PerspectiveAIOHTTPWebsocketConnection does not set max_msg_size
## Bug Report

The `connect` method receives `max_message_size` (which is passed in as `1024*1024*1024`), but then does not set that in the `ws_connect` method, which *should* be set thus:

```
 self._ws_cm = self._session.ws_connect(url, max_msg_size=max_message_size)
```

### Steps to Reproduce:
 1. Instantiate a `PerspectiveAIOHTTPClient`, connecting to a perspective server websocket, and loading a large view (> 4194304 bytes (4MB), which is the default `max_msg_size` for an aiohttp websocket):
```
client = await aiohttp_websocket("wss://server")
table = client.open_table("my_table")
view = await table.view(columns=["col1", "col2", "col3"]
records = await view.to_records()
```

### Expected Result:
`records` is populated with the view data

### Actual Result:
The aiohttp websocket silently disconnects (which is another problem as the exception is never bubbled up) and the coroutine never seems to yield anything, with the asyncio loop seemingly stuck forever.

### Environment:
Python 3.9.13 (though python version does not really matter here)
perspective-python 1.9.4

### Additional Context:
For anyone else who experiences this, it can be gotten around in the meanwhile by monkey patching the `connect` method of `PerspectiveAIOHTTPWebsocketConnection`

timkpaine:
fixes #2185

### DIFF
--- a/python/perspective/perspective/client/aiohttp.py
+++ b/python/perspective/perspective/client/aiohttp.py
@@ -48,8 +48,13 @@ class PerspectiveAIOHTTPWebsocketConnection(PerspectiveWebsocketConnection):
             elif msg.type == aiohttp.WSMsgType.CLOSE:
                 return
 
-    async def connect(self, url, on_message, max_message_size) -> None:
-        self._ws_cm = self._session.ws_connect(url)
+    async def connect(self, url, on_message, **websocket_connect_kwargs) -> None:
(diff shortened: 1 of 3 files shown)


texodus (approved):

PR #2192, opened 2023-04-17
New `playwright` test suite

texodus:
Forked from #2170 

* Ports _all_ tests and removes `jest` entirely.
* Archives snapshot files.
* Splits "standard" tests to individual runtimes.
* Disables jupyterlab (accidently everywhere) - but this should only run during the `test_js_and_python` phase.
* Force reload browser between tests (otherwise test order impacts the results, as the configs are only partial settings).
* Fixes broken tests, `ts-ignore` annotations, test bugs & general cleanup.

There are still test failures.  Some are `style` attributes with `px` position fields, some are mis-placed text-overflow ellipses, some I have not debugged.  I'm not sure how to deal with the ellipsis - perhaps disable them globally via CSS rule in the test? - but the `style` attribtues can be cleaned with regex as we do for other fields.  Otherwise, this seems stable now.

I suspect a lot of weight can be cut from the `build.yml` with this change, I'd be interested in this cleanup landing as well.


texodus:
Closing for #2196

PR #2197, opened 2023-04-20
Add type-specific aggregates for median and middle

### LINKED ISSUE #1928: Shouldn't the median value for numbers 1 and 2 be equal to 1.5 instead of 2?
## Support Question

Hi!

During the calculation of the aggregation as the median of an even number of values, I noticed that the result is calculated as the value of the element of the sequence, which is at the position calculated by the "formula" - **Median = n/2 + 1**, where **n** - is the number of elements in the sequence.

Example:

1, 2 - Median = 2
5, 8, 9, 15 - Median = 9

Based on this [source](https://www.scribbr.com/statistics/median/#:~:text=To%20find%20the%20median%2C%20calculate,and%20dividing%20them%20by%20two.) I expect the Median to be 1.5 and 8.5 respectively.

Please tell me if this is the expected behavior of the library? If so, do you have any reference to explain why that logic was chosen.

Thank you very much and wish you good luck :)

I am also attaching a piece of JavaScript code for reproducing:

    const data = [
      {
        Fruit: 'Apple',
        Price: 1,
      },
      {
        Fruit: 'Apple',
        Price: 2,
      },
    ];
    const options: ViewConfig = {
      columns: ['Price'],
      aggregates: { Price: 'median' },
      group_by: ["Fruit"],
    };
    const schema =
    {
      Gender: 'Fruit',
      Price: 'float'
    };
    const table = await perspective.worker().table(schema);
    table.update(data)
    const view = await table.view(options)
    const aggregationResult = await view.to_json();

_aggregationResult here is '[{"__ROW_PATH__":[],"Price":2},{"__ROW_PATH__":[null],"Price":2}]'_

orimir:
Refactor median aggregate to support two type-specific aggregates (median and middle) to ensure consistent behavior for numeric and category columns. Type dispatching is performed internally to handle even and odd numbers of values. This simplifies and improves the implementation. fixes #1928

Javascript code to reproduce:
```javascript
const perspective = require("@finos/perspective")
const worker = perspective.worker()
const main = async () => {
    const data = [
        {
            Fruit: 'Apple',
            Price: 1,
        },
        {
            Fruit: 'Apple',
            Price: 2,
        },
        {
            Fruit: 'Apple',
            Price: 4,
        },
        {
            Fruit: 'Orange',
            Price: 5,
        },
        {
            Fruit: 'Orange',
            Price: 6,
        },
    ];
    const options = {
        columns: ['Price'],
        aggregates: { Price: 'median' },
        group_by: ['Fruit'],
    };
    const schema = {
        Fruit: 'string',
        Price: 'float'
    };
    const table = await worker.table(schema);
    table.update(data)
    const view = await table.view(options)
    const aggregationResult = await view.to_json();
    console.log(aggregationResult);
}; 
main();
```
**Aggregation Result**
```
[
  { __ROW_PATH__: [], Price: 4 },
  { __ROW_PATH__: [ 'Apple' ], Price: 2 },
  { __ROW_PATH__: [ 'Orange' ], Price: 5.5 }
]
```

### DIFF
--- a/cpp/perspective/src/cpp/sparse_tree.cpp
+++ b/cpp/perspective/src/cpp/sparse_tree.cpp
@@ -1120,7 +1120,17 @@ t_stree::update_agg_table(t_uindex nidx, t_agg_update_info& info,
                             std::nth_element(
                                 values.begin(), middle, values.end());
 
-                            return *middle;
+                            if (values.size() % 2 == 0) {
+                                // If there are an even number of values,
+                                // return the average of the two middle elements
+                                t_tscalar median1 = *middle;
+                                t_tscalar median2 = *(middle - 1);
+                                return (median1 + median2) / 2.0;
+                            } else {
+                                // If there are an odd number of values,
+                                // return the middle element
+                                return *middle;
+                            }
                         }
                     }));
 


orimir:
@cla-bot[bot] check

TheJuanAndOnly99:
@cla-bot[bot] check

texodus:
Thanks for the PR!

This doesn't implement the described feature - it either needs to preserve the old behavior for non-numeric types via type-dispatch, or implement a secondary num-type-only version which implements the special case for even elements. It also fails the test suite due to the change behavior, and of course needs additional tests to cover whichever solution you propose.

texodus (changes requested):
[ ] Needs both numeric and non-numeric implementations via some method
[ ] Fix tests
[ ] Add test coverage for the new cases

texodus:
Superseded by #2224

PR #2201, opened 2023-04-21
New `playwright` test suite

texodus:
Previously #2196, #2192, #2170

* Ports _all_ tests and removes `jest` entirely.
* Archives snapshot files.
* Splits "standard" tests to individual runtimes.
* Disables jupyterlab (accidently everywhere) - but this should only run during the `test_js_and_python` phase.
* Force reload browser between tests (otherwise test order impacts the results, as the configs are only partial settings).
* Fixes broken tests, `ts-ignore` annotations, test bugs & general cleanup.
* Removes some unnecessary steps from CI.


texodus:
Thanks for the PR @brochington!  Looks good!

PR #2206, opened 2023-04-25
Allow multiplicities in bucketing expressions

sinistersnare:
This PR allows you to do things like `bucket("Order Date", '3M')` in the ExprTK expressions.

One of the down sides of this approach is it loses a little bit of prettiness. Before, bucketing weeks (`bucket("...", 'W')`) would always bucket 7 days, but now it will make buckets with the following start days: `1, 8, 15, 22, 29`. This change may confuse people. It also does not cross barriers. `bucket("...", '27D')` will bucket days 1-27, and then the last couple days of the month, not the next 27 days, bleeding into the next month. Should we do these buckets another way? We can maintain the behavior from before where '3W' would give 3 7-day weeks, but the code for that is much more hairy.

To reiterate, due to the nature of the change, this changes the output of `bucket` even when the multiplicity is `1`.

TODO: 
- [ ] Only evaluate the regex once per column
- [ ] Test that it actually works
- [ ] Actually test it
- [ ] clean up code
- [ ] Write docs for `bucket` function.


timkpaine:
I don't think there's a ton of utility in a fixed day-of-week alignment. If anything it should align to calendar weeks with either Monday or Sunday start (configurable). Even better would be to specify "weeks" vs "weekly", the former should be offset based and the latter aligned to calendar week. There's no need to stick to NumChar syntax, we can make it (col, num, variant).

texodus:
These cases cover what we need, are even factors, and don't require regex:
* `5s`, `10s`, `15s`, `20s`, `30s`
* `5m`, `10m`, `15m`, `20m`, `30m`
* `2h`, `3h`, `4h`, `6h`, `8h`, `12h`
* `3M`, `4M`, `6M`
* `5Y`, `10Y`

As you say, cases like `'27D'` don't make sense. There's no obvious way to support the these cases without more context, and there's not an obvious use for them in a data visualization - so let's disallow them at type validation.

We definitely don't want to change the behavior of `'W'`, this is a common visual bucketing factor. I would propose instead we implement `'2W'` and `'6W'` as special-cases of "1/2 month" and "1.5 months" respectively, which are pinned to the 1st and 15th, while `W` remains as it is now (whatever the current locale defines as the first day of the week).

If you really want symmetry - you can do the math yourself in posix time with `datetime` columns and it would be easy to add for `date`:

```
// 27D Bucket
var day := 1000 * 60 * 60 * 24;
datetime(bucket(float("Order Date"), day * 27))
```

sinistersnare:
Ok, for now I have removed `day` and `week` bucketing because they have ugly edge cases that need to be worked out.

I have also added tests, improved `bucket` documentation, and swapped the regex for manually parsing the unit string.

cpp/perspective/src/cpp/computed_function.cpp
```diff
@@ -1121,11 +1121,11 @@ namespace computed_function {
...
-    tsl::hopscotch_map<std::string, t_date_bucket_unit> bucket::UNIT_MAP = {
-        {"s", t_date_bucket_unit::SECONDS}, {"m", t_date_bucket_unit::MINUTES},
-        {"h", t_date_bucket_unit::HOURS}, {"D", t_date_bucket_unit::DAYS},
-        {"W", t_date_bucket_unit::WEEKS}, {"M", t_date_bucket_unit::MONTHS},
-        {"Y", t_date_bucket_unit::YEARS}};
+    tsl::hopscotch_map<char, t_date_bucket_unit> bucket::UNIT_MAP = {
+        {'s', t_date_bucket_unit::SECONDS}, {'m', t_date_bucket_unit::MINUTES},
+        {'h', t_date_bucket_unit::HOURS}, {'D', t_date_bucket_unit::DAYS},
+        {'W', t_date_bucket_unit::WEEKS}, {'M', t_date_bucket_unit::MONTHS},
+        {'Y', t_date_bucket_unit::YEARS}};
```
texodus, line 1128: :100:

rust/perspective-viewer/exprtk.md
```diff
@@ -37,6 +37,22 @@ Bucket x by y
...
+-   `h` -- hours
+-   `D` -- days
+-   `W` -- weeks
+-   `M` -- months
+-   `Y` -- years
+
+In all of these except for days and weeks, a non-negative number can be placed before the unit character.
+For example, `3M` would bucket by 3 months. These must divisible numbers; you can not bucket by
+`7M` or `22s`, but by `30m`, `12h`, etc.
+
```
texodus, line 55: This file is auto-generated from the [rust source](https://github.com/finos/perspective/blob/emsdk-update/rust/perspective-viewer/src/rust/exprtk/language.rs#L42). Currently, this same text is used to generate the auto-complete hints and it does not support markdown, so this text is not currently suitable there either.

We'll need to add something like a short/long text description field to the exprtk language spec to support this.

Screenshot 2023-04-29 at 5 26 39 PM

tools/perspective-test/results.tar.gz
texodus, line 1: This file shouldn't have been altered (but does indeed pass).  Not sure what's going on here yet ...

texodus (commented):
Thanks for the PR!

The implementation and tests look great! I'm going to close this PR for #2210 which addresses the docs and test artifact issues noted.

PR #2208, opened 2023-04-28
add parquet viewer support to jupyterlab mime renderer 

timkpaine:
Adds JupyterLab mimerender viewer for `.parquet` file using @kylebarron's https://github.com/kylebarron/parquet-wasm library. Note that since we're already compiling arrow to wasm anyway, we could consider also compiling the parquet bits and letting perspective natively read parquet files as it does arrow ipc data.

I've also renamed the context menu labels from e.g. `CSVPerspective` to `Perspective-CSV`, because I think it looks better. 

tmp


kylebarron:
This is cool! As a note I made some updates for 0.4.0-beta.5 but then hit some issues with webpack where I wasn't sure if it was a regression or not and then never ended up making an 0.4.0 release. In case you need an async API with remote files, you might want to try that latest version

PR #2209, opened 2023-04-28
rename context menu names for perspective renderers in jupyterlab

timkpaine:
I think these names look a bit cleaner than the original ones.


texodus (approved):

PR #2213, opened 2023-05-02
Moving Expression Editor modal to a sidebar.

sinistersnare:
I rewrote the expr-editor components a bit, simplifying their implementation. Then some of the code was so simple, I just converted over to function components because the song-and-dance of struct-components was unneeded.


texodus:
Closing in favor of #2239 which rebases this PR

Issue #2036, opened 2022-12-06
Column expr of `integer` column sum renders all rows as 0.


sinistersnare:
## Bug Report

When creating a copy of a column by doing `"ColumnName"` as the entire expression text, it calculates the sum wrong.

### Steps to Reproduce:
 1. Go to the editable example
 2. Set zipcode as the pivot (group-by)
 3. Create a copy of the Quantity column (expr exactly `"Quantity"`)
 4. Note that the 'real' quantity and the copy are different when marked as sums.

### Expected Result:

The real sum, not 0.

### Actual Result:

![screenshot](https://i.imgur.com/Cpu81M2.png)

### Environment:

Firefox and Chrome

### Additional Context:

```
{"plugin":"Datagrid","plugin_config":{"columns":{},"editable":false,"scroll_lock":true},"settings":true,"theme":"Material Light","group_by":["State"],"split_by":[],"columns":["Quantity","\"Quantity\""],"filter":[],"sort":[],"expressions":["\"Quantity\""],"aggregates":{}}
```

zxy994:
Hi @texodus , I have been able to reproduce this issue and would like to give it go. When I delve further, I find rust files that contain expression_editor.rs, which emits a JSValue callback function when the save button to add an expression column is clicked. Perhaps I am looking in the wrong direction. Could you please point me in the right direction for the project files representing the JavaScript execution environment that handles the addition of expression columns?

PR #2216, opened 2023-05-06
Fix `sum` aggregate for `integer` type expression columns

zxy994:
#2036 Column expr of integer column sum renders all rows as 0

BEFORE:
image

AFTER:
image


texodus (approved):

I had some comments for the draft version of this PR but you've addressed them and added some additional tests :)

zxy994:
@texodus Thanks for the review!

Issue #2212, opened 2023-05-01
Broken Link in Docs


riziles:
I believe the following link:

https://github.com/finos/perspective/blob/e89ae0ab2aca1d903a0d3738ee2e8d9f67c1c684/packages/perspective/README.md?plain=1#L7

which shows up in the docs here: https://perspective.finos.org/docs/obj/perspective/

should point here instead:

https://perspective.finos.org/docs/js/

ayushkcs:
Hi @riziles @timkpaine ,
Can I please try to fix this?
Thanks

texodus:
Appreciate the report @riziles, we'll fix this.

Closing as to dissuade future PRs that exclusively update links or spelling - the notification of the issue is enough for now.

PR #2219, opened 2023-05-09
Updated broken links

singhalashish16:
Fixed broken links mentioned in Issue
#2212 

- Updated  [perspective/packages/perspective/README.md](https://github.com/finos/perspective/blob/e89ae0ab2aca1d903a0d3738ee2e8d9f67c1c684/packages/perspective/README.md?plain=1#L7)

- Updated [docs/src/components/HomepageFeatures/python.md](https://github.com/finos/perspective/blob/master/docs/src/components/HomepageFeatures/python.md)

### DIFF
--- a/docs/src/components/HomepageFeatures/python.md
+++ b/docs/src/components/HomepageFeatures/python.md
@@ -1,7 +1,7 @@
 ## Python
 
 \`perspective-python\`, built on the same C++ data engine used by the
-[WebAssembly version](https://perspective.finos.org/docs/md/js.html), implements
+[WebAssembly version](https://perspective.finos.org/docs/obj/perspective-python/), implements
 the Perspective API directly in Python, either as a virtualized server for
 Production, or as an embedded JupyterLab Widget for Research.
 
--- a/packages/perspective/README.md
+++ b/packages/perspective/README.md
@@ -4,7 +4,7 @@
 The main API module for `@finos/perspective`.
 
 For more information, see the
-[Javascript user guide](https://perspective.finos.org/docs/md/js.html).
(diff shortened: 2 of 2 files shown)


docs/src/components/HomepageFeatures/python.md
```diff
@@ -1,7 +1,7 @@
 ## Python
 
 \`perspective-python\`, built on the same C++ data engine used by the
-[WebAssembly version](https://perspective.finos.org/docs/md/js.html), implements
+[WebAssembly version](https://perspective.finos.org/docs/obj/perspective-python/), implements
```
sinistersnare, line 4: This seems like the wrong link. It used to link to a `/js` page but now `/python`

singhalashish16: The page demands link to be directed to perspective-python api doc.
The current https://perspective.finos.org/docs/md/js.html gives error.
If it shoud direct to another relevent link please give it.

sinistersnare: The [actual Python documentation](https://perspective.finos.org/docs/python/) has similar copy at the top, and links to [The JS docs](https://perspective.finos.org/docs/js/) (https://perspective.finos.org/docs/js/). So I think we should link to the JS docs there. 

However, it probably is a good idea to link to the Python docs here, probably make the `perspective-python` at the start of the paragraph a link to https://perspective.finos.org/docs/python/, and not the API docs themselves?

Regardless of that though, this should be a link to the JS docs, as the Python docs do.

texodus:
Wrong links, as per @sinistersnare.  This is easier to just fix than invite further iteration, so closing.

PR #2220, opened 2023-05-09
Minor DOC fix

ViratCh04:
## Issue
The README.md file for the perspective repository currently has two broken table cells in the **Example** section as can be seen below
image

## Fix
I have removed these two cells as having less two examples is better than displaying two dysfunctional thumbnails(and examples as well, as I later learnt). Originally I was trying to get the thumbnails working but that is no longer possible since the **magic** example is no longer functioning while the **jupyter notebook** example is not available for more requests.

After removing the HTML code for the two cells, The **examples** section is now looking much better

image


texodus:
Thanks for the report, but this is not a fix.

PR #2224, opened 2023-05-17
Fix median miscalculation for even-sized item list

adetsi:
Fix for issue (#1928) where median value for even-sized item list is miscalculated, see below.

### Before
image

### After
image

### DIFF
--- a/cpp/perspective/src/cpp/sparse_tree.cpp
+++ b/cpp/perspective/src/cpp/sparse_tree.cpp
@@ -1127,30 +1127,13 @@ t_stree::update_agg_table(t_uindex nidx, t_agg_update_info& info,
                 new_value.set(reduce_from_gstate<
                     std::function<t_tscalar(std::vector<t_tscalar>&)>>(gstate,
                     expression_master_table, spec.get_dependencies()[0].name(),
-                    pkeys, [](std::vector<t_tscalar>& values) {
+                    pkeys, [&](std::vector<t_tscalar>& values) {
                         if (values.size() == 0) {
                             return t_tscalar();
                         } else if (values.size() == 1) {
                             return values[0];
                         } else {
-                            int size = values.size();
-                            bool is_even_size = size % 2 == 0;
-
-                            if (is_even_size){
-                                t_tscalar median_average;
-                                std::vector<t_tscalar>::iterator first_middle = values.begin() + ((size - 1) / 2);
-                                std::vector<t_tscalar>::iterator second_middle = values.begin() + (size / 2);
-
-                                nth_element(values.begin(),  first_middle, values.end());
-                                nth_element(values.begin(), second_middle, values.end());
-
-                                median_average.set((*first_middle + *second_middle) / static_cast<t_tscalar>(2));
-                                return median_average;
-                            }else{
-                                std::vector<t_tscalar>::iterator middle = values.begin() + (size / 2);
-                                std::nth_element(values.begin(), middle, values.end());
-                                return *middle;
-                            }
+                            return get_aggregate_median(values);
                         }
                     }));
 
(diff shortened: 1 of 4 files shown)


texodus:
@adetsi There is already an open PR for this issue #2197.

That said, neither this PR nor #2197 actually address the issue [I described in my response to #1928](https://github.com/finos/perspective/issues/1928#issuecomment-1367067946), which is that the `median()` aggregate _is also used for non-numeric values_ and therefore we can't naively calculate the average when the column has an even number of elements.

adetsi:
@texodus

Thanks for the clarification,  I have added implementations to accommodate for non-numeric types as well (preserves the old behavior).

adetsi:
@texodus 
Broken test cases have been fixed

cpp/perspective/src/cpp/sparse_tree.cpp
```diff
@@ -2020,6 +2014,28 @@ t_stree::get_aggregate(t_index idx, t_index aggnum) const {
...
+        t_tscalar median_average;
+        std::vector<t_tscalar>::iterator first_middle = values.begin() + ((size - 1) / 2);
+        std::vector<t_tscalar>::iterator second_middle = values.begin() + (size / 2);
+
+        nth_element(values.begin(),  first_middle, values.end());
+        nth_element(values.begin(), second_middle, values.end());
+
+        median_average.set((*first_middle + *second_middle) / static_cast<t_tscalar>(2));
+        return median_average;
+    }else{
```
texodus, line 2032: Formatting - please run `clang-format`.

cpp/perspective/src/cpp/sparse_tree.cpp
```diff
@@ -2020,6 +2014,28 @@ t_stree::get_aggregate(t_index idx, t_index aggnum) const {
     return extract_aggregate(m_aggspecs[aggnum], c, agg_ridx, agg_pridx);
 }
 
+t_tscalar
+t_stree::get_aggregate_median(std::vector<t_tscalar>& values) const {
+    int size = values.size();
+    bool is_even_size = size % 2 == 0;
+
+    if (is_even_size && values[0].is_numeric()){
```
texodus, line 2022: See review summary - `is_numeric()` case here should be exclusively `float` column types via  `is_floating_point()`.

cpp/perspective/src/cpp/sparse_tree.cpp
```diff
@@ -2020,6 +2014,28 @@ t_stree::get_aggregate(t_index idx, t_index aggnum) const {
...
+        t_tscalar median_average;
+        std::vector<t_tscalar>::iterator first_middle = values.begin() + ((size - 1) / 2);
+        std::vector<t_tscalar>::iterator second_middle = values.begin() + (size / 2);
+
+        nth_element(values.begin(),  first_middle, values.end());
```
texodus, line 2027: `nth_element()` does not need to be called twice here, the column is guaranteed to be even and min 2, so this is equivalent to `*(second_middle - 1)`.

cpp/perspective/src/cpp/sparse_tree.cpp
```diff
@@ -1127,19 +1127,13 @@ t_stree::update_agg_table(t_uindex nidx, t_agg_update_info& info,
...
                             return values[0];
                         } else {
-                            std::vector<t_tscalar>::iterator middle
-                                = values.begin() + (values.size() / 2);
-
-                            std::nth_element(
-                                values.begin(), middle, values.end());
-
-                            return *middle;
+                            return get_aggregate_median(values);
```
texodus, line 1136: Just a suggestion - if we're going to factor out this logic into a function called `get_aggregate_median(values)`, maybe we should also move the `0` and `1` cases into this definition to 1) make it complete and 2) move the entire closure body to a single method call.

texodus (changes requested):
Thanks for the update @adetsi.  This implementation is better but the failing tests indicate another issue, that `median` can't be calculated for `integer` columns without changing the aggregate column's output type to `float`; trying to calculate the median without this will generate all 0's.  In order to do this, you'd need to set `t_tscalar median_average` explicitly, as well as setting the associated metadata for the schema transition in a few other places around the codebase. (this is done already for e.g. `count` which goes from any type to `integer`, if you need an example to grep for).

In lieu of this, it would be easier to just implement this behavior for `float` columns instead of _numeric_ ones, by changing the feature-check from `is_numeric` to `is_floating_point`.  `integer` columns will then stay `integer` when applied `median()` and exhibit the old behavior, while columns which are float will use the new behavior. This would be an acceptable alternative if the former implementation for `integer`-column-type-upgrading is too complex (it may involve updating python and the UI as well).

See inline review notes below, there are also a few small style and efficiency improvements, and we should preserve the current test inputs and their integer columns.

cpp/perspective/src/cpp/sparse_tree.cpp
adetsi: Done!

cpp/perspective/src/cpp/sparse_tree.cpp
adetsi: Done!

cpp/perspective/src/cpp/sparse_tree.cpp
adetsi: Done!

adetsi:
@texodus thank you for the review and comments, I hope everything looks good now.

texodus (approved):

texodus:
Thanks for the PR @adetsi! Looks good!

PR #2227, opened 2023-05-18
Pyodide wheel

tomjakubowski:
Passes CI! But, and this is new, trying to _install_ the wheel crashes Pyodide. Suspiciously the wheel is only 2.3M when a wheel built on my machine is 4.4M. Will look into how the wheels are different (size difference might come down to zip implementations/parameters?)

TODOs:

- [ ] Rebase on master
- [x] Get a working wheel in CI
- [x] Build libpsp as a STATIC and back out the other build changes so it's less invasive
- [ ] Probably some debug logs to clean up
- [x] Squash
- [ ] Determine which github action events should trigger the emscripten python ("pyodide") build
- [ ] Use caches where possible in pyodide CI build
- [ ] Possibly merge pyodide CI build into build.yml (at least rename it, since the JS build is already an emscripten build)
- [ ] em++: warning: EXPORTED_FUNCTIONS is not valid with LINKABLE set (normally due to SIDE_MODULE=1/MAIN_MODULE=1) since all functions are exported this mode.  To export only a subset use SIDE_MODULE=2/MAIN_MODULE=2 [-Wunused-command-line-argument]
  - solving that may fix the problems related to having multiple .so's in the wheel
- [ ] final smoke check 
- [x] hunt down why boost headers are logging warnings in pyodide, should be isystem


tomjakubowski:
You can get the latest uploaded wheel here: https://nightly.link/tomjakubowski/perspective/workflows/emscripten/jupyterlite-nasty-v2.0.1/perspective-python-dist-pyodide-311.zip

tomjakubowski:
We can drop `perspective.client` from the pyodide wheel; at least right now I don't think it will be usable or useful in pyodide. The pyodide wheel can then remove dependencies on future, tornado, etc. which are a little awkward to deal with in the browser

texodus:
Closing for #2241

Issue #2203, opened 2023-04-22
Incorrect timezone offset when using datetime filters


ming08108:
## Bug Report

Datetime filters do not respect DST.

### Steps to Reproduce:
1. Create a table with a datetime column
2. Add a where clause on that datetime column

### Expected Result:
Rows should be correctly filtered.

### Actual Result:
Rows are filtered but the filtered time could be off by 1 hour due to DST offsets.

### Environment:
Any

### Additional Context:

Problematic code is likely here: https://github.com/finos/perspective/blob/91102a05032a133e8648032f7f3d7c8f84998076/rust/perspective-viewer/src/rust/utils/datetime.rs#L24

`new Date(0).getTimezoneOffset() != new Date(1682122531000).getTimezoneOffset()` in all timezones.

texodus:
Duplicate #2023

Issue #2023, opened 2022-11-29
Date filtering off by 1 day


rigogsilva:
## Bug Report

Trying to filter by date doesn't really work. It is off by 1 day. 

<img width="735" alt="Screenshot 2022-11-29 at 3 38 27 PM copy" src="https://user-images.githubusercontent.com/31558765/204655260-54085244-814f-4e3b-b631-ebc07aba0b19.png">

### Steps to Reproduce:
 1. Setup perspective table with date field in shema:

```js
import { worker } from "https://cdn.jsdelivr.net/npm/@finos/perspective@1.7.2/dist/cdn/perspective.min.js";

const PERSPECTIVE_CONFIG = {
  types: {
    date: {
      filter_operator: "==",
      aggregate: "count",
      format: {  dateStyle: "short" },
      null_value: -1
    },
  }
};

const worker_ = worker(PERSPECTIVE_CONFIG);
const table = await worker_.table(schema);
table.update(data);
```

2. filter by date as shown in picture above.

### Expected Result:

Should return results with dates 10/23/2022.

### Actual Result:

Returns rows with data 10/22/2022.

### Environment:

* @finos/perspective@1.7.2
* mac m1: ventura

dev-navian:
Hi. This looks interesting and I'd like to work on it. I am still trying to navigate Perspective, but it looks as if the code provided above is incomplete (`schema` and `data` are not defined). Can I be directed to a useful resource or shown where to look to resolve this? I'd appreciate it.

Martinlopez-hash:
Yes I want you to resolve the issue

texodus:
You'll need to load your own data to repro, anything with a `date` column (for example - [superstore demo](https://perspective.finos.org/block/?example=editable)). I reckon the issue lies [here](https://github.com/finos/perspective/blob/master/rust/perspective-viewer/src/rust/components/column_selector/filter_column.rs#L93), as it does not take into account the browser's local time zone ala [here](https://github.com/finos/perspective/blob/master/rust/perspective-viewer/src/rust/utils/datetime.rs#L32).

We'll need tests for this as well.

dev-navian:
Hello @texodus, I have reproduced this at my end and it appears that the filtering works fine (i.e. not off by one day as stated). I even tested it with the example you provided earlier and that works too. I am thus not able to make any modification to the code and to test it. I'd like to know if this is okay or maybe I'm missing something.

Tanushreee:
[...]

PR #2228, opened 2023-05-18
Update datetime.rs for #2203 issue

Himasnhu-AT:
Error Resolved: #2203.
Reasoning: 
By using 1682122531000 (an actual timestamp) instead of 0, you ensure that the DST offset is correctly taken into account.
Assumption: 
Error is with this code snippet.


texodus:
This isn't a fix for this issue.  The code snippet you've copied from the linked issue is an illustration of how the equality doesn't hold, not a suggested implementation. Regardless, this issue is incorrect (or at least unrelated to the bug in question) and is a duplicate of #2023.

Issue #1819, opened 2022-05-18
Null value updates do not propogate to perspective viewer


nickroci:
Non null values in persptive viewer will not get updated with null vales. I tested this from python server. 

I think maybe its related to arrow byte updates as doing a byte update to a table in python also does not update the python table will nulls. It's especially problematic as remove doesn't propogate to ui either in my experience, so you can't really blank out data (I've hacked something into websocket to call remove on ui table as a workaround).

```
import perspective
import pyarrow as pa
import pandas as pd

def get_bytes(tbl):
    stream = pa.BufferOutputStream()
    writer = pa.RecordBatchStreamWriter(stream, tbl.schema)
    writer.write_table(tbl)
    writer.close()
    return stream.getvalue().to_pybytes()

table = perspective.Table([{'a': '1', 'b': '2'}, {'a': '3', 'b': '4'}], index='a')
print(table.view().to_records())

update = get_bytes(pa.Table.from_pandas(pd.DataFrame([{'a': '1', 'b': None}])))

# unchanged

table.update(update)

print(table.view().to_records())

# changed - why?

table.update([{'a': '1', 'b': None}])

print(table.view().to_records())

# >>>>>

# [{'a': '1', 'b': '2'}, {'a': '3', 'b': '4'}]

# [{'a': '1', 'b': '2'}, {'a': '3', 'b': '4'}]

# [{'a': '1', 'b': None}, {'a': '3', 'b': '4'}]
```

texodus:
This was fixed in `3.0.0` (with tests).

PR #2230, opened 2023-05-21
Add skip test: _update_with_missing_or_null_values

zxy994:
#1819 Null value updates do not propagate to perspective viewer

BEFORE:
image

AFTER:
image

### DIFF
--- a/python/perspective/perspective/tests/table/test_update.py
+++ b/python/perspective/perspective/tests/table/test_update.py
@@ -6,11 +6,37 @@
 # the Apache License 2.0.  The full license can be found in the LICENSE file.
 #
 import numpy as np
+import pyarrow as pa
+import pandas as pd
 from datetime import date, datetime
 from perspective.table import Table
+from pytest import mark
 
 
 class TestUpdate(object):
+    @mark.skip(reason="Fix for null values update yet to be implemented")
+    def test_update_with_missing_or_null_values(self):
+        tbl = Table([{"a": "1", "b": "2"}, {"a": "3", "b": "4"}], index="a")
+        tbl.update([{"a": "1", "b": None}])
+        assert tbl.view().to_records() == [{"a": "1", "b": None}, {"a": "3", "b": "4"}]
+
+        data = pd.DataFrame({"a": ["1"], "b": [None]})
+
+        arrow_table = pa.Table.from_pandas(data, preserve_index=False)
+
+        # write arrow to stream
+        stream = pa.BufferOutputStream()
+        writer = pa.RecordBatchStreamWriter(
+            stream, arrow_table.schema, use_legacy_format=False
+        )
+        writer.write_table(arrow_table)
+        writer.close()
+        arrow = stream.getvalue().to_pybytes()
+
+        tbl.update(arrow)
+        assert tbl.size() == 2
+        assert tbl.view().to_records() == [{"a": "1", "b": ""}, {"a": "3", "b": "4"}]
+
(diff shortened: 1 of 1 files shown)


python/perspective/perspective/table/table.py
```diff
@@ -303,6 +304,21 @@ def update(self, data, port_id=0):
                 data = "_" + data
 
         if _is_arrow:
+            if not _is_csv:
```
timkpaine, line 307: This block incurs ~7 extra copies of the data. Even if optimized, I don't think this is the right location or mechanism (arrow->pandas->arrow) for such a change.

texodus: I concur - though this codepath is already guilty of over-copying in at least a dozen other places.

I think this can be done similary in the data accessor code which creates [the column's null mask when reading numpy arrays](https://github.com/finos/perspective/blob/master/python/perspective/perspective/core/data/np.py#L21) without additional copying.

texodus: @zxy994 Thanks for the update but this won't work either -
* Can't add `pyarrow` as a dependency - this is a _huge_ library impacting decision not to be made casually as a side effect of a bug fix.
* Still does multiple copy+iteration, but now to `pylist` which as _far worse_.
* .. now also stacks on a conversion from bytes->Arrow->bytes again.

More importantly, now that I've actually had time to review this PR unfortunately this approach just isn't going to work. In your example, you're not actually forcing update of `null` values - you're passing out-of-range values (`nan` for floats and `string` for _everything_ else). I do not think the issue reported can be fixed in Python, pretty sure it is due to how Arrow null values are [instantiated in `arrow_loader.cpp`](https://github.com/finos/perspective/blob/master/cpp/perspective/src/cpp/arrow_loader.cpp#LL617C6-L617C6).  IIRC this method sets the wrong `STATUS_INVALID` flag for updates - this is treated as an ignore, it needs to be `STATUS_CLEAR` for _update_ operations only (which may involve passing down a flag from all methods which call this).

zxy994: Understood, thanks! @texodus

zxy994: @texodus I have updated pr accordingly. I left out the formatting which was breaking =>_Error:   1) [perspective-cli-desktop-chrome] › ../../packages/perspective-cli/test/js/superstore.spec.js:15:5 › CLI › Tests something_

I looked into your earlier correction, and it's been interesting understanding the grander scale of things. I wanted to share my findings, and incur further correction/understanding where applies.

1. Binary and string datatypes are treated differently from eg a dictionary where both data has a None value. This part of code is called for an update using binary data=> https://github.com/finos/perspective/blob/master/cpp/perspective/src/cpp/arrow_loader.cpp#L320-L336  as compared to https://github.com/finos/perspective/blob/master/cpp/perspective/src/cpp/arrow_loader.cpp#L479-L483 .
2. Based on the above, I also find that using a binary data for update => https://github.com/finos/perspective/blob/master/cpp/perspective/src/cpp/arrow_loader.cpp#L603-L604 it reports a null count of zero even with a None value in data. This makes it bypass the set_valid method =>https://github.com/finos/perspective/blob/master/cpp/perspective/src/cpp/arrow_loader.cpp#L603-L604 which is called if update is done with a dictionary.
3. Albeit, I admit I am still looking into finding the link in using STATUS_CLEAR for update operations.

texodus (approved):
Thanks for the PR @zxy994! Looks good!

PR #2238, opened 2023-05-30
Disallow non-1 multiplicity for bucketing weeks and days

sinistersnare:
I fixed the code to disallow `bucket("...", '2W')`, etc, and added a test to confirm they are disallowed right now.


texodus (approved):

You have an extra `results.tar.gz` in your commit but this isn't your fault - the logic to diff this is faulty.

sinistersnare:
I figured it would change because I added a test, but yeah idk. I can remove it from the changeset if you want.

PR #2256, opened 2023-06-13
Validate installed emsdk version

tomjakubowski:
# Feature

It's easy to forget to upgrade Emscripten when perspective updates, and your penalty if you forget is you might see weird build errors. Update run_emsdk.js to additionally check the version with `emsdk activate`

- [ ] Breaking feature
- [x] Non-breaking feature

## Changelog

- Add a new script with_emsdk.sh, call it from run_emsdk.js

## Testing

- Try changing `emscripten` version in package.json, then run: `scripts/run_emsdk.js echo hello world`


texodus (commented):
Thanks for the PR! This is a good idea. The [CI build failed](https://github.com/finos/perspective/actions/runs/5250635632/jobs/9484779163?pr=2256#step:18:24) with a curious error which needs to be fixed first.

Historically, Perspective does not use `.sh` scripts because we supported building on Windows. We don't _currently_ test this anywhere that requires `emsdk`, but maybe worth just moving this logic into JavaScript section as it is currently the solo `.sh` in the repo, since it doesn't work on CI anyway :)

texodus (approved):

One potential issue I have not tested:

1) User has old emsdk installed.
2) Upgrades Perspective repo to a version that needs a newer emsdk.
3) `activate` doesn't work because upgraded version is not installed yet.

... though the `emsdk` error on trying to activate this version should be descriptive enough.

Issue #2199, opened 2023-04-20
`color` column field does not work with `string` typed columns in `"X/Y Scatter"` chart


texodus:
### Steps to Reproduce:
Create an `X/Y Scatter` chart, with a `string` field in column position 3. As discusssed in #1487 

### Expected Result:
A chart with category-colored scatter points, ala `Treemap` and `Sunburst` charts.

### Actual Result:
A console error and no chart

### Environment:
Chrome

jaivardhan-bhola:
can i work on this ?

jaivardhan-bhola:
can I work on this issue? Also can I know what file needs to be changed?

texodus:
@jaivardhan-bhola please stop the spam, you can ask questions in [Discussion](https://github.com/finos/perspective/discussions).

jaivardhan-bhola:
I just want to work on this?

RadhaGavade01:
Solution is to check the configuration of the X/Y Scatter chart and ensure that it is properly set up to display the string field in column position 3 as a category color. It's possible that there is a bug in the chart configuration that is causing the chart to fail.

texodus:
Fixed in #2303

PR #2258, opened 2023-06-15
fix "X/Y Scatter chart color column field not working with string typed columns"

zxy994:
#2199 color column field does not work with string typed columns in "X/Y Scatter" chart

BEFORE:

https://github.com/finos/perspective/assets/38087267/0eb44dcc-ca41-40dd-96f3-b2f49bb5efe0

AFTER:

https://github.com/finos/perspective/assets/38087267/c10fa595-206e-4a85-b703-f0feeecdbc3c


texodus (changes requested):
Thanks for the PR @zxy994!  This is great.  There are a few bugs but it's quite close.

It seems to only work when a column is also specified as a `split_by` - if no `split_by` is set, it now successfully renders (which is a huge improvement!), but does _not_ render the category colors:

Screenshot 2023-06-15 at 12 48 22 PM

Another issue - it seems when category colors are calculated from distinct values rather than groups, the legend still uses group colors (wrong) and group symbols (correct). Note below `Category` is red, yellow, green;  but the actual chart colors are 4 distinct value colors for `Region`:

Screenshot 2023-06-15 at 12 49 46 PM

Some good prior art in this case is the [`Treemap` chart](https://github.com/finos/perspective/blob/master/packages/perspective-viewer-d3fc/src/js/series/treemap/treemapColor.js#L29), which has some very similar logic for dispatching on distinct column value. I suspect the `X/Y Scatter` chart need something similar.

zxy994:
Thank you @texodus for the review! This new commit aims to address the legend color mismatching.
image


For the first issue (kindly correct me where my understanding may be limited), what i found is that using group symbols for pointDataSeries has a dependency on split_by as grouping relies on the keys of the data that are not __ROW_PATH__ hence along the cause of data processing is autogenerated  :
image

image

ada-x64:
Hi @zxy994! I created a [pull request on your fork](https://github.com/zxy994/perspective/pull/1) to get some changes merged. I'm going to copy/paste the contents of that PR here for future reference.

---

Hi @zxy994, thank you for the updates!

After talking with @texodus I decided to take a look at this PR. We also realized that there are quite a few changes to how legends work that we need to make on the WASM side. In the meantime, I have some changes I'd like to merge into your branch.

First, I refactored some of the legend-determining code to be more readable.

The biggest change though is your solution to matching the chart color to the legend color. Although your solution works in the case where we have a splitBy and colorByString, I found it didn't work for some other cases. Also, data coloring usually happens before making it to pointSeriesCanvas(), so we wanted to make it consistent.

In the case where we have splitBy and colorByString, we need a key which combines the values for colorBy and splitBy. For an example, try splitting by two values. However, combining the colorBy and splitBy values requires iterating over the entire dataset, which is something we definitely don't want to do in rendering code. For now, we decided to leave the legend with a single color, and use it only to indicate the splitBy shape. This issue, and others concerning the legends, will be tackled in an issue (link forthcoming).

I created seriesColorFromField() to tackle the case where we were not splitting, but still needed to color by the unique values of the colorBy string field. This should hopefully be refactored when we move the iteration WASM-side.

@zxy994, please take a look at these changes and let me know what you think.

zxy994:
Hello @ada-x64 , checked out your updates and have it merged! Thank you for adding the clarifications. This is clearer, and builds further understanding into the issue.

texodus:
Closing - continued in #2303 

@zxy994 make sure to rebase and squash PRs please, we ask contributors not to include merge commits in PRs (see [`CONTRIBUTING.md`](https://github.com/finos/perspective/blob/master/CONTRIBUTING.md)).

Issue #1312, opened 2021-02-02
12AM parsed as 12PM in US-format date string


sc1f:
In `strptime` as compiled for Emscripten, when given the US locale string format ("%m/%d/%Y, %I:%M:%S %p"), it incorrectly parses "12:00:00 AM" as "12:00:00 PM". In order to fix this, we will need a custom-written string parser similar to the `CustomISO8601Parser` from [arrow_csv.cpp](https://github.com/finos/perspective/blob/master/cpp/perspective/src/cpp/arrow_csv.cpp#L82).  This parser uses the `arrow::TimestampParser` interface from Apache Arrow, and should be relatively easy to write and test out. A custom parser is required because the current issue manifests itself inside a call to `strptime` (especially strange, I know), and not due to any custom code.

One can reproduce the issue by adding this block of code to `make_table` inside `emscripten.cpp` (where it is guaranteed to run as part of Perspective's Table constructor):
 
```c++
        std::tm res;
        char* ret = strptime("3/1/2020, 12:30:55 AM", "%m/%d/%Y, %I:%M:%S %p", &res);
        if (ret != NULLPTR) {
            // parses to 2020/3/1 12:30:55, which is incorrect
            // should parse to 2020/3/1 00:30:55
            std::cout << "parsed using strptime: " << res.tm_year + 1900 << "/" << res.tm_mon + 1 << "/" << res.tm_mday << " " << res.tm_hour << ":" << res.tm_min << ":" << res.tm_sec << std::endl;
        }
```

Adding this same block of code in `make_table` in `python/table.cpp` generates the correct output. Tests to reproduce this issue are in #1310 - although more tests should be added to make sure no other edge cases are created.

**For users, the best way to pass in datetime values is through `new Date()`** - date strings will always be parsed as UTC.

pawan521:
Can i work on this??

texodus:
@pawan521 You don't have to ask for permission, just take a crack at it and reply in this Issue have any questions.  No task marked `good first issue` should really merit exclusivity, these are small tasks that won't suffer from potentially duplicate work - learning how Perspective works is worth the time anyway!

mahiijainn:
you can indeed create a custom string parser using the arrow::TimestampParser interface from Apache Arrow. This custom parser will ensure that the time is correctly interpreted as "00:00:00 AM" instead of "12:00:00 PM."

Here's an implementation of the custom parser based on arrow::TimestampParser:


#include <iostream>
#include <arrow/api.h>
[...]

PR #2262, opened 2023-06-20
Write custom parser for us locale time string format

adetsi:
Fix for isssue( #1312 ) where US locale string format ("%m/%d/%Y, %I:%M:%S %p") is parsed wrongly by emscripten compiler. eg.  "03/15/2020, 11:30:45 AM" is incorrectly parsed as "03/15/2020, 11:30:45 PM"

Before Implementation - Test Fails
after_time_implementation_1

After Implementations - Test Passes
After_time_implementation


adetsi:
@texodus / @sc1f could you please review my PR on issue #1312

texodus (approved):

PR #2264, opened 2023-06-22
Make PerspectiveWidget optional

tomjakubowski:
Add a try/except around the re-export at the top level

### DIFF
--- a/python/perspective/perspective/libpsp.py
+++ b/python/perspective/perspective/libpsp.py
@@ -22,7 +22,13 @@ try:
     # dependent on `libpsppy` is exposed.
     from .table import *
     from .manager import *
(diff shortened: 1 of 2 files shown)


python/perspective/perspective/libpsp.py
```diff
@@ -22,7 +22,13 @@ def is_libpsp():
     # dependent on `libpsppy` is exposed.
     from .table import *
     from .manager import *
-    from .viewer import *
+    # Only try importing viewer if traitlets is available
```
tomjakubowski, line 25: less sure of this - traitlets is a none whl so it's not that bad to depend on it

tomjakubowski:
Workarounds are good enough

PR #2267, opened 2023-06-22
Add API to give number of views into table

sinistersnare:
The test describes the usage.

Would it possibly be better so simply give a `get_views` method, that returns each View object? Then we could just call `table.get_views().length`. 

Either way, happy to do it.

### DIFF
--- a/rust/perspective-viewer/src/rust/lib.rs
+++ b/rust/perspective-viewer/src/rust/lib.rs
@@ -12,7 +12,6 @@
 #![feature(lazy_cell)]
 #![feature(macro_metavar_expr)]
 #![feature(let_chains)]
-#![feature(anonymous_lifetime_in_impl_trait)]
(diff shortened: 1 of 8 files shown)


sinistersnare:
Killing this for now.

sinistersnare:
JK 😎

timkpaine:
This should be symmetric in python

sinistersnare:
Good call, added to the Python API.

rust/perspective-viewer/src/rust/lib.rs
```diff
@@ -12,7 +12,6 @@
 #![feature(lazy_cell)]
 #![feature(macro_metavar_expr)]
 #![feature(let_chains)]
-#![feature(anonymous_lifetime_in_impl_trait)]
```
texodus, line 15: Whoops

texodus (approved):

PR #2270, opened 2023-06-23
Make `@finos/perspective` importable in TypeScript ESM

NMinhNguyen:
This should make the package importable when setting [`moduleResolution: 'bundler'`](https://devblogs.microsoft.com/typescript/announcing-typescript-5-0/#moduleresolution-bundler).

### DIFF
--- a/packages/perspective/package.json
+++ b/packages/perspective/package.json
@@ -12,6 +12,7 @@
     "jsdelivr": "dist/cdn/perspective.js",
     "exports": {
         ".": {
+            "types": "./index.d.ts",
             "node": "./dist/cjs/perspective.node.js",
             "default": "./dist/esm/perspective.inline.js"
         },


texodus:
Thanks for the PR @NMinhNguyen!

This change doesn't work - it looks like `index.d.ts` is only added to one package.

NMinhNguyen:
> Thanks for the PR @NMinhNguyen!
> 
> 
> 
> This change doesn't work - it looks like `index.d.ts` is only added to one package.

Doesn't work as in the other packages break or you want me to apply this to all the packages?

texodus:
When I run our TypeScript examples with this change (e.g. `yarn start react-example`), after updating the project to Typescript 5.1.6 and the example's `tsconfig.json` updated to use `moduleResolution: "bundler"`, I see:

```javascript
ERROR in /perspective/examples/react-example/src/index.tsx
./src/index.tsx
[tsl] ERROR in /perspective/examples/react-example/src/index.tsx(20,8)
      TS7016: Could not find a declaration file for module '@finos/perspective-viewer'. '/perspective/rust/perspective-viewer/dist/esm/perspective-viewer.inline.js' implicitly has an 'any' type.
```

I am fr not a TypeScript expert, but I presume this means all packages which use `"types"` top-level and also `"exports"` won't work with this settings. I've [forked your PR and updated this package, as well as the TypeScript example itself](https://github.com/finos/perspective/pull/2289), going to close this PR in favor of that one #2289

NMinhNguyen:
Ah I see! Thanks for taking over! Yeah in my case I’m only using `@finos/perspective`, and this change was enough, you’re just making sure this change is applied consistently to the other packages, including `@finos/perspective-viewer` :)

PR #2276, opened 2023-06-28
Remove accidental test.only, ensure it never happens again

sinistersnare:
Well, on the bright side, this will never happen again!


tools/perspective-test/playwright.config.ts
```diff
@@ -150,6 +150,7 @@ export default defineConfig({
     expect: {
         timeout: 100_000,
     },
+    forbidOnly: !!process.env.CI,
```
tomjakubowski, line 153: :chefs-kiss:

PR #2279, opened 2023-06-28
Upload `wasm32` python wheel to release

tdbgamer:
This will allow us to keep a history of all Python builds with our Github Releases. This is mostly so that we can host the WASM wheels, but while I was in there I also added the other wheels. I can remove that if we want, but I figured it wouldn't hurt to just host all of the Python artifacts on Github as well.

#### Before release
We'll need to setup a Github secret with a personal access token with write permissions for this to work:
https://github.com/softprops/action-gh-release/tree/master#permissions

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1033,6 +1033,37 @@ jobs:
                   path: python/perspective/dist/*.tar.gz
               if: ${{ matrix.os == 'macos-11' }}
 
+            ###############################
+            # Upload artifacts to release #
+            ###############################
+
+            - name: Upload dist artifact for Windows
+              if: ${{ runner.os == 'Windows' }}
+              uses: softprops/action-gh-release@v1
+              with:
+                  files: |
+                      python/perspective/dist/*.whl
+              env:
+                  GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
+            - name: Upload wheelhouse artifact for non-Windows
+              if: ${{ runner.os != 'Windows' }}
+              uses: softprops/action-gh-release@v1
+              with:
+                  files: |
+                      python/perspective/wheelhouse/*.whl
+              env:
+                  GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
+            - name: Upload sdist artifact for MacOS
+              if: ${{ matrix.os == 'macos-11' }}
+              uses: softprops/action-gh-release@v1
+              with:
+                  files: |
+                      python/perspective/dist/*.tar.gz
+              env:
+                  GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
     build_pyodide:
         runs-on: ubuntu-22.04
         needs: [initialize]
@@ -1082,6 +1113,14 @@ jobs:
               with:
                   name: perspective-python-dist-pyodide
                   path: dist/*.whl
+
+            - name: Upload pyodide wheel to release
+              uses: softprops/action-gh-release@v1
+              with:
+                  files: |
+                      dist/*.whl
+              env:
+                  GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
 
     ##########################################################################################################################
     ##########################################################################################################################
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1033,37 +1033,6 @@ jobs:
                   path: python/perspective/dist/*.tar.gz
               if: ${{ matrix.os == 'macos-11' }}
 
-            ###############################
-            # Upload artifacts to release #
-            ###############################
-
-            - name: Upload dist artifact for Windows
-              if: ${{ runner.os == 'Windows' }}
-              uses: softprops/action-gh-release@v1
-              with:
-                  files: |
-                      python/perspective/dist/*.whl
-              env:
-                  GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-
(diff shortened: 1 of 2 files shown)


TheJuanAndOnly99:
@cla-bot[bot] check

.github/workflows/build.yml
```diff
@@ -1033,6 +1033,37 @@ jobs:
...
+
+            - name: Upload sdist artifact for MacOS
+              if: ${{ matrix.os == 'macos-11' }}
+              uses: softprops/action-gh-release@v1
+              with:
+                  files: |
+                      python/perspective/dist/*.tar.gz
+              env:
+                  GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
```
texodus, line 1066: This isn't the canonical publish artifact for these (pypi), and we don't use them in any workflows. In that sense, this change  exclusively creates more work; it's a thing that can break that we don't use or have tests for, create confusion or distrust (if we break this in the future without noticing and publish different versions to pypi and GH releases).

It also can't be said to be more consistent, as we're not releasing _all_ of the artifacts, just the Python wheels.

We _could_ however use these for publishing, which would guarantee the artifacts are the same _and_ preserve the originals (since GitHub Actions expire eventually). To do this, you'd need to:

1) Update [publish.mjs](https://github.com/finos/perspective/blob/master/tools/perspective-scripts/publish.mjs) to download artifacts from a release instead of a build.
2) Upload all artifacts (including NPM artifacts) to release.

CHANGELOG.md
```diff
@@ -1,3 +1,17 @@
+# [v2.3.1](https://github.com/finos/perspective/releases/tag/v2.3.1)
```
texodus, line 1: You rebased on a reverted commit. Drop this commit please [76913ac](https://github.com/finos/perspective/pull/2279/commits/76913acc2e08e2e6fde517eed83a5dcaa0cc72e9).

texodus (changes requested):

texodus (changes requested):
This PR seems to create a release with every run - including e.g. this PR.

We only want a release to be created for tagged builds, as [per docs](https://github.com/softprops/action-gh-release#-limit-releases-to-pushes-to-tags)

tools/perspective-scripts/publish.mjs
```diff
@@ -207,6 +216,23 @@ if (proceed.toLowerCase() === "y") {
         recursive: true,
     });
 
+    const release = await octokit.request(
```
texodus, line 219: `COMMIT` flags is a safety bool to prevent this script from having side effects unless intended. When run without `COMMIT`, I expect this script to run but not actually _publish_ in the end, so I was quite surprised when I ran this for this first time and it published a GitHub release.

Since this script is now being called from CI, the release-to-GitHub functionality in this script either needs to go behind the same `COMMIT` flag (see bottom of this script), or a separate flag just for CI.

.github/workflows/build.yml
```diff
@@ -1574,3 +1574,39 @@ jobs:
...
+              env:
+                  PSP_SKIP_EMSDK_INSTALL: 1
+
+            - name: Run publish script
+              run: node tools/perspective-scripts/publish.mjs
+              env:
+                  CI: true
+                  GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+                  GITHUB_WORKFLOW_ID: ${{ github.run_id }}
+                  COMMIT: ${{ github.sha }}
```
texodus, line 1612: This isn't the commit sha, it's a flag to tell `publish.mjs` whether to run in dry-run mode. Setting this to SHA causes this script (in CI) to attempt to publish to pypi & npm, which it isn't authenticated for (and we don't want to do this for now also).

texodus (commented):
This is the comment from the previous review:

> 1. Update [publish.mjs](https://github.com/finos/perspective/blob/master/tools/perspective-scripts/publish.mjs) to download artifacts from a release instead of a build.
> 2. Upload all artifacts (including NPM artifacts) to release.

This is the equivalent descriptor for what this PR does:

> 1. Update `publish.mjs` to upload artifacts to a github release and also pypi+npm.
> 2. Call `publish.mjs` from the Workflow `build.yml` in CI

This isn't quite what I wanted from this change, but I'm in a hurry to get the wasm32 disted so I let's try it. However there are also some blockers still:

* There are a few hard errors in this PR (looks like rebase errors).
* I'm quite skeptical, when I do run this on CI, that this task will be able to run `publish.mjs` successfully anyway, as I'm not sure the `octokit` workflow-downloading API can be used form within the workflow it is attempting to download from ... I suspect this needs to use the workflow's `actions/download-artifact` step instead?
* We can't publish to NPM or Pypi from CI yet.
* For the other wheel sit doesn't matter, but for the pyodide wheel a `.zip` doesn't help us - we need the unzipped `.whl` disted directly for it to work as a web host for this artifact form pyodide.

In the interest of getting this working quickly, I'm going to go ahead and close this PR and pick up these changes in #2294 which has been rebased and implemented a few fixes necessary to start testing

tools/perspective-scripts/publish.mjs
```diff
@@ -221,11 +247,25 @@ if (proceed.toLowerCase() === "y") {
...
+
+            if (!artifact.name.includes("pyodide")) {
+                // Write out the zip file
+                await fs.appendFile(
+                    `${dist_folder}/${artifact.name}.zip`,
+                    Buffer.from(download.data)
+                );
+            }
         })
     );
```
texodus, line 271: Need the same checks for `pyodide` everywhere `python_dist_folders` is iterated over

PR #2281, opened 2023-06-29
Add `get_hosted_table_names` to js and python server apis.

brochington:
This PR adds the `get_hosted_table_names` methods to server api so that table names can now be fetched via clients.


python/perspective/perspective/manager/manager_internal.py
```diff
@@ -158,6 +161,8 @@ def __process(self, msg, post_callback, client_id=None):
             elif cmd == "table_method" or cmd == "view_method":
                 # Call the method on the table/view instance
                 self._process_method_call(msg, post_callback, client_id)
+            else:
+                logging.error("Unknown client message " + str(msg))
```
texodus, line 165: This looks like leftover debug code, to get to this block you'd need a malformed network message or a typo in the server code, as these conditional structurally decompose `msg`.

texodus (approved):

PR #2282, opened 2023-06-29
Fix bug preventing blocks example server from starting.

brochington:
This PR fixes a bug where the blocks server could not be started with the command `yarn start blocks`.


texodus (approved):

Good catch - this was a blockhead mistake by me, suffered because it is common in Perspective development to just run this static file server in the background in effective perpetuity.

PR #2284, opened 2023-06-30
Tweak rounding display logic for float filters

timkpaine:
This is a somewhat silly implementation, but when you try to input floating point numbers for a filter, they will automatically be rounded and reformatted. This is a problem when you want to filter for decimals with leading `0`s, e.g. you won't be able to type in `1.0001`:

tmp-2

This PR "fixes" that by cheating, and treating the user-inputted `float` as a `string`. This works because the `input` element won't let you input characters, and the actual use of the field occurs in C++ which will do its own conversion logic (or might just naively assume it to already be a `float`):

tmp

This is a pretty bad implementation, but I didn't see an obvious better implementation since the float formatting code in rust is enough layers removed from the `input` component that it doesn't make a ton of sense to bridge the gap in a more *elegant* way.

### DIFF
--- a/rust/perspective-viewer/src/rust/components/column_selector/filter_column.rs
+++ b/rust/perspective-viewer/src/rust/components/column_selector/filter_column.rs
@@ -193,7 +193,13 @@ impl FilterColumnProps {
                     if val.is_empty() {
                         None
                     } else if let Ok(num) = val.parse::<f64>() {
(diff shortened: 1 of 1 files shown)


texodus:
Thanks for the PR!

I've fixed this _without_ cheating and added a test in #2353, closing in favor of that PR

PR #2285, opened 2023-07-01
Fix build regression in python dependency

timkpaine:
https://github.com/pyodide/pyodide/pull/3971
https://github.com/numpy/numpy/pull/24091


texodus (approved):

Must be a day that ends in the letter `-y`, because the Python build has spontaneously broken again!

PR #2303, opened 2023-07-10
Fix XY-Scatter Plots Colors - Rebased

ada-x64:
This PR continues the work in #2258. It needed to be rebased and tweaked. @zxy994 your commits are preserved.

### DIFF
--- a/packages/perspective-viewer-d3fc/src/js/charts/xy-scatter.js
+++ b/packages/perspective-viewer-d3fc/src/js/charts/xy-scatter.js
@@ -11,6 +11,7 @@
 // ┗━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┛
 
 import * as fc from "d3fc";
+import * as d3 from "d3";
 import { axisFactory } from "../axis/axisFactory";
 import { chartCanvasFactory } from "../axis/chartFactory";
 import {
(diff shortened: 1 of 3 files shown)


texodus:
Thanks @zxy994 @ada-x64 

For charts with _only_ a `split_by`, the chart no longer colors scatter points individually (as discussed) - but it should use the default color theme color `var(--d3fc-series)` (which is a pale blue in the Pro theme), instead it seems to be picking the first distinct color from the distinct color series calculation:

Screenshot 2023-07-10 at 11 52 53 PM

ada-x64:
@texodus Amended to use default colors. Also tweaked some code to be more readable.

finos-admin:
@cla-bot[bot] check

texodus (approved):
Thanks for the PR @zxy994 @ada-x64! Looks good!

I think there are still a few areas this should be improved on in the future:
* We definitely need to improve the legend situation when both `split_by` and `color` columns are specified - right now it only shows them symbol legend (and I'd argue if we're to choose one, it should be the color legend).
* Mapping `split_by` to symbols isn't great to begin with - we should add a new column for `symbol` and make `split_by` a multi-chart ala "Treemap" and "Sunburst" charts.
* It's disorienting when switching from `split_by` to non-`split_by` configs with a color column, as the color palette change seeming arbitrarily (because the columns are named differently). 
* In the future, we really want to see **tests for all commits**. As this PR has gone through 4 rounds of code review, we'll hold of on these but please aim to bring test coverage with all PRs.

PR #2309, opened 2023-07-16
Fix build.rs escaping OUT_DIR

tdbgamer:
According to [The Cargo Book](https://doc.rust-lang.org/cargo/reference/build-scripts.html#outputs-of-the-build-script) build scripts aren't supposed to modify files outside `OUT_DIR`. I modified the `build.rs` to follow that rule and made a separate `css` macro that doesn't point at the target directory.

The reason I made a new macro is that it's technically a breaking change to do this because any projects depending on this macro will break since it'll point to the `OUT_DIR` of _their_ crate which may note even have a `build.rs` at all. But also the previous version of this would've also pointed to the user's `CARGO_MANIFEST_DIR` which seems maybe not the most useful?

We could either break this and force users to use the `OUT_DIR` (which maybe they should already be doing if they have a build script generating CSS) or keep this PR as is.

### DIFF
--- a/rust/perspective-viewer/build.rs
+++ b/rust/perspective-viewer/build.rs
@@ -44,13 +44,16 @@ fn glob_with_wd(indir: &str, input: &str) -> Vec<String> {
 }
 
 fn main() -> Result<(), anyhow::Error> {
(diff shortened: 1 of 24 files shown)


texodus (approved):

PR #2310, opened 2023-07-17
Publish benchmarks

texodus:
Adds benchmark Arrow files to the release artifacts.


timkpaine:
I've got 0-10 ascii art if you want whole numbers 😂
https://github.com/python-project-templates/cpp/blob/6c7ac26fedabb3bb1e8b8ea716e743f25dfef41d/.github/workflows/build.yml#L442

PR #2314, opened 2023-07-21
Move `to_columns()` JSON creation logic to C++.

brochington:
This PR adds creation of a JSON data string from within the core C++ engine. Outputting an already formed JSON string is faster than doing the same in Javascript, while also helping to save a (couple?) (de)serialization round trips.


texodus:
Closing in favor of #2315

PR #2315, opened 2023-07-25
Add `to_columns_string()` C++ JSON API

texodus:
This PR adds `View.to_columns_string()` method, which works similarly to `to_columns()`, except it returns a JSON string instead of a JavaScript object (in JS) or Python Dict (in Python). `to_columns_string()` is implemented in C++ useing RapidJSON library, and re-used verbatim in both JavaScript and Python, reducing duplicate code between these two projects.

In JavaScript:
* In JavaScript, `to_columns()`, `to_json()` have been rewritten in terms of `to_columns_string()`, using `JSON.parse()` 
* As JavaScript is typically instantiated in a WebWorker, using `to_columns_string()` saves one Object->String->Object conversion (as the data can stay in string form until it is copied into the main renderer process).
* Linear performance is improved ~2x for these methods.
* `@finos/perspective-viewer-datagrid` and `@finos/perspective-viewer-d3fc` have been rewritten to use this method internally, leading to improved performance for both.

In Python
* `to_columns()`, `to_records()` have been rewritten in terms of `to_columns_string()`, using `json.loads()`
* All 3 of these methods are now GIL-less, so concurrency for virtual workloads is much improved.
* (Breaking) `Inf`, `-Inf` and `nan` support has been removed from these methods. These data structures _errored_ when serializing anyway, so were entirely useless for visualization purposes. `to_numpy()` and `to_arrow()` preserve these values still.

In both:
* `date` and `datetime` columns now use `YYYY-MM-DD` and `YYYY-MM-DD HH:MM:SS.SSSS` formatting respectively, for CSV export and `split_by` headers. The former can only be fixed at the expense of performance and/or significant asset size (bundling a locale collection); the latter can be fixed with a refactoring of the JSON representation.

Serial (linear) performance by version:
Screenshot 2023-07-26 at 11 58 01 AM

10 clients concurrent performance time series:
Screenshot 2023-07-28 at 10 12 02 PM


timkpaine:
just noting that you've auto formatted the python tests here. I'm a fan of this, but it does make for a big changeset

texodus:
@timkpaine The auto formatting is exclusively in [a single commit authored by `no author`](https://github.com/finos/perspective/pull/2315/commits/abfaa2729497b4a8251a712d7ffee1e630bd1e1b).

This code shouldn't have been excluded from formatting. The issue (currently, though the decision begets issues) is that VSCode does not respect the `pyproject.toml` configuration for `black` suddenly, resulting in these files being obliterated whenever they are edited. This PR modifies python tests, btu the modifications are not in the auto-format commit. We could move this commit to a separate PR but as it impacts nothing IMO don't think its worth the CI time.

packages/perspective/test/js/leaks.spec.js
```diff
@@ -111,6 +122,19 @@ test.describe("leaks", function () {
             view.delete();
             table.delete();
         });
+
+        test.skip("csv loading does not leak", async () => {
```
brochington, line 126: Just double checking, should this be skipped?

packages/perspective/test/js/pivots.spec.js
```diff
@@ -2355,25 +2355,21 @@ const std = (nums) => {
             table.delete();
         });
 
-        test("['z'] only, datetime column", async function () {
+        test("['z'] only, datetime column", async function ({ page }) {
```
brochington, line 2358: This `page` can be removed, I believe, as it was used for debugging. I can push an update.

packages/perspective/test/js/to_format.spec.js
```diff
@@ -431,8 +431,8 @@ const pivoted_output = [
             let view = await table.view();
             let json = await view.to_json({ formatted: true });
             expect(json).toEqual([
-                { datetime: "6/13/16" },
-                { datetime: "6/14/16" },
+                { datetime: "2016-06-13" },
+                { datetime: "2016-06-14" },
```
brochington, line 435: Curious, how important is it to maintain datatime strring parsing across different systems? Since it's possible to use a parsed date string as part of a column name, should we at one point add a C++ based function that both JS and Python can call to get a standardized formatted date string?

python/perspective/bench/runtime/perspective_benchmark.py
```diff
@@ -184,30 +184,38 @@ def benchmark_to_format_zero(self):
...
             "arrow",
             "csv",
+            "columns",
+            "records",
         ):
-            test_meta = make_meta("to_format", "to_{}".format(name))
-            func = Benchmark(lambda: getattr(self._view, "to_{0}".format(name))(), meta=test_meta)
+            method = "to_{0}".format(name)
+            test_meta = make_meta("to_format", method)
+            func = Benchmark(getattr(self._view, method), meta=test_meta)
```
brochington, line 192: If benchmarking results in static assets, maybe we can create an blocks example that pulls in the data.

packages/perspective/test/js/leaks.spec.js
texodus: It takes forever to run is the only reason to skip it

packages/perspective/test/js/to_format.spec.js
texodus: There are only two conditions where `date`/`datetime` columns need to be formatted:

(1) When they are used in a `split_by` and we need to stringify them to create compound column names. This is really a deficiency of the JSON encoding we use which doesn't permit true column paths. Fixing this encoding would remove the need for formatting for this case.

(2) When the `formatted` flag is passed. This is only used by the CSV exporter (as users typically expect CSVs to be "human readable"). Formerly, this used `sprintf` to try to replicate the `en/us` locale, as the browser native locale is not cheaply accessible from C++; however, this was never right, especially in non-`en/us` locales. The format in this patch is a compromise - it is not locale-formatted, but it is consistently emitted and parsed (though assumption are made about the encoding timezone in this case which are symmetric wrt Perspective but maybe not always desired).

Issue #1237, opened 2020-10-27
Add support for StringDtype (available in Pandas >=1.0)


pgagarinov:
## Feature Request
Allow for Pandas dataframe columns of "string" datatype - see https://pandas.pydata.org/pandas-docs/stable/user_guide/text.html

### Description of Problem:
Right now the following call of perspective's Table constructor fails:
```
from perspective import Table
import pandas as pd
df = pd.DataFrame({'a': ['aa','bbb'],'b':['dddd','dd']},dtype='string')
Table(df)
```
with
`PerspectiveError: Mixed datasets of numpy.ndarray and lists are not supported.`

### Potential Solutions:
Maybe just convert (internally) to dtype('o') for backward compatibility as the first step... At least `Mixed datasets of numpy.ndarray and lists are not supported.` needs to be replaced with something more meaningful as StringDType is neither list nor numpy array. It does also look that StringDType can be converted to arrow array, no problem (see https://github.com/pandas-dev/pandas/blob/v1.1.3/pandas/core/arrays/string_.py#L26-L97)

timkpaine:
We don't currently support the `string` datatype. We'll never support pure transparent pandas compatibility, so best bet is to convert to stable apis (like object) prior to trying to ingest.

texodus:
What is "pure transparent pandas compatibility"?  Support for `dtype="string"` seems simple to add, as does support for newer Pandas versions `>1`.

timkpaine:
@texodus as an example, pivot deconstruction and reconstruction from pandas is mostly broken right now, and there are lots of scenarios where you can't just go from pandas into perspective and get the same results, e.g. in a pivot table you might not have the ability to unpick and repivot.

Also we should be careful to avoid features explicitly marked "unstable" whether or not the exist in a newer version.

timkpaine:
From the docs "StringDtype is considered experimental. The implementation and parts of the API may change without warning."

PR #2316, opened 2023-07-28
1237: Add support for StringDtype

MaDufie:
#1237 Add support for StringDtype (available in Pandas >=1.0)

AFTER
perspective


python/perspective/perspective/table/table.py
```diff
@@ -71,6 +71,9 @@ def __init__(self, data, limit=None, index=None):
         if self._is_arrow:
             _accessor = data
         else:
+            if isinstance(data, pandas.DataFrame):
+                if data[data.columns[0]].dtypes == "string":
```
timkpaine, line 75: this checks if the first column is a string type, and if so converts all columns to object

timkpaine (changes requested):

finos-admin:
@cla-bot[bot] check

PR #2319, opened 2023-07-30
Add support for StringDtype (fixes #1237)

### LINKED ISSUE #1237: Add support for StringDtype (available in Pandas >=1.0)

MaDufie:
Add support for StringDtype (available in Pandas >=1.0)
Fixes #1237 

AFTER
image

### DIFF
--- a/python/perspective/perspective/core/data/pd.py
+++ b/python/perspective/perspective/core/data/pd.py
@@ -78,6 +78,12 @@ def deconstruct_pandas(data, kwargs=None):
                 if isinstance(v, pd.CategoricalDtype):
                     data[k] = data[k].astype(str)
 
+    # convert StringDtype to str
+    if isinstance(data, pd.DataFrame) and hasattr(pd, "CategoricalDtype"):
+        for k, v in data.dtypes.items():
+            if isinstance(v, pd.StringDtype):
+                data[k] = data[k].astype(str)
+
     if isinstance(data, pd.DataFrame) and isinstance(data.columns, pd.MultiIndex) and isinstance(data.index, pd.MultiIndex):
         # Row and col pivots
         kwargs["group_by"].extend([str(c) for c in data.index.names])
--- a/python/perspective/perspective/core/data/pd.py
+++ b/python/perspective/perspective/core/data/pd.py
@@ -79,7 +79,7 @@ def deconstruct_pandas(data, kwargs=None):
                     data[k] = data[k].astype(str)
 
     # convert StringDtype to str
-    if isinstance(data, pd.DataFrame) and hasattr(pd, "CategoricalDtype"):
+    if isinstance(data, pd.DataFrame) and hasattr(pd, "StringDtype"):
         for k, v in data.dtypes.items():
             if isinstance(v, pd.StringDtype):
                 data[k] = data[k].astype(str)
--- a/python/perspective/perspective/tests/table/test_table_pandas.py
+++ b/python/perspective/perspective/tests/table/test_table_pandas.py
@@ -597,3 +597,15 @@ class TestTablePandas(object):
         df_both = pd.DataFrame(np.random.randn(3, 16), index=["A", "B", "C"], columns=index)
         table = Table(df_both)
         assert table.size() == 48
+
+    def test_table_dataframe_for_dtype_equals_string(self):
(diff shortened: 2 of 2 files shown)


python/perspective/perspective/table/table.py
```diff
@@ -71,6 +71,10 @@ def __init__(self, data, limit=None, index=None):
         if self._is_arrow:
             _accessor = data
         else:
+            if isinstance(data, pandas.DataFrame):
+                for column in data.columns:
+                    if data[column].dtypes == "string":
+                        data = data.astype("str")
```
timkpaine, line 77: this converts the entire dataframe to “str” if any of the columns are a string

timkpaine (changes requested):

timkpaine:
You should try to implement this the same way as done here: https://github.com/finos/perspective/blob/37e3df35f0b6cc935137f669854b88b12641d3b4/python/perspective/perspective/core/data/pd.py#L75

MaDufie:
> You should try to implement this the same way as done here:
> 
> https://github.com/finos/perspective/blob/37e3df35f0b6cc935137f669854b88b12641d3b4/python/perspective/perspective/core/data/pd.py#L75
> 
> And please use the same PR as you make changes, rather than opening a new one.

@timkpaine I have implemented the changes. Kindly let me if there are any more changes after you review them. Thank you.

timkpaine:
It should be in the file linked, near the line linked.

python/perspective/perspective/table/table.py
```diff
@@ -9,7 +9,7 @@
 #  ┃ This file is part of the Perspective library, distributed under the terms ┃
 #  ┃ of the [Apache License 2.0](https://www.apache.org/licenses/LICENSE-2.0). ┃
 #  ┗━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┛
-
+import pandas as pd
```
timkpaine, line 12: Remove unneeded import
```suggestion

```

timkpaine (changes requested):

timkpaine:
@texodus this lgtm, I know we're looking to replace all of pandas / numpy with native arrow in/out so I leave up to you whether or not you want to merge.

python/perspective/perspective/core/data/pd.py
```diff
@@ -78,6 +78,12 @@ def deconstruct_pandas(data, kwargs=None):
                 if isinstance(v, pd.CategoricalDtype):
                     data[k] = data[k].astype(str)
 
+    # convert StringDtype to str
+    if isinstance(data, pd.DataFrame) and hasattr(pd, "CategoricalDtype"):
```
timkpaine, line 82: is this `hasattr` necessary?

python/perspective/perspective/core/data/pd.py
```diff
@@ -78,6 +78,12 @@ def deconstruct_pandas(data, kwargs=None):
...
+        for k, v in data.dtypes.items():
+            if isinstance(v, pd.StringDtype):
```
timkpaine, line 84: I think maybe this was meant to be used in the `hasattr`, `hasattr(pd, "StringDType")` since `StringDType` is the new type

texodus (approved):
Thanks for the PR @MaDufie! Looks good!

We should really be taking advantage of the internal dictionary for Categories, but we don't really provide any fast path loading for DataFrame yet and likely won't as we move to Arrow.

PR #2325, opened 2023-08-03
Port perspective-jupyterlab to TypeScript

tomjakubowski:
Partly undoes 2d6e400c2afb25e7c7790b118cb9b988e6afafd5

This change includes a restriction on the ipywidgets versions perspective declares that it supports.  The plugin probably didn't work with ipywidgets <= 7 anyhow, since it calls `processLuminoMessage`, which in those earlier versions of ipywidgets was named `processPhosphorMessage`.

see https://ipywidgets.readthedocs.io/en/8.0.7/migration_guides.html#phosphor-lumino

and

https://ipywidgets.readthedocs.io/en/8.0.7/migration_guides.html#updating-package-json

if we want to support older ipywidgets, we should do something like ipywidgets docs describe


packages/perspective-jupyterlab/src/ts/view.ts
```diff
@@ -37,11 +52,32 @@ function isEqual(a, b) {
...
+type ProxyTable = proxy_table & Table;
+
 /**
  * `PerspectiveView` defines the plugin's DOM and how the plugin interacts with
  * the DOM.
  */
 export class PerspectiveView extends DOMWidgetView {
+    luminoWidget: PerspectiveJupyterWidget;
+    perspective_client: PerspectiveJupyterClient;
+    _pending_binary?: unknown;
```
tomjakubowski, line 66: these values are only ever assigned, and not read, so I think it's safe to leave them unknown

packages/perspective-jupyterlab/src/ts/view.ts
```diff
@@ -300,7 +338,7 @@ export class PerspectiveView extends DOMWidgetView {
                 const table = this.perspective_client.open_table(
                     msg.data["table_name"]
                 );
-                this.luminoWidget.load(table);
+                this.luminoWidget.load(table as unknown as Promise<Table>);
```
tomjakubowski, line 341: this use of unknown is questionable, let me add a comment

packages/perspective-jupyterlab/tsconfig.json
```diff
@@ -0,0 +1,23 @@
...
+        "types": ["node"],
+        "paths": {
+            "@finos/perspective/api/*": ["../perspective/src/js/api/*"]
+        }
+    },
+    "include": [
+        "package.json",
+        "src/js/**/*",
+        "src/ts/**/*",
+        "../perspective/src/js/api/*"
```
tomjakubowski, line 21: this is a little nasty

packages/perspective-jupyterlab/package.json
```diff
@@ -39,7 +39,7 @@
         "@finos/perspective-viewer-d3fc": "^2.3.2",
         "@finos/perspective-viewer-datagrid": "^2.3.2",
         "@finos/perspective-viewer-openlayers": "^2.3.2",
-        "@jupyter-widgets/base": "^2 || ^3 || ^4 || ^5 || ^6",
+        "@jupyter-widgets/base": "^6",
```
timkpaine, line 42: https://ipywidgets.readthedocs.io/en/latest/migration_guides.html#updating-package-json

"Note that “@jupyter-widgets/base” version 5 is reserved for ipywidgets 7 support on JupyterLab 4, “@jupyter-widgets/base” version 6 is the version released with ipywidgets 8."

tomjakubowski: Yeah I think we want ipywidgets 8 (and base version 6). The extension's `PerspectiveJupyterWidget` class calls `this._view.processLuminoMessage`, a method which, according to that document, is not present in ipywidgets 7, where it was called `processPhosphorMessage`. 

https://ipywidgets.readthedocs.io/en/latest/migration_guides.html#phosphor-lumino

If we do currently support ipywidgets I'd like to know how, because I haven't seen any stub code in perspective to support `processPhosphorMessage`

packages/perspective-jupyterlab/src/ts/view.ts
```diff
@@ -12,11 +12,26 @@
...
+} from "@finos/perspective";
+
+import "@finos/perspective-viewer";
+import { proxy_table } from "@finos/perspective/api/table_api";
+
+// Type of first parameter to @finos/perspective UpdateCallback
+export type TableUpdate = {
+    port_id: number;
+    delta:
+        | Array<Record<string, Array<string | boolean | Date | number>>>
```
tomjakubowski, line 30: I need to take another pass at these types, based on standup discussion they were right before

packages/perspective-jupyterlab/package.json
timkpaine: We currently support ipywidgets 7

tomjakubowski: No, we don't. We say we support ipywidgets 7 but it's a lie. 

> The extension's PerspectiveJupyterWidget class calls this._view.processLuminoMessage, a method which, according to that document, is not present in ipywidgets 7, where it was called processPhosphorMessage.

tomjakubowski: if we do want to support ipywidgets 7 we'll need something like this. Not a big deal to add it but right now today HEAD of master won't work with ipywidgets 7.

https://github.com/jupyter-widgets/ipyleaflet/pull/969

tomjakubowski: If someone is successfully using perspective in a notebook with ipywidgets 7 I would be very interested in learning how they've done that.

tomjakubowski: I believe this commit broke support for ipywidgets 7 https://github.com/finos/perspective/commit/715cd7ad2a77cc6a8a7cf39e85ab618e6163e288

packages/perspective-jupyterlab/src/ts/renderer.ts
```diff
@@ -128,8 +151,20 @@ export class PerspectiveDocumentWidget extends DocumentWidget {
                         this.context.model.fromString(resultAsB64);
                         this.context.save();
                     } else if (this._type === "json") {
+                        // TODO: Gotta fix this - need to do something about Date values
+
                         const result = await view.to_json();
-                        this.context.model.fromJSON(result);
+                        for (let [_key, val] of Object.entries(result)) {
+                            if (val instanceof Date) {
+                                throw new Error("TODO: Dates unsupported");
```
tomjakubowski, line 159: this needs to be addressed. As of right now I don't recall why I added this loop to begin with - some cockamamie scheme in anger to satisfy the type checker? (this loop would be narrowing some of the types involved)

tomjakubowski:
TODOs:

- [ ] Verify that the type variants removed from TableUpdate truly aren't supported by perspective. Simple way would be to try constructing a table from an `Array<Record<string, Array<number>>>` (e.g. `[{"a": [1,2,3]}, {"a": [4,5,6]}]`), then try passing a value like that to update/replace, etc.
- [ ] Understand what's going on with Dates and remove the loop I added here https://github.com/finos/perspective/pull/2325#pullrequestreview-1588267923

packages/perspective-jupyterlab/package.json
tomjakubowski: https://github.com/finos/perspective/pull/2325/files#diff-6653218916957fa967fc60ecfce389d2126af0e69bac83f49a29a2a38a6ea8daR59

packages/perspective-jupyterlab/src/ts/renderer.ts
texodus: Perspective outputs `number` values in posix epoch time for columns of type `date` and `datetime`, ergo I think the fix for this is to remove `Date` from the [types of these methods in `@finos/perspective`](https://github.com/finos/perspective/blob/master/packages/perspective/index.d.ts#L106).

packages/perspective-jupyterlab/src/ts/view.ts
texodus: Looks like `delta` for this type actually no longer supports `Array<...>` either, it is exclusively Arrow eg `ArrayBuffer`

packages/perspective-jupyterlab/src/ts/renderer.ts
```diff
@@ -47,15 +53,26 @@ const baddialog = () => {
 
 const WORKER = perspective.worker();
 export class PerspectiveDocumentWidget extends DocumentWidget {
-    constructor(options, type = "csv") {
+    _monitor: ActivityMonitor<DocumentRegistry.IModel, void>;
+    _psp: PerspectiveWidget & {
+        theme?: string;
+    };
+    _type: string;
```
timkpaine, line 60: type is one of "csv", "arrow", or "json" so can be constrained to string literal orring.

timkpaine:
I would like some description of motivation if possible. One major motivation not to go back to typescript which won't be seen in this PR is the difficulty it presents when upgrading to new versions of Jupyter. Every time seems to present new and unique annoyances, with minor variances of the typescript compiler between what we use and what jupyterlab uses creating incomprehensible error messages. Given that the jupyter plugin is explicitly designed to NOT be used by anyone other than us, I don't think e.g. IDE convenience is a big motivation either. The jupyter plugin is not very large and not very actively developed, so there's not much devex benefit to be had either imo.

PR #2332, opened 2023-08-09
Remove async timeout and async pytest wrapper from flapping tornado tests

timkpaine:
Full Build: https://github.com/finos/perspective/actions/runs/5814011070


texodus:
Thanks for the PR!

I think we want the timeout, its the explicit failure condition of this test. This is def a concurrency bug in the functionality under test which we'll need to fix.

PR #2337, opened 2023-08-10
Perspective reading LZ4 compressed RecordBatches.

sinistersnare:
Now you can drop a file created by 

```py
import pandas as pd
import pyarrow.feather as feather
data = [['a', 10], ['b', 15], ['c', 14]]
df = pd.DataFrame(data, columns=['Name', 'Age'])
# This defaults to compressing with LZ4.
feather.write_feather(df, 'lz4_compressed_data.feather')
```

PR #2339, opened 2023-08-14
LZ4 compression for Apache Arrow 

texodus:
Adds LZ4 compression to Perspective's Apache Arrow read/write support, which also makes Perspective compatible with e.g. `pyarrow.feather.write_feather()` IPC methods which default to this compression codec. This PR addresses issues from #2337, adding:
* Upgrades Apache Arrow to `v12.0.0`
* A new option/kwarg for `to_arrow()`, `compression`, which can be `"lz4"` or `null`/`None` (for JavaScript/Python, respectively). This option defaults to `"lz4"`.
* Tests.
* Benchmarks.  There is a slight impact to LZ4 even with the relatively small datasets we benchmark with.

This PR is a breaking change, as the `compression` option causes default `.to_arrow()` output to be incompatible with older Perspective versions. To backport, set this option to `null`/`None`:

JavaScript:
```javascript
const arrow = view.to_arrow({compression: null});
```

Python:
```python
arrow = view.to_arrow(compression=None)
```

[full build](https://github.com/finos/perspective/actions/runs/5849495493)


cpp/perspective/CMakeLists.txt
```diff
@@ -3,7 +3,7 @@ project(psp)
 include(CheckCCompilerFlag)
 
 set(CMAKE_BUILD_TYPE "Release")
-set(CMAKE_CXX_STANDARD 14)
+set(CMAKE_CXX_STANDARD 17)
```
timkpaine, line 6: should update/remove gnu flag to match: https://github.com/finos/perspective/blob/88e08788340652e614c4755b7d6830ee64869ddf/cpp/perspective/CMakeLists.txt#L414

cpp/perspective/CMakeLists.txt
```diff
@@ -509,7 +509,7 @@ set(PYTHON_SOURCE_FILES ${SOURCE_FILES}
 )
 
 set(WASM_SOURCE_FILES ${SOURCE_FILES}
```
timkpaine, line 511: should remove whole block if not used anymore

cpp/perspective/CMakeLists.txt
texodus: Where is this tested?

cpp/perspective/CMakeLists.txt
texodus: [used here](https://github.com/finos/perspective/pull/2339/files#diff-50e5b576dee1bef31c114f1bef14d219054509aa2da37937228763998409692dR556)

Issue #1256, opened 2020-12-03
Handling null updates in sum aggregate


nmichaud:
### Steps to Reproduce:
 1. Create a table with an int/float column
 2. Tick updates to the table where the value in that column switches between null and a number
 3. Do a pivot with a sum on that column

### Expected Result:
Sum value should flip back and forth

### Actual Result:
Sum value constantly increases

Here's an example:
```
<!doctype html>
<html>
<head>
    <meta name="viewport" content="width=device-width, initial-scale=1, maximum-scale=1, minimum-scale=1, user-scalable=no">

    <script src="https://cdn.jsdelivr.net/npm/@finos/perspective/dist/umd/perspective.js"></script>
    <script src="https://cdn.jsdelivr.net/npm/@finos/perspective-viewer/dist/umd/perspective-viewer.js"></script>
    <script src="https://cdn.jsdelivr.net/npm/@finos/perspective-viewer-datagrid"></script>
    <script src="https://cdn.jsdelivr.net/npm/@finos/perspective-viewer-d3fc"></script>

    <link rel='stylesheet' href="https://cdn.jsdelivr.net/npm/@finos/perspective-viewer/dist/umd/material-dense.css" is="custom-style">

    <style>
      perspective-viewer { position: absolute; top: 0; left: 0; right: 0; bottom: 0; }
    </style>
</head>
<body>
  <perspective-viewer row-pivots='["ticker"]' columns='["pnl"]'></perspective-viewer>
  <script>
  window.addEventListener("WebComponentsReady", function() {   
    // Create a new Perspective WebWorker instance.
    var worker = perspective.worker();

    let table = worker.table({"ticker": "string", "pnl": "integer"}, { index: "ticker"});

    // Load the `table` in the `<perspective-viewer>` DOM reference.
    document.getElementsByTagName("perspective-viewer")[0].load(table);

    table.update([{"ticker": "IBM", "pnl": 100},{"ticker": "AAPL", "pnl": 100}]);

    let flip = true;
    let insert = () => {
      table.update([{"ticker": "AAPL", "pnl": flip ? null : 100}]);
      flip = !flip;
    }

    setInterval(insert, 500);

  });
  </script>
</body>
</html>
```

PR #2341, opened 2023-08-15
Handling null updates in sum test cases

adetsi:
Upon investigation, this  issue (#1256)  is not really an issue, it is as a result of the way perspective view is interacted with, creating new instances of perspective view upon table update flips the aggregate value back and forth, however, not creating new instances of perspective view upon table update increments the sum aggregate (issue complained about by @nmichaud ). @texodus @timkpaine @sc1f could you please review if all is good?

### Sum value flips back and forth when a new view is created upon table update

https://github.com/finos/perspective/assets/131965000/b0ee5b5b-e94b-47e2-9025-41894700f3e6

### Sum value increments when no new view is created upon table update

https://github.com/finos/perspective/assets/131965000/19173b9c-37bd-4afb-8c2d-39d7e21bad0a

### DIFF
--- a/packages/perspective/test/js/pivots.spec.js
+++ b/packages/perspective/test/js/pivots.spec.js
@@ -3082,5 +3082,124 @@ const std = (nums) => {
             view.delete();
             table.delete();
         });
+
+        test("Sum value flips back and forth when a new view is created upon table update", async function () {
+
+            var answer = [
+                [
+                    { __ROW_PATH__: [], ticker: 2, pnl: 200 },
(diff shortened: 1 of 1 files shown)


texodus (commented):
Thanks for the PR!

#1256 is definitely a bug. Your second test's assertion is the _wrong_ behavior, both of these tests should behave identically regardless of when `View` the is created. Perspective does not have a `sum()` function which behaves like this (though `high` and `low` aggregates _do_ accumulate perpetually in `View` state).

Please do lint your fixes before opening a PR in the future, our CI will immediately fail it anyway so no point committing with `--no-verify`.

PR #2352, opened 2023-08-30
Fix Jupyter plugin resize behavior

texodus:
Fixes `perspective-python`'s Jupyter plugin resize behavior to be vertical-only in notebook/cell mode, and non-resizable in other modes. Previous versions of Perspective allowed this plugin to be resized in both dimension, but this resizability overrides responsiveness to the notebook (or browser) dimension changes, leading to awkward user interactions when one wants to change the width of the notebook after using Perspective.

This PR may break Voila, as per the note in the CSS I removed. Instead, I made this removed behavior "default" and made `<perspective-viewer>` within a Notebook the selector for `height`, which should work? But we have no tests nor convenient way to verify this. Additionally, I maintain the Voila fix was not correct - if we're going to distinguish behavior that is specific to Voila, it can't share classnames with Jupyter.


timkpaine:
Voila moved to JupyterLab proper as a front end instead of something hand rolled, so it will likely have better compatibility now https://github.com/voila-dashboards/voila/releases/tag/v0.5.0

python/perspective/setup.py
```diff
@@ -311,6 +311,7 @@ def run_check(self):
         "Programming Language :: Python :: 3.9",
         "Programming Language :: Python :: 3.10",
         "Programming Language :: Python :: 3.11",
+        "Framework :: Jupyter :: JupyterLab :: Extensions :: Prebuilt",
```
timkpaine, line 314: Here are 2 others you might consider adding, though discoverability via this mechanism is not terribly popular https://github.com/timkpaine/ipydagred3/blob/0d8f858b84b7f65d9de5aa5ae59e54a502e3b5b2/pyproject.toml#L30

texodus: https://jupyterlab.readthedocs.io/en/stable/extension/extension_dev.html#pypi-trove-classifiers

timkpaine: Yep that's where I got mine from

PR #2366, opened 2023-09-12
Move styles to sidebar

ada-x64:
This PR moves the datagrid style custom elements to a new tabbed sidebar called the column settings sidebar.

All active columns now have an edit button which will open the column settings. Expression columns can be edited when inactive as well. Plugins other than the datagrid will have a "styles not implemented" placeholder until we eventually add styles to them.

The PR also adds significant playwright scaffolding for UI tests that are not based on snapshots. This includes updating the yarn file to use the latest version of Playwright.

- [x] Create tabbed sidebar
- [x] Move ExprTK editor to tabbed sidebar
- [x] Move Yew components to tabbed sidebar
- [x] Hook up plugins and sidebar
- [x] Make the style buttons on the datagrid plugin open the style sidebar
- [ ] NTH: Datagrid should autoscroll to currently selected column when editing.
- [x] Remove extraneous files
- [x] Add tests

KNOWN ISSUES:
- Column settings sidebar is drawn over the plugin, so it can overlap the focused column.
- Edit button is a bit sticky. Select a column, change something, then select another column. The edit button should move.
- Mousedown on the active columns list edit button always opens the sidebar, even if you drag the selected column.


rust/perspective-viewer/src/rust/components/column_settings_sidebar.rs
```diff
@@ -0,0 +1,128 @@
...
+    pub on_close: Callback<()>,
+    pub session: Session,
+    pub renderer: Renderer,
+    pub presentation: Presentation,
+}
+
+#[function_component]
+pub fn ColumnSettingsSidebar(p: &ColumnSettingsProps) -> Html {
+    let column_name = match p.selected_column.clone() {
+        ColumnLocator::Expr(maybe_name) => maybe_name.unwrap_or("New Column".to_string()),
```
texodus, line 57: This should be an abort - returning `"New Column"` shouldn't ever be reached but if it does, this error condition is just kicked down the road here where it may be difficult to debug.

rust/perspective-viewer/src/rust/components/column_settings_sidebar.rs
```diff
@@ -0,0 +1,128 @@
...
+        }
+    });
+
+    html_template! {
+        <LocalStyle href={css!("column-settings-panel")} />
+        <Sidebar
+            {title}
+            on_close={p.on_close.clone()}
+            id_prefix="column_settings"
+            icon={html_template!{
```
texodus, line 120: IMO this is overkill - the icon data is metadata that can be passed as a prop, it's never going to be some other DOM structure, in this case the `id` attribute.

rust/perspective-viewer/src/rust/components/column_settings_sidebar/attributes_tab.rs
```diff
@@ -0,0 +1,135 @@
...
+#[function_component]
+pub fn AttributesTab(p: &AttributesTabProps) -> Html {
+    let is_validating = yew::use_state_eq(|| false);
+    let on_save = yew::use_callback(
+        |v, p| {
+            match &p.selected_column {
+                ColumnLocator::Expr(None) => save_expr(v, p),
+                ColumnLocator::Expr(Some(alias)) => update_expr(alias, &v, p),
+                _ => {
+                    tracing::error!("Tried to save expression in bad state!")
```
texodus, line 43: `AttributesTab` cannot currently be instantiated for non expression columns (which this error alludes to). We can represent this illegal state in the type system and validate it at compile time by not using `ColumnLocator` in props,  rather just `Option<String>`.

rust/perspective-viewer/src/rust/components/column_settings_sidebar/attributes_tab.rs
```diff
@@ -0,0 +1,135 @@
...
+            if let ColumnLocator::Expr(Some(ref s)) = p.selected_column {
+                delete_expr(s, p);
+            }
+
+            p.on_close.emit(());
+        },
+        p.clone(),
+    );
+
+    let alias = yew::use_memo(
```
texodus, line 70: This doesn't need to be a `use_memo`, just calculate it in the component body. In this case it's not even conserving the `clone()` as you must clone into the closure.

rust/perspective-viewer/src/rust/components/column_settings_sidebar/style_tab.rs
```diff
@@ -0,0 +1,224 @@
...
+#[derive(Serialize, Deserialize, Debug)]
+struct Config {
+    columns: HashMap<String, serde_json::Value>,
+    default_config: Option<DefaultConfig>,
+}
+
+/// This function sends the config to the plugin using its `restore` method
+fn send_config<T: Serialize + Debug>(
+    renderer: Renderer,
+    presentation: Presentation,
```
texodus, line 59: If you pass these by reference, they will not need to be cloned at the call site.

The main place we want to `clone()` + move into a function is when spawning the first frame on a new async stack, as the async frame cannot reference the spawning stack.

rust/perspective-viewer/src/rust/components/column_settings_sidebar/style_tab.rs
```diff
@@ -0,0 +1,224 @@
...
+            .insert(column_name, serde_json::to_value(column_config).unwrap());
+        let js_config = JsValue::from_serde_ext(&current_config).unwrap();
+        elem.restore(&js_config);
+        // we don't need to block yew's rendering to restyle the element,
+        // nor do we need any returned result, nor do we care if it fails.
+        wasm_bindgen_futures::spawn_local(async move {
+            let view = view.unchecked_into();
+            elem.update(&view, None, None, false)
+                .await
+                .expect("Unable to call update!");
```
texodus, line 78: While we don't care if this fails, generally panicking in Yew is bad, as it can leave the app in an inconsistent state. `ApiFuture::spawn` expects a `ApiResult` return type which is compatible with update and correctly routes this failure to the `Promise.reject()` method, both reporting the failure in the console properly _and_ avoiding panic.

rust/perspective-viewer/src/rust/custom_elements/viewer.rs
```diff
@@ -617,4 +617,21 @@ impl PerspectiveViewerElement {
     pub fn unsafe_get_model(&self) -> *const PerspectiveViewerElement {
         std::ptr::addr_of!(*self)
     }
+
+    /// Opens the column edit sidebar for a specific column.
+    /// The Viewer component will then send a
+    /// "perspective-toggle-column-settings" event to the plugin with the
+    /// currently selected column. The plugin should respond to the event to
+    /// highlight the selected column appropriately.
```
texodus, line 625: These are the API docs which are published to https://perspective.finos.org/docs/obj/perspective-viewer/ , and unlike the rest of the repo, they should be written from the point-of-view of a developer using `<perspective-viewer>` in their application, not a developer working on the Perspective codebase itself.

rust/perspective-viewer/src/rust/session/metadata.rs
```diff
@@ -186,6 +187,23 @@ impl SessionMetadata {
...
+                    Some(ColumnLocator::Expr(Some(name)))
+                } else {
+                    meta.column_names
+                        .iter()
+                        .find_map(|n| (n == &name).then_some(ColumnLocator::Plain(name.clone())))
+                }
+            })
+        } else {
+            // new expr
+            Some(ColumnLocator::Expr(None))
```
texodus, line 203: Under what conditions does this function return `None`? I think this is a bit over-wrapped.

texodus (commented):


This looks great overall, and is a really cool feature to boot.

One issue we do need to address is `default_config` in `plugin_config` being persisted. This is likely a bug introduced in this PR, and while its mostly innocuous it is _also_ a breaking change to the package's public API, specifically the structure of the config returned by `save()`.  Perspective's config structure is frequently persisted in hard-to-update systems like Database tables and may be encoded in non-human-readable format, so it's important we conserve changes to this data structure.

IMO we should fix this before we merge this PR, so we can revert the test changes which assert the modified behavior.

rust/perspective-viewer/src/rust/components/column_settings_sidebar.rs
ada-x64: ColumnLocator::Expr(None) is used when creating a new Expression column. However, we shouldn't be hardcoding the name, so I used the same function as in the expression_editor to generate the name.

rust/perspective-viewer/src/rust/session/metadata.rs
ada-x64: This should return None when there is not a currently existing column. I removed the Expr(None) case.

rust/perspective-viewer/src/rust/custom_elements/viewer.rs
ada-x64: I updated the docs to something I think is more user-friendly. Let me know if it needs more changes.

packages/perspective-jupyterlab/test/jupyter/widget.spec.js
```diff
@@ -214,6 +214,7 @@ describe_jupyter(
                 });
 
                 // Check default config
+                config.plugin_config.default_config = {};
```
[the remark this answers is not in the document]
ada-x64: In order to fix this I added a default_config getter to the plugins which act as an opt-in to the style API. If this is not set, the plugins won't get styled.

rust/perspective-viewer/src/rust/components/column_settings_sidebar/style_tab.rs
```diff
@@ -0,0 +1,224 @@
...
+
+/// This function retrieves the plugin's config using its `save` method.
+fn get_config(renderer: &Renderer) -> Option<Config> {
+    let config = renderer.get_active_plugin().unwrap().save();
+    let stringval = js_sys::JSON::stringify(&config)
+        .ok()
+        .and_then(|s| s.as_string())
+        .unwrap_or_default();
+    serde_json::from_str(&stringval).ok()
+}
```
ada-x64, line 96: This has been updated to move the API changes from persistence to a getter.

texodus (approved):

PR #2371, opened 2023-09-28
fix some viewer bugs

ada-x64:
This PR fixes a few small bugs.
- The perspective-datagrid-plugin was throwing some annoying errors when you switched between plugins. This was because we are registering an event listener on the perspective-viewer element, and it is never being unregistered. The only way we can unregister the event listener is to add a `deactivate` method to the API where we can clean up the listeners. For now, I  added a quick hack.
- Fixed some visual bugs regarding borders on the side panels. The top border should no longer shrink and the column settings panel should now go beneath the settings panel when resizing.

### DIFF
--- a/rust/perspective-viewer/src/less/viewer.less
+++ b/rust/perspective-viewer/src/less/viewer.less
@@ -10,6 +10,10 @@
 // ┃ of the [Apache License 2.0](https://www.apache.org/licenses/LICENSE-2.0). ┃
 // ┗━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┛
 
+:host {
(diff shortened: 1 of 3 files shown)


rust/perspective-viewer/src/less/viewer.less
```diff
@@ -10,6 +10,10 @@
 // ┃ of the [Apache License 2.0](https://www.apache.org/licenses/LICENSE-2.0). ┃
 // ┗━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┛
 
+:host {
+    --settings-panel-z-index: 10;
```
texodus, line 14: IMO we want to save settings for theme-able properties, not just constants. Perhaps we can add a special namespace for "private" variables.

texodus (approved):

PR #2373, opened 2023-10-04
Symbols column in XY-scatter plot

ada-x64:
This PR aims to create a Symbols column field in the XY-scatter plot allowing users to specify a mapping between specific row values and symbols. Currently this only supports the basic d3 filled symbols, but in the future we may extend this to allow custom symbols.

CHANGES:
- Rename `default_config` plugin field to `plugin_attributes` and extend its functionality. It now handles general purpose plugin attributes, which determine which attribute and style fields show up in the sidebar.
- Add plugin_attributes to the d3fc plugin and update its restore and save methods to communicate with the viewer.
- Create a "symbols" key-value mapping section in the column settings sidebar's styles tab.
- Rearrange some files in the column settings sidebar for clarity.
- Allow CustomEvents to be triggered from a model, and use this to send configs to/from plugins.
- Always show the Attributes tab; add the "Type" section, and disabled the expression editor for non-expression columns.
- Address view/session and FFI concerns in #2384 
- Add tests

TODO:
- [x] Style the symbol attribute section + general UX improvements
- [x] Check edge cases
- [x] Ensure the plugin maps symbols based on the symbol field (this is currently working for legends but not the chart itself)
- [x] Override certain values based on the passed symbols config
- [x] Add and update tests!

KNOWN ISSUES:
- Incompatible with color-by
- Incompatible with filter (where)
- Incompatible with split-by
- Incorrect behavior with group-by


ada-x64:
https://github.com/ada-x64/perspective/pull/1
This should go on top before merging.

texodus:
Closing for #2394

PR #2374, opened 2023-10-05
Add `index()`, `col()`, and `vlookup()` to ExprTk

tdbgamer:
This adds `index()` which returns the primary key of the current row, `col('column_name')` which is essentially the same as `"column_name"` except that it takes a string literal, and `vlookup('column_name', index)` which will lookup a value in a column by primary key such that `col('a') == vlookup('a', index())`.


texodus:
#2041

texodus:
https://github.com/finos/perspective/assets/60666/78dbfad3-1009-4868-b9fa-7c24bff00609

texodus (approved):

PR #2384, opened 2023-10-10
fix gradient values

ada-x64:
# Bugfix

Number style gradient values were not being populated. This uses the same code as an older version, but reimplemented in rust.

## Testing

No tests were added or changed.

## Checklist

- [x] I have linted my code locally, following the project's code style
- [x] I have tested my changes locally, and changes pass on the Azure Pipelines CI.

### DIFF
--- a/rust/perspective-viewer/src/rust/components/column_settings_sidebar/style_tab.rs
+++ b/rust/perspective-viewer/src/rust/components/column_settings_sidebar/style_tab.rs
@@ -204,19 +204,21 @@ pub fn StyleTab(p: &StyleTabProps) -> Html {
                 NumberColumnStyleDefaultConfig,
             >(&renderer, &column_name, ty)
             .map(|(config, default_config)| {
-                let on_change = Callback::from(move |config| {
-                    send_config(
-                        &renderer,
-                        &presentation,
-                        view.clone(),
-                        column_name.clone(),
-                        config,
-                    );
-                });
+                let on_change = {
+                    clone!(column_name, view);
+                    Callback::from(move |config| {
+                        send_config(
+                            &renderer,
+                            &presentation,
+                            view.clone(),
+                            column_name.clone(),
+                            config,
+                        );
+                    })};
                 html_template! {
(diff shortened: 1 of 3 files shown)


rust/perspective-viewer/src/rust/components/number_column_style.rs
```diff
@@ -237,34 +259,41 @@ impl Component for NumberColumnStyle {
...
-                        self.bg_gradient = self.default_config.bg_gradient;
+                        self.bg_gradient = Some(self.default_config.bg_gradient);
                         self.config.bg_gradient = None;
                     }
                 };
 
                 self.dispatch_config(ctx);
                 false
             }
+            NumberColumnStyleMsg::DefaultGradientChanged(gradient) => {
```
texodus, line 290: Distributing the default value logic across multiple files like means we have the of the complexity of the `default_config`/`config` design, but none of the decoupling.  We should instead IMO make the `default_config()` getter an async method which can query it's container.

texodus (approved):

rust/perspective-viewer/src/rust/components/column_settings_sidebar/style_tab.rs
```diff
@@ -204,19 +204,21 @@ pub fn StyleTab(p: &StyleTabProps) -> Html {
...
+                            view.clone(),
+                            column_name.clone(),
+                            config,
+                        );
+                    })};
                 html_template! {
                     <div class="item_title">{title.clone()}</div>
                     <div class="style_contents">
-                        <NumberColumnStyle  { config } {default_config} {on_change} />
+                        <NumberColumnStyle column_name={column_name.clone()} view={view.clone()}  { config } {default_config} {on_change} />
```
texodus, line 221: Don't pass bind and pass `View` - pass the state object `Session` and request the view immediately where needed. This doesn't matter as much for this change due to the ephemeral nature of this component, but for components that persist between query changes, `View` can not only go stale, any remaining stale references will _segfault_ if you try to call them.

PR #2391, opened 2023-10-13
Fix color legend click hides series

ada-x64:
# Bugfix

This PR updates the legend's filterDataByGroup to account for the case where `settings.hideKeys` has values but `settings.splitValues` is empty. X/Y Scatter plots always use split-by values, so when we color by a string without a split-by we need to account for difference in the column names. If a split-by is set, columns have names of the following form: `splitval1|splitval2|columnName`. Previously, the filterDataByGroup function was relying on `groupFromKey`, which gets the column name's split-by values. This was returning an empty string, causing the legend's click handler to effectively do nothing. The changes in this PR cause the filter to actually filter the data, i.e. remove values. I'm not sure why this isn't being done already, but changing it to a filter seems to break things.

## Testing

No tests were added or changed.

## Screenshots (if appropriate)

https://github.com/finos/perspective/assets/41482263/94145a86-cffe-4e24-b93f-092b3e45ef29

## Checklist

- [x] I have linted my code locally, following the project's code style
- [x] I have tested my changes locally

### DIFF
--- a/packages/perspective-viewer-d3fc/src/js/charts/xy-scatter.js
+++ b/packages/perspective-viewer-d3fc/src/js/charts/xy-scatter.js
@@ -11,7 +11,6 @@
 // ┗━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┛
 
 import * as fc from "d3fc";
-import * as d3 from "d3";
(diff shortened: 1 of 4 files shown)


sinistersnare:
I find that clicking to hide data in a sunburst chart does not work

image

Note that I have clicked a couple of the items in the legend.

Wanted to point this out, I have not checked other chart types though, I can do that when we want to go fix this.

texodus:
This functionality has been superceded by #2394

PR #2399, opened 2023-10-25
Expressions API Updates

ada-x64:
Users can now specify expression columns using either an array of strings or an array of `{name: string, expr: string}` objects. This allows us to more easily rename expressions and separates parsing from the string type.
This API update is backwards-compatible, so plain-string expressions are still parsed correctly. Because of this I have not written any migration code. 

In addition, I have removed some extra stuff from the column settings panel, including the attributes tab "type" field and the entire attributes tab if the column is not an expression.


ada-x64:
Converting this to draft for now. I would like to see the name field be optional so that we can write `{expr: "'foo'"}` for expressions that don't have a distinct name field. This declutters the config for larger unnamed expressions - which will be many of them once migrations happen.

texodus:
Thanks for the PR!

I'm going to go ahead and merge this now as it looks good structurally and there is some backlog behind it. As discussed offline, I'd like to see some changes to both the UX and API before the next release:

* Expresison name `<input>` should move to the attributes header
* Should preserve default behavior of unnamed expressions are just named the expression itself (e.g. when the `input` is empty).
* Key in the dictionary tuple should IMO not be `expr`, but either `expression` (matching the viewer config param `expressions`) or `value` (matching `key`). We do not have to conserve characters over keys like this since they'll generally be compressed, and this is a public API where we should favor readability.

texodus (approved):

Issue #2369, opened 2023-09-26
drop-down lists are broken in chrome


wjackson:
### Steps to Reproduce:

 1. Open the perspective home page in Chrome or Chromium.
 2. Click on any drop-down.

### Expected Result:

I expect to see a drop-down list of all available options.

### Actual Result:

I actually see a a drop-down list with only the currently selected option visible.

### Environment:

This bug was introduced with version 2.3.1.

I know this because I have the same problem on an internal app. Works as expected on 2.3.0. Broken on 2.3.1.

I observed this bug on these platforms:

1. Fedora Linux 38 +  Chrome Version 117.0.5938.88 (Official Build) (64-bit)
2. RHEL 9.2 + Chromium Version: 116.0.5845.96 (Official Build) Fedora Project (64-bit)
3. Windows 10 Enterprise, Version: 22H2, OS build: 19045.3324, Experience: Windows Feature Experience Pack 1000.19041.1000.0 + Chrome: Version 115.0.5790.110 (Official Build) (64-bit)

### Additional Context:

I've only observed the problem on Chrome/Chromium. Firefox is fine.

Chrome:

![perspective-dropdown-chrome](https://github.com/finos/perspective/assets/315342/5a087753-a0c7-49bf-96dd-6d9f7a1ed218)

Firefox:

![perspective-dropdown-firefox](https://github.com/finos/perspective/assets/315342/3e4deb88-15f3-4179-9b3b-4cbed82cba66)

texodus:
@wjackson Thanks for the repot, but _please_ follow the bug template - we need your OS+version to identify this.

wjackson:
> @wjackson Thanks for the repot, but _please_ follow the bug template - we need your OS+version to identify this.

Cool. I added that info to the bug report. Let me know if there's any other info I can provide that might help.

autumn-absconds:
@texodus @wjackson I checked this (same thing is happening in Edge as in Chrome), I might be wrong, I don't know about the term broken but this
 ### problem is arising because of the dark mode ( the background-color used in select element ).


![Screenshot (13)](https://github.com/finos/perspective/assets/65112908/026a2e5f-6102-44f1-ae41-542c41bd714c)
 
----
here as hovering on different values displays itself
----
![Screenshot (14)](https://github.com/finos/perspective/assets/65112908/1732b47c-fbfc-4f2d-a1d8-5a9acfdc876b)

----
here on unselecting the bg class solves this 
----
[...]

PR #2400, opened 2023-10-25
let themes control select and option colors

wjackson:
This is a fix for #2369.

### DIFF
--- a/rust/perspective-viewer/src/less/dom/select.less
+++ b/rust/perspective-viewer/src/less/dom/select.less
@@ -51,7 +51,8 @@
         background-repeat: no-repeat;
 
         option {
-            background: var(--select--background-color, transparent);
+            color: var(--option--color, transparent);
+            background-color: var(--option--background-color, transparent);
         }
     }
 }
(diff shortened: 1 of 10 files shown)


ada-x64 (commented):
Tested this on my Manjaro Linux machine, can confirm that the changes fix the bug. LGTM! cc @texodus

texodus:
Thanks for the PR @wjackson!

Adding new CSS variables for these properties requires this variable to be specified for all themes, as you've noticed. As there is not compelling reason to style option background _in isolation_, this shouldn't be public (or at the very least, should default to the themes' palette).

I can't test this as its a _Windows_ bug (not a Chrome bug as the Issue says), but I've added this default in #2422 anyway, closing

wjackson:
> there is not compelling reason to style option background in isolation

Yeah that makes sense.

> I can't test this as its a Windows bug (not a Chrome bug as the Issue says), but I've added this default in https://github.com/finos/perspective/pull/2422

The screenshots in #2369 came from Chrome on Linux so it's not a Windows bug. As mentioned in the issue, I tested for and found the bug on a couple different Linuxes and Windows.

I tested your fix (`upgrade-emsdk:017b3d301`)  on Linux where I originally found the bug and it works. Thanks!

PR #2404, opened 2023-10-27
Fix scroll panel height calculation bug

texodus:
Perspective's virtual scroll panels use a size-hint system to create the panels geometry without rendering or measuring all children (of which there may be thousands). However, these values have bitrotted. This PR fixes the hints to reflect the current geometry, but also adds a validation step to log in debug mode when the hints are incorrect.

I don't think this is worth an complex sizing/scroll system like `regular-table` has, but it woul dbe nice to be able to specif these hard-coded geometry values in custom CSS rules at least.

Before:
Screenshot 2023-10-27 at 12 21 46 AM

After:
Screenshot 2023-10-27 at 12 22 44 AM


ada-x64:
This looks good! Tested on MacOS and Linux with Firefox and Chrome.

PR #2413, opened 2023-11-04
Add requirements regenerator script and action

timkpaine:
This PR does 3 things

- Add helper script for regenerating python frozen dependencies. We want CI to be stable, but we also need to move dependencies forward, so hopefully this will help with that. This change also comes with some slight refactoring.
- Add a github workflow for doing bullet 1 across all python versions, so we don't need to rely on having all pythons installed locally. It will push to a well-known branch name `tkp/update-frozen-python-dependencies`, which we can then PR to master
- Uses bullet 1 to update all deps, in a second commit

From my fork:

Screenshot 2023-11-04 at 16 36 20

Screenshot 2023-11-04 at 17 47 37

Screenshot 2023-11-04 at 18 09 10


texodus:
Thanks for the PR!

I understand why you want to upgrade the CI deps, but I don't want this (or any dev task) done on CI. These resources are of limited availability, slow, prone to environmental regressions and extremely hard to debug/develop against locally.

I've added a `pip` method for regenerating these files in #2422 you can try to use as an alternative.

timkpaine:
I don't have python 3.8, 3.9, or 3.10 installed locally, does the script handle installing those python versions?

PR #2418, opened 2023-11-06
`PerspectiveWidget` HTML export support in Jupyter

tomjakubowski:


python/perspective/perspective/__init__.py
```diff
@@ -16,4 +16,8 @@
 from .client import *
 from .handlers import *
 from .nbextension import _jupyter_nbextension_paths
-from .widget import *
+from .widget.widget import PerspectiveWidget
```
tomjakubowski, line 19: this is an affordance to make `%autoreload` magic work better. happy to revert

python/perspective/perspective/templates/bruce.html.jinja
```diff
@@ -0,0 +1,40 @@
+<script type="module" src="https://cdn.jsdelivr.net/npm/@finos/perspective/dist/cdn/perspective.js"></script>
```
timkpaine, line 1: Should we try to parameterize these to the same versions as used by the client?

python/perspective/perspective/widget/widget.py
```diff
@@ -514,4 +535,31 @@ def _make_load_message(self):
...
+
+        # Serialize viewer attrs + view data
+
+        viewer_attrs = self.save()
+        # simple case: serialize the view to arrow, embed it as base64
+        data = self.table.view().to_arrow()
+        b64_data = base64.encodebytes(data)
+
+        jinja_env = jinja2.Environment(
+            loader=jinja2.PackageLoader("perspective"),
```
timkpaine, line 553: Inject versions to load from jsdelivr here

python/perspective/perspective/widget/widget.py
```diff
@@ -514,4 +535,31 @@ def _make_load_message(self):
...
+            autoescape=jinja2.select_autoescape(),
+        )
+        template = jinja_env.get_template("bruce.html.jinja")
```
timkpaine, line 556: Can inject version here

python/perspective/perspective/templates/bruce.html.jinja
tomjakubowski: yeah for sure

python/perspective/perspective/widget/widget.py
```diff
@@ -514,4 +535,31 @@ def _make_load_message(self):
...
+        viewer_id = f"viewer-{self.model_id}"
+
+        return super_bundle | {
+            "text/html": template.render(
+                viewer_id=viewer_id,
+                viewer_attrs=viewer_attrs,
+                b64_data=b64_data.decode("utf-8"),
+            )
```
timkpaine, line 564: Version=whatever so we pull same js assets

python/perspective/setup.py
```diff
@@ -66,7 +66,7 @@
 
 requires_aiohttp = ["aiohttp>=3,<4"]
 
-requires_jupyter = ["jupyterlab>=3.2,<4"]
+requires_jupyter = ["jupyterlab>=3.2,<4", "Jinja2>=3.0,<4"]
```
timkpaine, line 69: Use 2 as lower bound if you can

python/perspective/perspective/__init__.py
timkpaine: We need to go through and make sure all our * imports pull from files with `__all__`, this is a good first issue I'll put in another ticket

timkpaine: https://github.com/finos/perspective/issues/2419

tomjakubowski:
As it is currently, rendering the HTML output is unconditional and this will add some bloat to all .ipynb using `PerspectiveWidget`. Every render of a PerspectiveJupyterWidget adds the HTML output to its cell's output in the `.ipynb` . This includes a base64-serialized arrow of every viewer's data, without deduplication, and a few lines of Javascript and HTML to glue it together.

We can make this configurable and/or optional in Perspective, for users who don't care to support exporting. Would a global setting or per-widget setting make sense?

timkpaine:
per-widget, what about if I stream data in and call update() a bunch of times, will it reflect the latest?

tomjakubowski:
I don't have very much experience doing streaming perspectives but I'll check. I think it reflects whatever was in the table at the time the cell first rendered, which is probably undesirable.

tomjakubowski:
Added a class variable on `PerspectiveWidget` to gate this new feature behind an undocumented flag while we're tooling on it. Also fixed a bug in datagrid

It can be set on the class to make all widgets render HTML output, or on an instance to make only that widget do it

python/perspective/perspective/widget/widget.py
```diff
@@ -201,6 +203,8 @@ class PerspectiveWidget(DOMWidget, PerspectiveViewer):
     _view_module = Unicode("@finos/perspective-jupyterlab").tag(sync=True)
     _view_module_version = Unicode("~{}".format(__version__)).tag(sync=True)
 
+    _enable_experimental_html_export = False
```
texodus, line 206: It's ok to make this non-default but let's make it a real API. My proposal is two-fold:
1. Environment variable `PSP_JUPYTER_HTML_EXPORT` whose presence enables this behavior, so `nbconvert` + `nbclient` can be called from the command line.
2. A `import set_jupyter_html_export from perspectve` method which sets this env variable for the current process, so the behavior can be defined in the notebook itself (alternatively they could just set the env var but I think this is cleaner and parallels e.g. `set_num_threads()` global config APIs we already expose).

python/perspective/perspective/templates/exported_widget.html.jinja
```diff
@@ -0,0 +1,41 @@
...
+        const currentScript = document.scripts[document.scripts.length - 1];
+        const envelope = document.getElementById(`perspective-envelope-${viewerId}`);
+        const dataScript = envelope.querySelector('script[type="application/vnd.apache.arrow.file"]');;
+        if (!dataScript)
+            throw new Error('data script missing for viewer', viewerId);
+        const data = base64ToBytes(dataScript.textContent);
+        const viewerAttrs = {{ viewer_attrs | tojson }};
+
+        // Create a new worker, then a new table promise on that worker.
+        const table = await perspective.worker().table(data.buffer);
```
texodus, line 36: Use [`shared_worker`](https://github.com/finos/perspective/blob/master/packages/perspective/src/js/perspective.browser.js#L195) or you will get a WebWorker per cell.

TODO we should probably add documentation for `shared_worker` and friends as we apparently don't yet generate JsDocs anymore for this module.

python/perspective/setup.py
```diff
@@ -66,7 +66,7 @@
 
 requires_aiohttp = ["aiohttp>=3,<4"]
 
-requires_jupyter = ["jupyterlab>=3.2,<4"]
+requires_jupyter = ["jupyterlab>=3.2,<4", "Jinja2>=2.0,<4"]
```
texodus, line 69: IMO this is just a regular requirement. `requires_jupyterlab` is used to install Jupyter Lab in CI, real users will likely already have this installed in their environment and would never think to install the undocumented `pip install perspective-python[jupyterlab]`, making this extra `Jinja` dep easy to miss.

Also please pin this in the CI `requirements` files.

texodus (changes requested):
Thanks for the PR!

python/perspective/setup.py
tomjakubowski: I misunderstood what requires_jupyter was for… appreciate the explanation!

python/perspective/perspective/__init__.py
tomjakubowski: reverted this

python/perspective/perspective/widget/widget.py
tomjakubowski: Left is running jupyterlab with `PSP_JUPYTER_HTML_EXPORT=1`, right is running without the environment variable set. It would have been cool including a cell with the value of `os.environ.get("PSP_JUPYTER_HTML_EXPORT", None)` but I didn't think of it in time.

Screenshot 2023-11-14 at 3 49 59 PM

python/perspective/setup.py
tomjakubowski: oops, need to pin this in the requirements file.

tomjakubowski: I bumped the deps using the new update_python script

It made a really noisy diff which was made a lot cleaner once I sorted the lines in the updated requirements-*.txt files. Maybe this is something we could add to the update_python script? (or to some documentation about how to update deps, which I'm happy to write)

tomjakubowski: Added sorting to the requirements.txt updater

PR #2421, opened 2023-11-08
Column Settings Panel UI Tweaks

ada-x64:
Based on #2399 for column settings panel changes.

- Hide edit button for cols with no styles or attrs
- Hide the style tab unless needed (hack!)
	- In order to do this in a sane way, we need to update the plugin_attributes API.
- Add a size trap for column settings panel width
- Update column settings panel title to be less verbose
- Update tests


texodus (approved):

PR #2422, opened 2023-11-09
Add CI requirements update script

texodus:


tools/perspective-scripts/update_python.mjs
```diff
@@ -0,0 +1,56 @@
...
+    const data = JSON.parse(fs.readFileSync("report.json"));
+    let output = "";
+    for (const {
+        metadata: { version, name },
+    } of data.install) {
+        output += `${name}==${version}\n`;
+    }
+
+    fs.writeFileSync(
+        `python/perspective/requirements-${version.replace(".", "")}.txt`,
```
timkpaine, line 51: Can we use this opportunity to move the requirements to a folder and exclude the folder from distribution (in MANIFEST.in)

texodus: [Why?](https://files.pythonhosted.org/packages/bf/05/b1af85af0ca4c2617b75b5acd6b8ced6d01b1bc57d0e0813106e29f9edb2/perspective-python-2.6.1.tar.gz)

timkpaine: They're only used for testing, and it's confusing because many other libraries use similarly named files as their runtime deps

texodus: What is the difference between CI deps and runtime deps?

timkpaine: When an end user does `pip install perspective-python` and installs perspective from pypi, they will pull down a dynamic set of dependencies based on their currently installed dependencies and perspective's defined bounds. These are our runtime deps. In CI, we don't do this, we install a fixed set of known-good dependencies to run lints and tests against. These are our CI deps.

Our runtime deps are defined in `setup.py`, but some repos use `requirements.txt`. I think it would be good to clearly put these deps in their own folder so people don't look and think "Oh im on Python3.8, perspective only supports the fixed set of pinned dependencies in requirements-38.txt and that doesn't work for me". A good example of this is airflow, which creates the same type of "known-good" requirements files as we do, but they put them in their own space: https://github.com/apache/airflow/tree/constraints-2-7.

texodus: You'd have to dig 2 levels deeps into our source repo to even know these files exist. How is moving them to 3 levels deep going to correct developers' expectations?

timkpaine: They're at the root of our python project.

texodus: [`pandas`](https://github.com/pandas-dev/pandas) has the same configuration we have - developer/ci requirements in `requirements-dev.txt` in the root of the project next to `setup.py`, and no one is confused. Moreover, the [`pip` documentation](https://pip.pypa.io/en/stable/reference/requirements-file-format/) says that the file name can be anything, and mentions absolutely nothing about conventional use.

I will change this because I'm tired of discussing it, but I'm doing so under protest.

dhirschfeld: As a user who is heavily involved in python packaging, I can confirm that there is a strong convention for storing runtime dependencies in a file named `requirements.txt`.

Using a file named `requirements.txt` for anything other than runtime dependencies would be quite surprising/confusing.

The convention is also extended to optional dependencies (or [extras](https://peps.python.org/pep-0508/#extras)) where `requirements-<name>.txt` maps to the optional dependency group `<name>`. It is quite often that you will see `requirments-dev.txt`, `requirements-test.txt` and `requirements-docs.txt` files alongside the `requirements.txt` file.

This convention is strong enough that there is a [`hatch-requirements-txt`](https://github.com/repo-helper/hatch-requirements-txt) plugin for the [`hatch`](https://hatch.pypa.io/latest/) packaging tool to dynamically map the contents of these files into the `pyproject.toml`.

`requirements-38.txt` *isn't* `requirements.txt` but perhaps it's close enough to still cause confusion. If it costs nothing to reduce the possibility of confusion, that's probably a good thing.

PR #2429, opened 2023-11-13
X/Y Scatter Symbol Serialization as Dictionary

ada-x64:
This PR changes the X/Y Scatter plot's symbol serialization format to a dictionary. It also fixes a bug in the datagrid. (Editable bools were not toggling correctly.)


texodus (approved):

PR #2430, opened 2023-11-13
API Versioning + Migration

ada-x64:
Based on #2429

This PR adds versioning to the perspective viewer's save/restore API.
It also adds a new format for migrations and adds migrations for the new expression and X/Y scatter symbol serialization formats.

### DIFF
--- a/docs/docs/development.md
+++ b/docs/docs/development.md
@@ -98,6 +98,18 @@ To install a specific version of Emscripten (e.g. `2.0.6`):
 
 ---
 
+## `perspective-python`
+
+To build the Python library, first configure your project to build Python via
+`yarn setup`. Then, install the requirements corresponding to your version of
+python, e.g.
+
+```bash
+pip install -r python/perspective/requirements-311.txt
+```
+
+`perspective-python` supports Python 3.8 and upwards.
+
 ### `perspective-jupyterlab`
 
 To install the Jupyterlab/Jupyter Notebook plugins from your local working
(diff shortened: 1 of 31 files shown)


docs/docs/development.md
```diff
@@ -98,6 +98,18 @@ To install a specific version of Emscripten (e.g. `2.0.6`):
...
+To build the Python library, first configure your project to build Python via
+`yarn setup`. Then, install the requirements corresponding to your version of
+python, e.g.
+
+```bash
+pip install -r python/perspective/requirements-311.txt
+```
+
+`perspective-python` supports Python 3.8 and upwards.
+
```
texodus, line 112: `yarn _requires_python` is the correct way to do this. While this is not documented, a good place to look is always the [CI actions yaml](https://github.com/finos/perspective/blob/master/.github/workflows/build.yml)

docs/docs/development.md
```diff
@@ -108,17 +120,18 @@ Afterwards, you should see it listed as a "local extension" when you run
...
+As an example, your setup process might look like this:
 
 ```bash
+python -m venv ./venv
+pip install -r python/perspective/requirements-311.txt
+yarn setup # choose python
 yarn build
+yarn setup # choose javascript > jupyterlab
+yarn build
+yarn jlab_link # run this whenever you need to update a local perspective package
```
texodus, line 132: This doesn't actually work due to a bug. `yarn jlab_link` currently _installs_ `perspective-python` in your virtual env, including the python and C++ extension, so further rebuilds in your source repo won't get picked up.

texodus (approved):

It should be noted that versioning in this change is runtime-transparent, the version field is not read or validated in any way by `<perspective-viewer>` at runtime, only in wrt the migration script. IMO we either need to warn on all deprecated versions, or at least specifically warn of the version when a `restore()` call fails in addition to the actual error.

The dev notes are not correct, but I will update these.

PR #2436, opened 2023-11-16
Serialize Expressions to Objects

ada-x64:
This PR changes expression serialization so that expressions will always serialize to objects while allowing users to import array-based expressions.


rust/perspective-viewer/src/rust/config/expressions.rs
```diff
@@ -13,110 +13,84 @@
...
 use std::collections::HashMap;
 
-use itertools::Itertools;
 use serde::{Deserialize, Serialize};
 
-#[derive(Serialize, Deserialize, Clone, PartialEq, Debug)]
+#[derive(Deserialize, Clone, PartialEq, Debug)]
 #[serde(untagged)]
-pub enum Expressions {
+pub enum ExpressionsDeserde {
```
texodus, line 20: What makes this type unique that it requires the qualifier `Deserde` in the name (as opposed to the other components of `ViewerConfig`)? There is no other `Expressions` type, so the fact that it only supports deserialization doesn't disambiguate it.

texodus (approved):

PR #2437, opened 2023-11-16
Move Expression Names to Column Settings Header

ada-x64:
~~Based on #2436 and #2421~~

This PR moves expression names to the column settings header.
To save the expression, press enter or tab.
Currently when you save the expression name the panel closes. I was originally planning on having the column auto-save, but keeping the panel open requires some significant changes.


texodus (changes requested):
Thanks for the PR!

This has some bugs and feature regressions that need to be fixed up:

* We've lost the default naming feature - this was _also_ lost in #2399 and re-added in #2431. Please add a test for this case as well - we _really_ should have caught this and I'm quite surprised the regression made it this far as it involved deleting _alot_ of code, twice. Was this a rebase gone wrong?
* The title clear button has disappeared - regression of the `<input type="search">`.
* In these invalid cases - the save button still works and lets you save columns with broken names.

Before - title reflects the default when empty and has a clear button:


https://github.com/finos/perspective/assets/60666/2f181baa-ad7d-45bf-af05-7b6a59dfccef


Now - title shows an error state that was added, but this is incompatible with the default naming functionality which was removed (?), and the clear button is gone:

https://github.com/finos/perspective/assets/60666/f7a44424-691c-45da-abd2-28320f70d126

ada-x64:
I've updated this PR to address the feedback.
There are still a few UX questions I think need to be addressed.

0. When the title is changed, should the save button for the expression editor be enabled?
0. Should saving the expression also save the title?
0. Should saving the title (pressing enter on the header) also save the expression?

I'm inclined to say that an unsaved title should be updated when the expression is saved because we are currently closing the column settings sidebar on change. However, with #2459 we'll be keeping the sidebar open. In that case, I'm inclined to say that saving the title and the expression content should be separate operations.

texodus:
This is really well done!

However, it honestly it took me a few minutes to understand what was going on :)

My confusion came from muscle-memory, I tried _editing_ an expression column, then changing the title, but _not_ pressing "enter", instead using the mouse to move focus to the expression editor. In this case, the "title" control shows the dashed-border "edited" state, but the expression editor save button is disabled. From this state it wasn't obvious to me that I needed to re-focus the "title" input and press "enter" to submit the changed title!

ada-x64:
> This is really well done!
> 
> However, it honestly it took me a few minutes to understand what was going on :)
> 
> My confusion came from muscle-memory, I tried _editing_ an expression column, then changing the title, but _not_ pressing "enter", instead using the mouse to move focus to the expression editor. In this case, the "title" control shows the dashed-border "edited" state, but the expression editor save button is disabled. From this state it wasn't obvious to me that I needed to re-focus the "title" input and press "enter" to submit the changed title!

I agree, the UX on this is pretty confusing. I originally wanted to do an auto-save feature, where the column title would update on blur. That lead to an _awful_ experience where the sidebar would close all the time. That was the impetus for https://github.com/finos/perspective/pull/2459. I'd be happy to talk more about the UX with this feature; fwiw auto-save felt very nice, and it was cool to see the name automatically update in three places (in the datagrid, the title itself, and the settings panel).

ada-x64:
Closing in favor of #2459

PR #2443, opened 2023-11-20
Refactor How Column Style Components Are Selected

ada-x64:
This PR does what it says on the tin. The goal is to prevent unnecessary style component renders and to replace the default "no styles available" content with a hard error.

There are some remaining issues with the symbol column settings when aggregating. In particular, the column value selector does not get the aggregated values, but the raw table values, so the key-value selector doesn't match up with what is in the plugin.

### DIFF
--- a/rust/perspective-viewer/src/rust/components/column_selector/active_column.rs
+++ b/rust/perspective-viewer/src/rust/components/column_selector/active_column.rs
@@ -70,10 +70,16 @@ impl ActiveColumnProps {
         }
     }
 
-    fn get_type(&self) -> Option<Type> {
+    fn get_table_type(&self) -> Option<Type> {
         self.get_name()
             .as_ref()
             .and_then(|x| self.session.metadata().get_column_table_type(x))
+    }
+
+    fn get_view_type(&self) -> Option<Type> {
+        self.get_name()
+            .as_ref()
+            .and_then(|x| self.session.metadata().get_column_view_type(x))
     }
 }
 
@@ -155,7 +161,6 @@ use ActiveColumnMsg::*;
 #[derive(Default)]
 pub struct ActiveColumn {
     add_expression_ref: NodeRef,
-    column_type: Option<Type>,
     is_required: bool,
     mouseover: bool,
 }
@@ -165,17 +170,14 @@ impl Component for ActiveColumn {
     type Properties = ActiveColumnProps;
 
     fn create(ctx: &Context<Self>) -> Self {
-        let column_type = ctx.props().get_type();
         let is_required = ctx.props().get_is_required();
         Self {
-            column_type,
             is_required,
             ..Default::default()
         }
     }
 
     fn changed(&mut self, ctx: &Context<Self>, _old: &Self::Properties) -> bool {
-        self.column_type = ctx.props().get_type();
         self.is_required = ctx.props().get_is_required();
         true
     }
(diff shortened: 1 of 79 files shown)


rust/perspective-viewer/src/rust/components/column_selector/active_column.rs
```diff
@@ -333,8 +334,14 @@ impl Component for ActiveColumn {
...
-                    "Datagrid" => col_type != Type::Bool,
-                    "X/Y Scatter" => col_type == Type::String && label.as_deref() == Some("Symbol"),
+                    "Datagrid" => table_type != Type::Bool,
+                    "X/Y Scatter" => {
+                        ctx.props()
+                            .get_view_type()
+                            .map(|ty| ty == Type::String)
+                            .unwrap_or_default()
+                            && label.as_deref() == Some("Symbol")
+                    },
```
texodus, line 344: This is getting outlandish :)

rust/perspective-viewer/src/rust/components/column_settings_sidebar/style_tab/column_style.rs
```diff
@@ -163,64 +162,8 @@ pub fn ColumnStyle(p: &ColumnStyleProps) -> Html {
...
-
-        { components }
-        if !is_ok && !is_symbol {
-            <div class="style_contents">
-                <div id="column-style-container" class="no-style">
-                    <div class="style-contents">{ "No styles available" }</div>
-                </div>
-            </div>
-        }
+        {opt_html.expect_throw("Could not generate HTML to style this column!")}
```
texodus, line 167: Errors are not the right way to handle this (or really in general), not even JavaScript errors. It is _really_ hard to predict when something that is the result of a long complex process will adhere to restriction like this that cannot (or at least, isn't) be represented in the type system. More importantly, throwing an error here can corrupt/crash the rust app.

There is a similar circumstance in [`active_column.rs`](https://github.com/finos/perspective/blob/master/rust/perspective-viewer/src/rust/components/column_selector/active_column.rs#L394), where there is a small corner case where the `View` must exists before the UI can be rendered. This render path shouldn't be hit, so we render a stub.

In this case, we probably want to keep the stub, we just don't want it ever to show. If for some reason it does, we don't want to segfault, so it needs to render something because we can't (AFAIK) prevent this logic from being called with a missing view type (?). Probably also worth warning the console as this is programmer error.

rust/perspective-viewer/src/rust/components/column_settings_sidebar/sidebar.rs
```diff
@@ -132,7 +136,8 @@ pub fn ColumnSettingsSidebar(p: &ColumnSettingsProps) -> Html {
                             { renderer }
                             { custom_events }
                             { selected_column }
-                            { on_close }/>
+                            { on_close }
+                            />
```
texodus, line 140: We need to sync up on style guide for `html!` macro content, because it is immune to `rustfmt` and changes like these make it difficult for me to use file history to bisect issues.

texodus (changes requested):
Thanks for the PR!

`expect_throw()` isn't something we should use in the middle of `yew` component tree evaluation I think. Your options here are (in order of preference):
1. Prevent these cases in the type system if possible so they can be factored away at compile time. (sound like not the case here but mentioning it all the same).
2. If this is not possible, render a blank stub and spam the console with warnings, plus a test for this functionality would be nice.

ada-x64:
This has been updated to add a stub, remove expect_throws, and add a fixture test. In order to add the fixture test, I rebased this on https://github.com/finos/perspective/pull/2447

texodus (approved):
Looks good!

PR #2444, opened 2023-11-20
Fix 2.7.0 Migration

ada-x64:
This PR ensures that the version created by the migration script matches the version in the package.json. It does so by inlining the package version and bumping the version as needed.

### DIFF
--- a/rust/perspective-viewer/.gitignore
+++ b/rust/perspective-viewer/.gitignore
@@ -1,1 +1,0 @@
-src/ts/package.json
--- a/rust/perspective-viewer/.gitignore
+++ b/rust/perspective-viewer/.gitignore
@@ -1,0 +1,1 @@
+src/ts/package.json
--- a/rust/perspective-viewer/build.js
+++ b/rust/perspective-viewer/build.js
@@ -80,12 +80,13 @@ async function build_all() {
     execSync(`cargo bundle ${IS_DEBUG ? "" : "--release"}`, INHERIT);
 
     // JavaScript
+    const { default: cpy } = await cpy_mod;
+    await cpy("package.json", "src/ts");
     execSync("npx tsc --project tsconfig.json", INHERIT);
     await Promise.all(BUILD.map(build)).catch(() => process.exit(1));
     await Promise.all(POSTBUILD.map(build)).catch(() => process.exit(1));
 
     // legacy compat
-    const { default: cpy } = await cpy_mod;
     await cpy("target/themes/*", "dist/css");
     await cpy("dist/pkg/*", "dist/esm");
 }
--- a/rust/perspective-viewer/build.js
+++ b/rust/perspective-viewer/build.js
@@ -80,13 +80,12 @@ async function build_all() {
     execSync(`cargo bundle ${IS_DEBUG ? "" : "--release"}`, INHERIT);
 
     // JavaScript
-    const { default: cpy } = await cpy_mod;
-    await cpy("package.json", "src/ts");
     execSync("npx tsc --project tsconfig.json", INHERIT);
     await Promise.all(BUILD.map(build)).catch(() => process.exit(1));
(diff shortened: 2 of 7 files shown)


rust/perspective-viewer/.gitignore
```diff
@@ -0,0 +1 @@
+src/ts/package.json
```
texodus, line 1: Don't add arbitrary files to `.gitignore`, tasks like `twine` and `npm publish` are not aware of these directives which is quite dangerous. If you need to create files as part of the build process, put them in `/build` so they are cleaned and not published - if they need to be published put them in `/dist`.

rust/perspective-viewer/build.js
```diff
@@ -80,12 +80,13 @@ async function build_all() {
     execSync(`cargo bundle ${IS_DEBUG ? "" : "--release"}`, INHERIT);
 
     // JavaScript
+    const { default: cpy } = await cpy_mod;
+    await cpy("package.json", "src/ts");
```
texodus, line 84: Don't copy files within the source tree (see above notes on `/build` and `/dist`), but also _especially_ don't do this with the `package.json`, which breaks node's module resolution in this directory (or any package root metadata like `Cargo.toml`).

rust/perspective-viewer/src/ts/migrate/2-6-1.ts
```diff
@@ -19,12 +19,12 @@ import { cmp_semver, parse_semver } from "../migrate";
...
 export default function migrate_2_6_1(old, options) {
-    if (cmp_semver(old.version, "2.7.0")) {
-        return;
+    if (cmp_semver(old.version, "2.6.1")) {
+        return old;
     } else if (options.warn) {
         console.warn("Migrating from 2.6.1");
     }
-    old.version = parse_semver("2.7.0");
+    old.version = parse_semver("2.6.1");
```
texodus, line 27: This line is inert (because it is overwritten by `assure_latest()`), but if it weren't it would be wrong. If `old == 2.6.1`, the migration is applied (conditional on line 22 is false) but then the version field is set back to `2.6.1` again (line 27). 

This specific script functionally upgrades _from_ `2.6.1` _to_ `2.7.0` (and presumably beyond), ergo `old.version` should be some version after `2.6.1` (which we wouldn't know at the time of authoring the migration).

texodus (changes requested):
Thanks for the PR!

Please find another way to do this without copying `package.json` around.

ada-x64:
@texodus I've updated this PR with some general cleanup of the migration scripts and I've replaced the copied file with an import from node_modules.

I think there is still some thinking to be done about how we're going to go about migrations in the future.
The way I understand it, there are two ways to do API versioning. I'll call them _from x_ and _to x_. 

*To-x* migration scripts migrate a configuration *from* version `x-1` *to* `x`. For example, 2-6-1.ts would migrate *from* 0.0.0 *to* 2.6.1.

*From-x* migration scripts migrate a configuration *from* version `x` *to* `x+1`. For example, script 2-6-1.ts would migrate *from* 2.6.1 *to* 2.7.0.

We're currently implementing the _from-x_ migrations. The main issue with this approach, as we've discussed, is that there is no way of knowing what the next version will be. I've worked around this by simply placing the latest package version on top. In addition to fixing the latest-package version, the troubles we were having with versioning I think were caused by the mere fact that version migrations did not take the package version into account, which has been fixed in this PR.

There are other problems. Because we'll be developing API changes on master, and those migrations will be happening on the _same version_ of perspective as the previous API version, we're still going to have to deal with a form of idempotency in the migrations. That is, we'll need to make sure that the current migration script can upgrade from any version x-n to version x, and stay at version x without change. (i.e., `convert(convert({version: x-n})) == convert({version: x-n})` This shouldn't be too hard to accomplish, but is just something we'll need to keep in mind.

texodus (approved):

PR #2447, opened 2023-11-22
Remove Any Old Expressions from Codebase

ada-x64:
This PR looks huge but it's actually quite small!

It does two things:
1. Datagrid was using an old-style expression when populating the view. This has been fixed.
2. I added a fixture to the playwright tests to ensure that no old-style expressions are still around.
 
Main changes occur in:
- `tools/perspective-test/src/js/fixture.ts` (adds fixture)
- `rust/perspective-viewer/test/js/settings.spec.js` (uses the new log tracking functionality)
- `packages/perspective-viewer-datagrid/src/js/model/create.js` (fixes datagrid bug)

The rest are dependency updates to use the fixture.


ada-x64:
I've updated this PR to add a [timeout configuration](https://playwright.dev/docs/api/class-page/#page-set-default-timeout) to the fixture. Now any function that can take a `timeout` option will wait for at most 5 seconds. This should save us substantial time on `waitFor` and other locator functions.

PR #2454, opened 2023-11-27
Fix Blocks build script execution

sinistersnare:
Also included is some style and spelling fixes.

### DIFF
--- a/examples/blocks/index.mjs
+++ b/examples/blocks/index.mjs
@@ -14,6 +14,7 @@ import * as fs from "fs";
 import { get_examples, LOCAL_EXAMPLES } from "./examples.js";
 import sh from "@finos/perspective-scripts/sh.mjs";
 import * as url from "url";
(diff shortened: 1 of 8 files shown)


yarn.lock
```diff
@@ -2834,11 +2834,121 @@
     url-loader "^4.1.1"
     webpack "^5.73.0"
 
+"@esbuild/android-arm64@0.19.8":
```
texodus, line 2837: You've added esbuild `0.19.8`, but the project workspace uses `0.14.54`, hence the new dependencies in this file. Please upgrade the rest of the project so we're removing this dupe - esbuild is a binary platform-specific blob we need to download and has a non-trivial impact on the dev dependency footprint.

texodus (changes requested):
Thanks for the PR! 

Please fix the `esbuild` dep version

texodus (approved):

PR #2459, opened 2023-11-30
Expression Names in Column Settings Header + Refactors

ada-x64:
Based on https://github.com/finos/perspective/pull/2461, https://github.com/finos/perspective/pull/2437

Many things happen in this PR.
- Expression names get moved to the column settings sidebar's header.
- This header is editable when the attributes tab is open.
- The expression editor's save, reset, and delete buttons have been moved to their own components to allow them to save and reset all attribute settings. This required a good bit of refactoring.
- Column settings is now able to stay open when saving. This is accomplished by updating the viewer's "open column" message, incorporating the last opened column in the presentation, and sending that message as needed.
- Expression types hadn't been updated in the perspective .d.ts file, so I fixed that
- Added a TypeIcon component to replace the ::before rules on column selectors. This could be extended into a generalized Icon component.
- Found a pattern to simplify parameters, i.e. including the parameters for child components as parameters of the wrapper component. This simplified several components in the settings sidebar.
- Moved expression update functions to a model.


ada-x64:
I updated this PR to include about 40 new tests to cover the behavior changes. This helped me catch a number of bugs.
I was having some issues with flakiness in the tests (timing issues), but those should be fixed.
Part of that bugfixing was adding more functionality in response to the `renderer.session_updated` pubsub event. This occurs in the viewer component and in the column settings sidebar. There are again some issues with the plugin API which need to be addressed, but I plan on covering those as I finish #2469 .

texodus (approved):

Reviewed (extensively) offline with @ada-x64 - comments are from an earlier version of this PR that I did not publish at the time but I'm leaving in for posterity!

PR #2461, opened 2023-12-01
Expression Editor Tweaks

ada-x64:
Based on #2437.

- Sidebar trapdoor sizing now stays sticky on any parameter change instead of just sticking on column name change.
- Expression editor has more breathing room for errors.
- Expression editor styling now takes the size of the unavailable buttons into account by using `visibility: none`.
- Update tests to use new playwright functionality in order to fix bugs.
- Added a test to ensure the trapdoor sizing remains at max width.

Before (In #2437):
Screenshot 2023-12-01 at 2 54 33 PM Screenshot 2023-12-01 at 2 54 20 PM

After (this PR):
Screenshot 2023-12-01 at 2 52 57 PM Screenshot 2023-12-01 at 2 51 56 PM


ada-x64:
Closing in favor of #2459

PR #2463, opened 2023-12-05
Update ipynb examples with new expression syntax

tomjakubowski:
The .ipynb examples were still using the list-based syntax for expressions.

I updated them to the dict-based syntax.


texodus (approved):

Issue #2464, opened 2023-12-05
Simple PerspectiveWidget errors restoring its own saved config


tomjakubowski:
### Steps to Reproduce:

1. Create perspective-python notebook
2. Add code snippet below to a cell
3. Execute the cell

```
w = PerspectiveWidget(
    {"a": [1]}
)

config = w.save()

PerspectiveWidget(df, **config)
```

### Expected Result:
Widget restores config

### Actual Result:

Error is raised:

<details>
<summary>(open to see screenshot)</summary>
<img width="785" alt="Screenshot 2023-12-04 at 6 23 53 PM" src="https://github.com/finos/perspective/assets/667901/a527429f-2add-46af-9e1d-de8870256d05">
</details>

```
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
Cell In[17], line 7
      1 w = PerspectiveWidget(
      2     {"a": [1]}
      3 )
      5 config = w.save()
----> 7 PerspectiveWidget(df, **config)

File /opt/homebrew/Caskroom/miniconda/base/envs/perspective-jupyterlab/lib/python3.11/site-packages/perspective/widget/widget.py:289, in PerspectiveWidget.__init__(self, data, index, limit, server, client, **kwargs)
    286         kwargs.update({"columns": config["columns"]})
    288 # Initialize the viewer
--> 289 super(PerspectiveWidget, self).__init__(**kwargs)
    291 # Handle messages from the the front end
    292 # `PerspectiveJupyterClient.send()`:
    293 # - The "data" value of the message should be a JSON-serialized string.
    294 # - Both `on_msg` and `@observe("value")` must be specified on the
    295 # handler for custom messages to be parsed by the Python widget.
    296 self.on_msg(self.handle_message)

File /opt/homebrew/Caskroom/miniconda/base/envs/perspective-jupyterlab/lib/python3.11/site-packages/ipywidgets/widgets/widget.py:503, in Widget.__init__(self, **kwargs)
    501 """Public constructor"""
    502 self._model_id = kwargs.pop('model_id', None)
--> 503 super().__init__(**kwargs)
    505 Widget._call_widget_constructed(self)
    506 self.open()

TypeError: PerspectiveViewer.__init__() got an unexpected keyword argument 'version'
```

### Environment:

perspective-python 2.7.1, python 3.11.5

### Additional Context:

I imagine we need to either ignore the `version` in the config, or else use it, but I'm lacking context at the moment.

PR #2465, opened 2023-12-05
Fix Jupyterlab Widget Save/Restore Symmetry

ada-x64:
Fixes https://github.com/finos/perspective/issues/2464

The version keyword should be ignored when restoring the widget, as it does not affect the widget itself. The version keyword is currently only used for migrations. Eventually, I would like to see automatic migrations based on version number, which will require a more sophisticated solution. This works for now :)


texodus:

texodus (approved):

Issue #2419, opened 2023-11-06
Ensure all python * imports pull from `__all__` exports


timkpaine:
We use a lot of `from .subpackage import *`, which is good but we really should be defining `__all__` in the files/folders from which we pull to avoid extra exports and to better aide with tools like intellisense and jupyter autoreload.

texodus:
> Don't ask for issues to be assigned to you if you're a first-time contributor.

[`CONTRIBUTING.md`](https://github.com/finos/perspective/blob/master/CONTRIBUTING.md#guidelines)

texodus:
Closing this as it's become a honeypot for low-effort & chatbot-generated PRs. We IMO really don't need maintenance tasks here anyway as they don't really have relevance to a user of Perspective looking at Issues.

PR #2477, opened 2023-12-12
issue #2419 Ensure all python * imports pull from __all__ exports

aciffone23:


python/perspective/perspective/handlers/__init__.py
```diff
@@ -11,18 +11,8 @@
 #  ┗━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┛
 
 from .common import PerspectiveHandlerBase
+from .aiohttp import PerspectiveAIOHTTPHandler
```
timkpaine, line 14: These should remain in try/excepts.

timkpaine (changes requested):

python/perspective/perspective/handlers/__init__.py
```diff
@@ -11,8 +11,20 @@
...
+    from .starlette import PerspectiveStarletteHandler
+except ImportError:
+    ...
+
+try:
+    from .tornado import PerspectiveTornadoHandler
+except ImportError:
+    ...
+
+__all__ = ["PerspectiveHandlerBase", "PerspectiveAIOHTTPHandler", "PerspectiveStarletteHandler", "PerspectiveTornadoHandler" ]
```
timkpaine, line 30: if the above imports fail, this will cause attribute errors when importing.

timkpaine (changes requested):

python/perspective/perspective/handlers/__init__.py
aciffone23: Ok I was thinking about something like this, should deal with that, this way if it only puts it to __all__ if it passes. What do you think?

try:
    from .aiohttp import PerspectiveAIOHTTPHandler
    __all__.append("PerspectiveAIOHTTPHandler")
except ImportError:

texodus:
Closing due to bitrot. Please feel free to reopen after the CLA process + review fixes.

PR #2495, opened 2024-01-10
Fix output scrolling in Jupyter when `perspective-python` is installed

### LINKED ISSUE #2494: JupyterLab: "Enable Scrolling for Outputs" does not work when perspective is installed
### Steps to Reproduce:
```bash
$ micromamba create -y -p /tmp/venv.perspective jupyterlab perspective -c conda-forge
+ jupyterlab 4.0.10
+ perspective 2.7.1
$ micromamba activate /tmp/venv.perspective
$ jupyter lab
```
In an empty notebook, run anything that generates reasonably long output like
```python
for i in range(500):
    print(i)
```
(note that `perspective` does not need to be imported) and then Right Click and select "Enable Scrolling for Outputs" in the context menu. No scroll bar appears.

### Expected Result:
Scrollbar should appear when using "Enable Scrolling for Outputs". [Here is someone else's video of what this looks like.](https://www.youtube.com/watch?v=U4usAUZCv_c)

### Actual Result:
No scroll bar appears when using "Enable Scrolling for Outputs" when `perspective` is installed in the environment.

### Environment:
I've observed that this bug happens on MacOS & Chrome, Safari; Ubuntu & Chrome, Firefox; and across various versions of `jupyterlab >= 3.5` and `perspective >= 2.0`, installed via any of pip / conda / mamba.

texodus:
Fixes #2494

### DIFF
--- a/packages/perspective-jupyterlab/src/less/index.less
+++ b/packages/perspective-jupyterlab/src/less/index.less
@@ -30,12 +30,6 @@ div.PSPContainer {
     height: 520px;
 }
 
-// override for small style
-.jp-CodeCell.jp-mod-outputsScrolled .jp-Cell-outputArea {
-    max-height: unset;
-    margin: auto;
-}
-
 // Widget height for Jupyter Notebook
 .jupyter-widgets-view div.PSPContainer {
     height: 520px;


timkpaine:
Are we sure this doesn't have unintended consequences? I thought that block was there to make sure perspective renders properly in an output and can be resized

PR #2469, opened 2023-12-07
Persist column settings across plugins

ada-x64:
First step in fixing https://github.com/finos/perspective/issues/1610

Based on #2443 

This PR moves the datagrid's integer and floating point precision out of the plugin settings and into a new perspective-viewer config field called `column_config`. This is communicated to the plugin by passing the viewer config to the plugin directly on restore.

CHANGES:
- Create a column style control abstraction and integrate a precision control for numeric types. 
- Add `column_config` to the viewer API.
- Update `plugin.restore` to accept both the plugin config and the total viewer config, allowing it to access `column_config`.
- Integrate the precision control for Datagrid.
- Add DEFAULT_CONFIG to perspective-viewer exports for use in tests.

TODO:
- ~~Integrate the new config settings in the datagrid.~~
- ~~Write migrations.~~
- ~~Write tests.~~
- Future PR: Use this to integrate X/Y Scatter Plot axis labelling with `column_config` precision status.

PR #2500, opened 2024-01-16
Add column settings render query to plugins

ada-x64:
Based on #2459 due to column settings sidebar refactors.

This PR introduces `plugin.can_render_column_styles()` in order to remove hardcoded string matching in the viewer.
```typescript
    /**
     * Given a column's grouping (determined by indexing it in `plugin.config_column_names`)
     * and its view type, determines whether or not to render column styles in the settings sidebar.
     */
    can_render_column_styles(
        view_type: perspective.Type,
        group: string
    ): boolean;
```
(from `rust/perspective-viewer/src/ts/plugin.ts`)

I believe these two parameters are all the possible requirements to determine whether or not a plugin can render column styles. If not, we can always update the API. We need this information to determine whether or not to show the edit button in the settings panel's column selector and on update to determine whether or not to keep the column settings open for that particular column, e.g. if aggregating and the view type changes to something the plugin cannot style.

A future PR will take the same parameters and produce data about which style controls a particular column can have. This will replace #2469, so you can consider this a Part 1 for those changes.


ada-x64:
Updated this to fix a bug where switching between datagrid date and datetime columns with the style tab open was not rerendering.

texodus:
Thanks for the PR!

Superseded by #2563

PR #2502, opened 2024-01-17
Add Docs Tests

ada-x64:
Currently on https://perspective.finos.org, there are broken examples on the home page. This occurs on X/Y scatter plots due to a failing migration.
Building and testing on main does not replicate the bug, as the website is running on v2.7.1 and [a fix has been added](https://github.com/finos/perspective/commit/3b648543e60805c98c0e5bb4fece03caa4341f6d) since that release.
This PR simply adds tests for the examples (one for each example, totaling 75) to ensure this doesn't happen again.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -706,7 +706,7 @@ jobs:
             # Fancy HTML reports #
             ######################
             - uses: actions/upload-artifact@v3
-              if: always()
+              if: failure()
               with:
                   name: playwright-report
                   path: tools/perspective-test/playwright-report/
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -702,6 +702,16 @@ jobs:
             - name: WebAssembly Test
               run: yarn test_js --ci
 
+            ######################
+            # Fancy HTML reports #
+            ######################
+            - uses: actions/upload-artifact@v3
+              if: always()
+              with:
+                  name: playwright-report
+                  path: tools/perspective-test/playwright-report/
+                  retention-days: 30
+
     #############################         ##############################
     #~~~~~~~~~~~~~~~~~~~~~~~~~~~#         #~~~~~~~~~~~~~~~~~~~~~~~~~~~~#
     #~~~~~~~~~~/########\~~~~~~~#         #~~~~~~~~~|##########|~~~~~~~#
--- a/docs/build.js
+++ b/docs/build.js
@@ -49,6 +49,11 @@ async function run_with_theme(page, is_dark = false) {
     await page.goto("http://localhost:8080/");
     await page.setContent(template(is_dark));
     await page.setViewport(DEFAULT_VIEWPORT);
+    await page.evaluate(async () => {
+        while (!window.__TEST_PERSPECTIVE_READY__) {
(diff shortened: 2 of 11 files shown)


ada-x64:
We've added [HTML reporting](https://playwright.dev/docs/test-reporters#html-reporter) to this PR. This allows us to download an artifact containing the results of the test.
File sizes might be an issue with this. See the [1.27GB playwright-report artifact](https://github.com/finos/perspective/actions/runs/7700622915?pr=2515) generated in #2515. This was with 75 errors and 3 retries, bloating the results to 225 trace files, 250 webms, and 180 pngs.

ada-x64:
Another update:
SVG snapshot comparisons exist, but were broken on master. There were several tests which were producing empty snapshots, and some which were broken because of bunk selectors. I've updated the SVG comparison code to account for this. In addition, the SVG comparison function now checks _all_ element children, so it should be suitable for general use. To reduce the number of changes in the PR, I decided not to implement this change.

tools/perspective-test/playwright.config.ts
```diff
@@ -157,7 +161,7 @@ export default defineConfig({
     forbidOnly: !!process.env.CI,
     retries: process.env.CI ? 2 : 0,
     quiet: true,
-    reporter: process.env.CI ? "github" : "dot",
+    reporter: process.env.CI ? [["github"], ["html"]] : [["dot"]],
```
texodus, line 164: ?

docs/build.js
```diff
@@ -49,6 +49,11 @@ async function run_with_theme(page, is_dark = false) {
     await page.goto("http://localhost:8080/");
     await page.setContent(template(is_dark));
     await page.setViewport(DEFAULT_VIEWPORT);
+    await page.evaluate(async () => {
+        while (!window.__TEST_PERSPECTIVE_READY__) {
```
texodus, line 53: If `__TEST_PERSPECTIVE_READY__` were a `Promise`, we could just `await` it here.

docs/build.js
```diff
@@ -73,13 +78,13 @@ async function run_with_theme(page, is_dark = false) {
...
         );
         console.log(JSON.stringify(new_config));
+
         await page.evaluate(async (config) => {
             const viewer = document.querySelector("perspective-viewer");
             await viewer.reset();
             await viewer.restore(config);
         }, new_config);
 
-        await page.waitForSelector("perspective-viewer:not([updating])");
```
texodus, line 82: What is the story behind this + [lines 52-56 in this file](https://github.com/finos/perspective/pull/2502/files#diff-4e0e40e9d4c5f482f9ba06ed27a0747b5ddf5fec750a3a34709fca59a7f2f3e0R52)?

.github/workflows/build.yml
```diff
@@ -702,6 +702,16 @@ jobs:
...
 
+            ######################
+            # Fancy HTML reports #
+            ######################
+            - uses: actions/upload-artifact@v3
+              if: failure()
+              with:
+                  name: playwright-report
+                  path: tools/perspective-test/playwright-report/
+                  retention-days: 30
```
texodus, line 713: I will keep an eye on the impact of these artifacts on our GitHub footprint!

texodus (approved):

PR #2511, opened 2024-01-26
docs: fix links on website and MD docs

maoo:
Credits to Tyler Male, for submitting these fixes via help@finos.org . Thank you!


texodus:
Thanks for the PR!

I've cleaned up the docs as part of the `v2.10.0` release - we don't need the repo churn of a PR, commits, `CHANGELOG.md` entry, etc. for minor documentation updates like these.

Also, moot point but neither of these corrections are technically correct either - `lerna` hasn't been a dependency in some time, and Finos does not answer questions about contributing to Perspective.

PR #2521, opened 2024-02-01
Move Column Settings from Plugin to Viewer with Column Settings Layout API

ada-x64:
Based on #2500; merges #2502 for SVG testing support

This PR aims to create a control-based API for column settings, bringing column-based settings from plugin configs to the viewer. This will allow us to support column settings which persist across plugins, e.g. numeric precision.

The design intends to allow plugin authors to use a number of viewer-defined "style controls." These style controls will affect column_config settings per row. This PR only changes the plugin API and maintains the existing style controls on the datagrid and x/y scatter chart. Following PRs will break up the current monolithic and type-based controls into more fine-grained and reusable controls.

Changes:
- [x] Create and integrate new plugin API
	- [x] Remove `plugin.pluginAttributes`
	- [x] Add `plugin.column_style_controls(type, group)`
	- [x] Move column configurations from plugins to viewer
		- [x] Layout API - returned from `plugin.column_style_controls`
		- [x] Data API - added to `plugin.restore(plugin_config, column_config)`
- [x] Integrate plugin API to the datagrid
	- [x] Tests pass
- [x] Integrate plugin API to the x/y scatter chart
	- [x] Tests pass 
- [x] Add migrations
- [x] Clean up cruft

Notes:
- Is modifying `restore` the best option for updating the plugins? We could also use `restyle`.


rust/perspective-viewer/test/js/migrate_viewer.spec.ts
```diff
@@ -342,44 +494,39 @@ test.describe("Migrate Viewer", () => {
...
+                const converted = convert(JSON.parse(JSON.stringify(old)), {
+                    replace_defaults: true,
+                    verbose: true,
+                });
                 expect(converted).toEqual(current);
             });
         }
     });
 
     test.describe("migrate", async () => {
```
ada-x64, line 506: Previously this tested that 1) the migration script did what was expected and that 2) the viewer restored to the expected value. This caused a lot of noise when debugging, so the test has been reconfigured to only test 2) since we test 1) above.

texodus:
Thanks for the PR!

Superseded by #2563

Issue #1610, opened 2021-11-14
Precision settings do not carry over to plots


ptomecek:
### Steps to Reproduce:
 1.  Create a table with data y = [0.001, 0.001, 0.003]
 2. Open a PerspectiveWidget with the table
 3. Set the precision of the column to 3 or 4 decimal places
 4. Switch the plugin to "Y-line", "Y-bar" or one of the others

### Expected Result:
The precision of the axis labels corresponds to the precision set in the tabular view (i.e. 3 or 4 places)

### Actual Result:
Axis labels remain set to the default 2 digit precision, with no way to change them. Thus all the labels display as "0.00" and the graph is not useful.

### Environment:
Perspective 1.0.3

### Additional Context:
Separately, I can't find a way to programmatically retrieve/set the precisions. Maybe this belongs as a new feature request...

ceikit:
any update on this?

texodus:
This was implemented in `v4.5.0` with the introduction of the `@perspective-dev/viewer-charts` package, which uses the full `Intl` formatting panel for all chart types, and uses these values for axes and tooltips:

<img width="505" height="435" alt="Image" src="https://github.com/user-attachments/assets/efa30699-9a63-4bde-8a3b-ffeb7a5b2204" />

PR #2522, opened 2024-02-02
Update axis tick labels to accomodate small ranges

ada-x64:
This PR updates d3fc chart axis tick values to accommodate ranges with small differences. It does so by calculating a precision based on the tick range's number of leading zeros. This is done with log10, but could probably be done with string manipulation as well.


texodus:
Before:

Screenshot 2024-02-15 at 8 09 17 AM

After:

Screenshot 2024-02-15 at 8 09 04 AM

texodus:
This PR does not fix #1610, but it does help address the underlying use case.

PR #2525, opened 2024-02-05
Bugfix: Settings panel width changes on split-by

ada-x64:
https://github.com/finos/perspective/assets/41482263/fe2e91e0-8212-472c-8777-f5355c559551

Before

https://github.com/finos/perspective/assets/41482263/9577e361-e6c7-47ed-83f5-e04e52fe7cff

After

This also fixes an alignment bug with the active column names. A better fix would probably be to remove the `:host *` rule which sets `line-height: 1.5`.

### DIFF
--- a/rust/perspective-viewer/src/rust/components/column_selector/active_column.rs
+++ b/rust/perspective-viewer/src/rust/components/column_selector/active_column.rs
@@ -404,7 +404,7 @@ impl Component for ActiveColumn {
 
                             <div class="column-selector-column-border">
 
-                                <TypeIcon ty={col_type} />
+                                <TypeIcon absolute=true ty={col_type} />
 
                                 if ctx.props().is_aggregated {
                                     <AggregateSelector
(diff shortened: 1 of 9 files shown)


rust/perspective-viewer/src/rust/components/column_selector/active_column.rs
```diff
@@ -404,7 +404,7 @@ impl Component for ActiveColumn {
 
                             <div class="column-selector-column-border">
 
-                                <TypeIcon ty={col_type} />
+                                <TypeIcon absolute=true ty={col_type} />
```
texodus, line 407: This is purely a style property, and if we can avoid expressing it in the Yew code, we should. In this case, `"absolute"` is a class name that sets a property in CSS already ... the `TypeIcon`s that use this property are all children of the `#settings_panel`, so this can be done in just that CSS rule.

texodus (commented):
Thanks for the PR!

There's still a small issue here - the width is off by a measly ***_1px_***!

You can test this by using the _longest column name_ as a test, in superstore this is `"Customer Name"`. This issue seems to predate the affected patch (it's broken in 2.7.1 as well), but since we're fixing column width there's no point in just fixing what was broken most recently :)

I've fixed this and addressed the other comment below in #2527 - closing in favor of this newer PR.

PR #2526, opened 2024-02-06
A few CSS fixes for regular-table in shadow DOM

tomjakubowski:
Moving datagrid's regular-table into shadow DOM caused some CSS regressions. These changes fix a few that we've spotted, particularly in the table header.

### DIFF
--- a/packages/perspective-viewer-datagrid/src/less/regular_table.less
+++ b/packages/perspective-viewer-datagrid/src/less/regular_table.less
@@ -81,9 +81,12 @@ perspective-viewer[settings],
     }
 }
 
-perspective-viewer regular-table table thead tr:last-child th {
-    border-bottom-width: 1px;
-    border-bottom-color: var(--inactive--border-color, #8b868045);
+perspective-viewer,
+:host-context(perspective-viewer) {
+    regular-table table thead tr:last-child th {
+        border-bottom-width: 1px;
+        border-bottom-color: var(--inactive--border-color, #8b868045);
+    }
 }
 
 .psp-sort-enabled:hover {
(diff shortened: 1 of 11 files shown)


packages/perspective-viewer-datagrid/src/less/regular_table.less
```diff
@@ -120,8 +123,11 @@ td:focus {
     outline-width: 1px;
 }
 
-perspective-viewer.dragging regular-table {
-    pointer-events: none;
+perspective-viewer.dragging,
+:host-context(perspective-viewer.dragging) {
```
tomjakubowski, line 127: I don't know how to test this rule - couldn't find anywhere where this rule is set, and various drag/drop operations I tried didn't seem to add the class name to the viewer.

rust/perspective-viewer/src/themes/vaporwave.less
```diff
@@ -93,45 +94,51 @@ perspective-string-column-style[theme="Vaporwave"] {
...
-            #b03e38,
-            #8c3a36,
-            #643633,
-            #07081d) !important;
+    --d3fc-negative--gradient: linear-gradient(
+        #f3d431,
+        #efb92d,
+        #ed9c25,
+        #eb7e20,
+        #e75d1e,
```
tomjakubowski, line 103: there were several formatting-only changes like this. I'm not sure how this happened - I have format with save enabled, in VSCode but it should be using the same prettier settings that the linter uses. And the linter accepts it in either style

rust/perspective-viewer/src/themes/monokai.less
```diff
@@ -64,7 +64,8 @@ perspective-string-column-style[theme="Monokai"] {
     --rt-neg-cell--color: #ff6188 !important;
     --rt--border-color: #444444;
 
-    // TODO: probably doesn't work with shadow DOM
+    // FIXME: broken in shadow DOM.  Can be fixed with a new custom property allowing empty header cell backgrounds
```
tomjakubowski, line 67: I took a stab at this, but I couldn't work out how to declare the "default" value for the CSS custom property so that the theme override's was more specific than it

texodus (approved):

Reviewed offline.

PR #2539, opened 2024-02-20
Number string formatting in datagrid

ada-x64:
Based on #2521 

This PR aims to create a universal number string formatter using the [Intl.NumberFormat() constructor](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/NumberFormat/NumberFormat) as the base datatype.
This PR creates a column settings control on the viewer and integrates it into the Datagrid. Further work will integrate this control into the D3FC plugins and the OpenLayers plugin.


texodus:
Thanks for the PR!

Superseded by #2563, closing

PR #2551, opened 2024-03-02
Fix get_hosted_table_names in Python client

tomjakubowski:
It was returning None. It now returns a future which resolves to the list of hosted table names.

### DIFF
--- a/python/perspective/perspective/tests/handlers/test_starlette_handler.py
+++ b/python/perspective/perspective/tests/handlers/test_starlette_handler.py
@@ -16,6 +16,7 @@
 import pytest
 import random
 
+from contextlib import asynccontextmanager
 from datetime import datetime
 
 from fastapi import FastAPI, WebSocket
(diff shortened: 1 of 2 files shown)


texodus (commented):
Thanks for the PR! 

Good catch! Can you add a test please?

python/perspective/perspective/tests/handlers/test_starlette_handler.py
```diff
@@ -64,6 +65,14 @@ async def websocket_client(self):
         """
         return await websocket(CLIENT, "/websocket")
 
+    @asynccontextmanager
+    async def managed_websocket_client(self):
```
tomjakubowski, line 69: I wrote my test first in the style of the others. But when I undid the fix to confirm it would make my test red, it also made the test hang: when the test is red, `await get_hosted_table_names()` throws, which exits the test before the `client.terminate()` call.

The other tests in this file would also hang if they broke. I can rewrite the other tests in this style if you want

tomjakubowski:
Added a test to the starlette integration tests

python/perspective/perspective/tests/handlers/test_starlette_handler.py
tomjakubowski: It was so easy I went ahead and made the change and split it out into a separate commit. Let me know if you'd like any changes

texodus (approved):

PR #2570, opened 2024-03-15
Fix compatibility with other MSVC versions by explicitly converting to string

timkpaine:
This is a really minor bug, but causes some annoyance during our build process and is the reason we have to explicitly set `PSP_GENERATOR` in a few places.

### DIFF
--- a/python/perspective/setup.py
+++ b/python/perspective/setup.py
@@ -176,7 +176,7 @@ class PSPBuild(build_ext):
                 "14.1": "Visual Studio 15 2017",
                 "14.2": "Visual Studio 16 2019",
                 "14.3": "Visual Studio 17 2022",
-            }.get(dm.get_build_version(), "Visual Studio 15 2017")
+            }.get(str(dm.get_build_version()), "Visual Studio 15 2017")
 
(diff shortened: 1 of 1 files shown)


texodus:
What is the bug? What is `PSP_GENERATOR`? How do I know not to revert this in the future?

timkpaine:
`PSP_GENERATOR` is the user-configureable override (similar to e.g. `PSP_DEBUG`) for the [cmake-generator](https://cmake.org/cmake/help/latest/manual/cmake-generators.7.html) used by the C++ build system. This doesn't matter too much on Mac and Linux, where you can likely use `makefiles` or `ninja` interchangeably, but on Windows you may want to override the default generator to be a certain version of Visual Studio for compatibility with python or other libraries. In the perspective codebase, we use this [here](https://github.com/finos/perspective/blob/5c8736982e1e5dcfd536b16040d38887a6210d0c/.github/workflows/build.yml#L1157C19-L1157C28) and its also used in e.g. conda-forge or when building from source.

texodus (approved):
Looks good!

PR #2571, opened 2024-03-15
Add better error logging to perspective manager message handling

timkpaine:
I recently had to debug a nasty issue which ended up being entirely due to some misuse of react. However, in the course of doing so, I noticed some server-side errors being swallowed. This PR adds a bit more verbose logging when client/server errors occur.

It will soon be obsolete due to the port to `rust`.


texodus:
Is this still a draft @timkpaine ?

timkpaine:
@texodus its good to go but technically all this code will go away with the client/server overhaul.

texodus:
Thanks for the PR!

I've cherry-picked this into #2623, closing.

PR #2574, opened 2024-03-18
Debug component

texodus:
Adds a new debug component to `<perspective-viewer>`, accessible via a tab on the settings panel, which shows the live configuration of the viewer in JSON in a code editor along-side the viewer panel.

https://github.com/finos/perspective/assets/60666/9ab07105-bb55-4068-9d03-635d728302e3


gtxca:
how can i disable this  JSON in a code editor component?

Reason being, I wouldn't want users to be able to run JSON queries in production.

Instead, they should use the UI to drag/drop or type the column name out using the viewer panel.

texodus:
This is not supported, nor is there any reason to do this - there is nothing the JSON editor can do that the UI cannot.

PR #2588, opened 2024-04-05
reenable minimum height for voila

timkpaine:
Tweaks and partially reverts https://github.com/finos/perspective/pull/2352

Currently, Perspective works great in voila except that it does not have any height::

Screenshot 2024-04-05 at 17 57 46

This PR gives it a minimum height:
Screenshot 2024-04-05 at 17 57 26

The sentiments in the original PR still hold, ~~I'm looking for an alternative way to tell if specifically in voilà.~~ but with https://github.com/voila-dashboards/voila/pull/1457 we can now disambiguate and apply only in voilà

### DIFF
--- a/packages/perspective-jupyterlab/src/less/index.less
+++ b/packages/perspective-jupyterlab/src/less/index.less
@@ -35,6 +35,11 @@ div.PSPContainer {
     height: 520px;
 }
 
+// Widget height for Voila
+body[data-voila="voila"] .jp-OutputArea-output div.PSPContainer {
+    min-height: 520px;
+}
+
 div.PSPContainer perspective-viewer[theme="Pro Light"] {
     --plugin--border: 1px solid #e0e0e0;
 }


timkpaine:
https://github.com/voila-dashboards/voila/pull/1457/files

timkpaine:
Not done in this PR, but we could use the same idea for `notebook`, which might help if/when we work in e.g. vs code: https://github.com/jupyter/notebook/blob/3b8e6736075279f80cd1e590c71a872dc5ad2c11/notebook/templates/notebooks.html#L15

texodus (approved):

I will merge this as-is, but I have to ask, what is the purpose of all of these custom height overrides?  Is it because we want `100%` to be the default in JupyterLab's "Output View" mode? Can we just change it to be e.g.:

```css
div.PSPContainer {
    height: 520px;
}

:not(.jp-NotebookPanel-notebook) div.PSPContainer {
    height: 100%;
}
```

I need to _ask_, because this feature has _no tests_, and I don't know how to build a voila app. Also, the PR has the word "reenable" in the _title_, as in, this feature has regressed _already_ ...

timkpaine:
In voila (and vs code and jupyterlab and Jupyter notebook for that matter) the output container expands to the size of its contents, so if you set `height: 100%` you'll end up with a height 0 div since perspective tries to expand to fill its container. Thus we need to give the container div some minimum height so that it displays.

You can run a voila app with `voila <notebook.ipynb>` and it will run locally, I don't think it really requires tests since for all these various platforms we just need to find the container element and set a height [next to the other container blocks](https://github.com/finos/perspective/blob/84605851230c3dd1cc58652b47babd04cc9fcc31/packages/perspective-jupyterlab/src/less/index.less#L28).

texodus:
I've cherry-picked this into #2623, closing in favor of this PR.

PR #2598, opened 2024-04-18
Fix gil unlocking

abalabin-bamfunds:


texodus:
Thanks for the PR!

I've re-opened this in #2608, which applies `lint` (`clang-format` from the [build failures](https://github.com/finos/perspective/actions/runs/8744857488/job/24110424008?pr=2598)) as well as signs the commits.

PR #2613, opened 2024-05-10
Upgrade node version

timkpaine:


texodus:
Thanks for the PR!

This doesn't build on CI or locally (for me). I'm sure you experienced an issue but I can't say what it was.

timkpaine:
I think it's only for node 22 which is not yet on GHA, I'm going to back out those changes and just leave the actions upgrades (which we should take care of as they're spitting out deprecation warnings).

PR #2618, opened 2024-05-16
Add region, column and row selection modes to Perspective Datagrid

texodus:
Adds row, column and region selection modes to Perspective Datagrid, as well as Export and Copy awareness of selection state.  Based on the [`regular-table` area selection example](https://finos.github.io/regular-table/block/?example=area_mouse_selection) from @JHawk.

The new selection modes are toggle-able (and mutually exclusive with) the "Edit"/"Read-Only" button in the toolbar.

https://github.com/finos/perspective/assets/60666/3c91d0bb-a45a-4c57-bc63-e46bfc1da5a0

TODO:
- [ ] Make compatible with `<perspective-workspace>` global filter hierarchical selection mode.
- [ ] Fix persistence (output `edit_mode` as part of `plugin_config`
- [ ] Tests


mhkeller:
What release is this functionality in? I tried it in a barebones 2.10.1 recreated from the website but I don't see it. Looks really useful!

PR #2621, opened 2024-05-17
Update toolchain, and update macro metavar syntax 

tomjakubowski:
The syntax change is needed for perspective to compile on newer toolchains

the relevant rust repo change is here:
https://github.com/rust-lang/rust/commit/027850569165192ce8912a985e9489e2ca5e8b65

I also updated some code to address new lints from the toolchain update, in a separate commit for readability


texodus:
Thanks for the PR!

I've cherry-picked this into #2623, closing.

PR #2625, opened 2024-05-21
Perspective Virtual API (Python)

texodus:
## Python Library

Building on #2615, this PR updates the Python library bindings. This PR does not include:
* `PerspectiveWidget` support #2658
* `pyodide` build support #2660 .
* Windows, OSX, or Python versions other than 2.9 (pending).

Breaking API changes:

- `date` and `datetime` are _not_ supported by the `Table` constructor anymore. These are not valid JSON values and JSON parsing is now language-agnostic. A subsequent PR will add the ability to read JSON strings directly, where this can be overcome as Perspective 2.x did - add a handler for these types to the default `json` module `dumps()` function.
- `pandas.DataFrame` is now handled by `pyarrow`, which needs to be installed for `DataFrame` type to work. Perspective implements no special inference/behavior for these, so it now inherits `pyarrow`'s semantics regarding column types, precision, etc.

## Rust Library

As a bonus which came together quite naturally, this PR also introduces a native Rust client and server for Perspective, which allows hosting the Perspective engine in e.g. [`axum`](), as well as the entirety of the Perspective API as previously implement for JavaScript and Python. The following new crates are introduced:

* `perspective` (native) has been repurposed as the root native-Rust crate, re-exporting the Rust Client and Server APIs as well as conveniences for linking the two with other libraries like [`tokio`]().
* `perspective-python` (native) contains the Python (pyO3) bindings, replacing the old C++ based `perspective-python`.
* `perspective-server` (native) wraps the C++ engine for both `perspective` and `perspective-python`.

An `axum` server example has also been added to the `examples/` directory.


python/perspective/perspective/nbextension/__init__.py
```diff
@@ -1,22 +0,0 @@
-#  ┏━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┓
```
timkpaine, line 1: This file is still required

rust/perspective-python/Cargo.toml
```diff
@@ -10,45 +10,51 @@
 #  ┃ of the [Apache License 2.0](https://www.apache.org/licenses/LICENSE-2.0). ┃
 #  ┗━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┛
 
-from enum import Enum
+[package]
```
timkpaine, line 13: This should probably remain in the `python` folder as it is still the python codebase, it can be linked to the rust code straightforwardly with a cargo workspace

texodus: There is no `python` folder on this branch.

timkpaine: thats fine then, just an organizational choice as `perspective-python` is not a useable standalone rust package. But its convenient to keep things together, and `packages` is already taken.

examples/python-tornado/server.py
```diff
@@ -81,5 +90,7 @@ def make_app():
     app.listen(8082)
     logging.critical("Listening on http://localhost:8080")
     loop = tornado.ioloop.IOLoop.current()
-    loop.call_later(0, init_table)
+    t = threading.Thread(target=update_thread)
```
timkpaine, line 93: is a thread actually necessary for this?

package.json
```diff
@@ -16,7 +16,7 @@
     "emscripten": "3.1.54",
     "llvm": "17.0.6",
     "engines": {
-        "node": ">=14.18.2 <22"
+        "node": ">=16"
```
timkpaine, line 19: no problem with this, just note that our build scripts don't work on node 22

rust/perspective-python/perspective/legacy.py
```diff
@@ -0,0 +1,142 @@
...
+
+def PerspectiveManager():
+    return create_sync_client(default_loop_cb)
+
+
+# def async_client():
+#     return create_async_client(default_loop_cb)
+
+
+def set_threadpool_size(num_cpus):
```
timkpaine, line 57: I think we can eliminate this, we've bumping a major version so no need for backwards compatible on this func

rust/perspective-python/pyproject.toml
```diff
@@ -11,19 +11,27 @@
...
-build-backend = "jupyter_packaging.build_api"
 
-[tool.black]
-line-length = 200
-target-version = ['py37', 'py38']
-include = '\.pyi?$'
+[project.optional-dependencies]
+dev = ["pytest>=7.4.3", "ruff>=0.4.6"]
+
+[tool.maturin]
```
timkpaine, line 30: we should just make sure to double check that the JS/JSON assets get bundled here now that we moved off the old packaging style, if Maturin isn't doing the correct thing by default and recursively including the package tree, we can easily explicitly include the files: https://www.maturin.rs/config#maturin-options

.github/actions/init/action.yaml
```diff
@@ -0,0 +1,133 @@
...
+                  ${{ runner.os }}-emsdk-
+
+        - name: Setup pip cache
+          uses: actions/cache@v4
+          if: ${{ inputs.skip_cache == 'false' }}
+          with:
+              path: |
+                  ~/.cache/pip
+                  ~/py_modules/
+              key: ${{ runner.os }}-pip-${{ hashFiles('**/setup.py') }}
```
timkpaine, line 57: this is gonna be `pyproject.toml` now

.github/actions/init/action.yaml
```diff
@@ -0,0 +1,133 @@
...
+          uses: hendrikmuhs/ccache-action@v1.2
+          if: ${{ inputs.skip_cache == 'false' }}
+          with:
+              key: ${{ github.job }}-${{ matrix.os }}
+
+        - name: Set up Python ${{ matrix.python-version }}
+          uses: actions/setup-python@v5
+          with:
+              python-version: ${{ matrix.python-version }}
+              cache: "pip"
```
timkpaine, line 85: you can also put the caching part here

.github/actions/init/action.yaml
```diff
@@ -0,0 +1,133 @@
...
+              pip3 install maturin
+
+        - name: Linux init steps
+          shell: bash
+          run: sudo node tools/perspective-scripts/install_tools.mjs
+          if: ${{ runner.os == 'Linux' }}
+
+        - name: Install CCache
+          shell: bash
+          run: sudo apt install -y ccache
```
timkpaine, line 133: this step is handled by `hendrikmuhs/ccache-action`

rust/perspective-python/perspective/legacy.py
texodus: The name of this file is `'legacy.py`

python/perspective/perspective/nbextension/__init__.py
texodus: I've moved the `PerspectiveWidget` upgrade to a dedicated branch #2658

examples/python-tornado/server.py
texodus: No, these updates are incomplete and have been used for testing. I've created a dedicated PR to track this and removed these changes #2659

package.json
texodus: Any idea why? I run my local build on node 22.

rust/perspective-python/pyproject.toml
texodus: #2658

.github/actions/init/action.yaml
texodus: It's now `requirements.txt` because https://github.com/pypa/pip/issues/12191

.github/actions/init/action.yaml
texodus: 👍 

I've implemented this in #2659

.github/actions/init/action.yaml
texodus: #2659

.github/actions/init/action.yaml
texodus: Implemented in #2659

Issue #2347, opened 2023-08-20
[UX] "enter" key on last cell in a datagrid does not trigger update


timkpaine:
### Steps to Reproduce:
Given an editable datagrid, the enter key will save the new cell contents and bring you to the cell immediately after it. However, at the last cell, it will not. 
![tmp](https://github.com/finos/perspective/assets/3105306/2ebffec3-c843-4f72-bd01-7cbc31fbe026)

### Expected Result:
This is a slightly weird UX, I think the expected UX would be that **enter on the last cell in a datagrid saves the new cell contents and removes cell focus**. I don't think other UX, either the current or e.g. going to the first cell of the next column, make sense.

### Actual Result:
Key has no effect until focus is moved.

### Environment:
Chrome 115.0.5790.170 (Official Build) (arm64) 
Perspective 2.5.0

PR #2650, opened 2024-06-27
Fix Enter key behavior on last cell in datagrid #2347

gitRasheed:
This PR addresses issue #2347.

Changes made:
- Modified the `keydownListener` function in `src/js/event_handlers/keydown/edit_keydown.js`
- Added logic to detect when the last cell is being edited and implemented behavior to save content and remove focus when Enter is pressed on the last cell

Testing:
- Tested with various datasets to ensure the new behavior works as expected
- Verified that existing functionality for non-last cells remains unchanged

### DIFF
--- a/packages/perspective-viewer-datagrid/src/js/event_handlers/keydown/edit_keydown.js
+++ b/packages/perspective-viewer-datagrid/src/js/event_handlers/keydown/edit_keydown.js
@@ -41,52 +41,51 @@ function getPos() {
     }
 }
 
-const moveSelection = lock(async function (
-    table,
-    selected_position_map,
-    active_cell,
-    dx,
-    dy
-) {
-    const meta = table.getMeta(active_cell);
-    const num_columns = this._column_paths.length;
-    const num_rows = this._num_rows;
-    const selected_position = selected_position_map.get(table);
-    if (!selected_position) {
-        return;
-    }
+const moveSelection = lock(
+    async function (table, selected_position_map, active_cell, dx, dy) {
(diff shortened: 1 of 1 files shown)


packages/perspective-viewer-datagrid/src/js/event_handlers/keydown/edit_keydown.js
```diff
@@ -82,6 +84,17 @@ const moveSelection = lock(async function (
     }
 });
 
+function isLastCell(model, table, target) {
+    const meta = table.getMeta(target);
+    return meta.y === model._num_rows - 1 && meta.x === model._column_paths.length - 1;
+}
+
+function saveCellContent(model, table, target) {
```
texodus, line 91: This is not needed, the cell contents are already saved on blur - and this can be an expensive call if the `View` is remote.

packages/perspective-viewer-datagrid/src/js/event_handlers/keydown/edit_keydown.js
```diff
@@ -82,6 +84,17 @@ const moveSelection = lock(async function (
     }
 });
 
+function isLastCell(model, table, target) {
+    const meta = table.getMeta(target);
+    return meta.y === model._num_rows - 1 && meta.x === model._column_paths.length - 1;
```
texodus: This checks both `x` _and_ `y`, e.g. the bottom-right corner - we only want to check the `y` parameter here, as per #2347

PR #2657, opened 2024-07-03
updated test_exception.py

KrishnaR7626:


timkpaine:
User @KrishnaR7626 reported for spam

texodus:
[We have a lint to protect against this :)](https://github.com/finos/perspective/blob/master/tools/perspective-scripts/lint_headers.mjs)

Issue #2221, opened 2023-05-11
Provide linux aarch64 wheels


0x26res:
## Feature Request

Provide wheels for linux aarch64 architecture.

This will allow users to run perspective on arm architecture (eg: aws graviton)

### Description of Problem:

It's not easy to install perspective from source, so providing a wheel will make it easier for linux aarch64 users.

### Potential Solutions:

I had a brief look at this MR that added wheels for macos arm64 https://github.com/finos/perspective/pull/2072

Another options would be to use [cibuildhweel](https://cibuildwheel.readthedocs.io/en/stable/)

Once this works we can also considering adding perspective to [arm64-python-wheel-tester](https://geoffreyblake.github.io/arm64-python-wheel-tester/)

zstsr:
I really need this

0x26res:
@texodus correct me if I'm wrong. Since version 3, perspective uses maturin to build the python binary wheels. That means you could technically cross compile on linux x86 to get the aarch64 wheel. But you won't be able to test the wheel. 

I think other libraries are using [arm64-python-wheel-tester](https://github.com/geoffreyblake/arm64-python-wheel-tester) in order to test that their library works on arm 64. 

One caveat is that the testing is limited to what you can call from a python virtual environment where only your library (python-perspective) is installed. Also the testing only consist of a python code snippet. See some [examples here](https://github.com/geoffreyblake/arm64-python-wheel-tester/blob/69d7052127372e8252fc045f76ae6ec6a089f3fa/test/packages.yaml) 

But I guess it's a start.


https://github.com/geoffreyblake/arm64-python-wheel-tester/blob/69d7052127372e8252fc045f76ae6ec6a089f3fa/test/packages.yaml#L4

texodus:
@0x26res I appreciate the suggestion but this approach is not going to work for our testing standards. Open source Perspective will not support this target until our open source CI platform supports it, which AFAIK is currently not the case for GitHub Actions, though [soon maybe?](https://github.blog/news-insights/product-news/arm64-on-github-actions-powering-faster-more-efficient-build-systems/).

We support exotic builds (including linux/aarch64) on [PROspective](https://prospective.co/) in the interim!

0x26res:
> though [soon maybe?](https://github.blog/news-insights/product-news/arm64-on-github-actions-powering-faster-more-efficient-build-systems/).

[...]

PR #2658, opened 2024-07-04
`PerspectiveWidget` support for new Python API

texodus:
Adds 

* Support for `PerspectiveWidget`. Lots of small changes applied, most notably removal of _client-only_ e.g. `client=True` mode, but some care has been taken to maintain the rudiments of 2.10.1's user-facing API, at least so far. I would like to consider removing the `ViewConfig` traitlets as well in the future, as `restore()` is the preferred way to do this from both a usability and performance pov. This component needs _a lot_ more cleanup, this PR just updates the internal API relevant parts.
* Proper packaging of `perspective-jupyterlab` client extension assets. These now use `maturin`'s data directory support to package the assets, which are built using `jupyter-packaging`'s `labextension install` webpack builder.
* Manylinux builds for linux.
* `aarch64` builds for OSX. I've also added support for this architecture on Linux #2221, but it is not yet enabled on CI (and hence will not distribute), as I am not sure how we can test it just yet. Suggestions welcome.
* A rewrite of the project's Github Actions CI script which is substantially faster. The first part of this work was merged in #2659, but as we did not yet have support for `PerspectiveWidget`, these steps were omitted from this PR and included here.
* `perspective-server` and `perspective-client` have been reorganized to allow a new variant of the `Session` type introduced in #2625, `ProxySession`, which allows an existing `Client` instance to itself act as a `Session` for another `Client`. This change was necessary to support `PerspectiveWidget`, as the API was designed for (and remains convenient for) working on `Table` instances directly, for which the new API prevents access to the originating `Server`.
* Removed a number of unused build automations scripts and removed some crufty code generation, such as `_version.py` in the Python package.
* Fix `perspective-jupyterlab` and `perspective-python` dependency ranges to allow JupyterLab 4. This was already technically supported but will no longer warn of version mismatches.


tools/perspective-test/src/js/results.tar.gz
timkpaine, line 1: should this be omitted?

texodus: This file is the HTML diff test baseline archive, it is required to run the test suite and must be checked in when the integration tests are modified. This is [addressed in the Developer Guide](https://perspective.finos.org/docs/development/#javascript) (though this document is out of date and refers to `results.json` from before we ported to `playwright`).

Issue #2632, opened 2024-06-03
Python 3.12 wheels


0x26res:
### Description of Problem:

I can't install python-perspective on with python 3.12 (there's no wheel so I have to build it from source).

### Potential Solutions:

Upload wheels for python 3.12. I believe some work got done to get partial python 3.12 support, but it's not fully done yet. See https://github.com/finos/perspective/pull/2458

texodus:
#2625 migrates `perspective-python` to the [Python Stable ABI](https://docs.python.org/3/c-api/stable.html#stable-abi) which will support 3.12 and all future 3.x Python versions.

Smrithishera:
This issue of installing python-perspective on Python 3.12 can be addressed by uploading pre-built wheels specifically designed for Python 3.12 to the Python Package Index (PyPI). 
(You can use https://pypi.org/project/wheel/ for reference) providing these wheels, users will benefit from a smoother installation process and improved compatibility with the latest Python version.

Advantages:
Simplified Installation: Users will no longer need to build from source, streamlining the installation process and reducing potential errors.
Enhanced Compatibility: Pre-built wheels ensure compatibility with Python 3.12, enabling users to leverage the latest features and improvements.

Implementation Details:
To upload Python 3.12 wheels to PyPI, the maintainers can follow these steps:

Build wheels for python-perspective specifically targeting Python 3.12.
Upload the generated wheels to PyPI using the twine package or another suitable tool.
Update the package metadata on PyPI to reflect the availability of Python 3.12 wheels.

anoopg512:
If sticking with Python 3.12 is mandatory, using Docker or Conda can provide a more flexible environment to handle such dependencies.

0xg0nz0:
As part of this enhancement it would be very helpful to include aarch64 Linux as well as MacOS ARM64 wheels; I think I am not alone running Linux on MacOS X via Docker.

Note that I was able to build from source for Python 3.12, Ubuntu 24.04 on aarch64 by installing `libboost-all-dev` so there is a workaround for now.

Ubaidullah2023:
yes
**Steps to Include aarch64 Linux and MacOS ARM64 Wheels:**

**1. Set Up Cross-Compilation Environments:**

Use cross-compilation tools to build the wheels for different architectures without requiring physical hardware for each architecture.

**2. Build aarch64 Linux Wheels:**

[...]

PR #2659, opened 2024-07-04
Docs and examples update for new API

texodus:
* Updates the `docs` project to build against the new API
* Updates the `examples` projects as well, removing deprecated examples.
* Updates the `perspective-python` API for feedback from @timkpaine
* Adds Windows and OSX build support to CI

PR #2661, opened 2024-07-04
Use ABI3 for `perspective-python` shared library

texodus:
Adds `abi3-py39` build target to `perspective-python`


rust/perspective-python/src/client/client_sync.rs
```diff
@@ -40,7 +40,7 @@ pub struct PySyncClient(PyClient);
 #[pymethods]
 impl PySyncClient {
     #[new]
-    pub fn new(callback: Py<PyFunction>) -> PyResult<Self> {
+    pub fn new(callback: Py<PyAny>) -> PyResult<Self> {
```
timkpaine, line 43: TODO check for call

timkpaine (changes requested):
Tests/any other places should remove spurious lambdas as in https://github.com/finos/perspective/pull/2667

rust/perspective-python/src/client/client_sync.rs
texodus: The previous version of Perspective did not do this, and Python has adequate error handling for this already.

timkpaine: Its quite bit uglier now though:
```raw
perspective.PerspectivePyError: External error: PyErr { type: <class 'TypeError'>, value: TypeError("'<Type>' object is not callable"), traceback: Some(<traceback object at 0x15c1e5780>) }
```

texodus:
Fixes #2632

rust/perspective-python/src/client/client_sync.rs
texodus: I've fixed this error message on #2659

PR #2663, opened 2024-07-05
[proposal] Rename rust-to-python exports to be cleaner, normalize exception names

timkpaine:
I think `PySyncClient` et al are a bit weird in python land, so I propose normalizing on:
- `Client`/`Table`/etc for synchronous
- `ClientAsync` / `TableAsync`/etc for asynchronous 
This is pretty trivial to do it seems, so little downside

Separately, I propose using the following names:
- `PerspectiveError`: Error in the C++/Rust perspective engine
- `PerspectivePythonError`: error in just the python slice of the codebase

Ideally these should inherit from a common base class (maybe `PerspectivePythonError` can inherit from `PerspectiveError`, but it would be nice to very clearly disambiguate engine errors from python land errors (e.g. web server or other utilities' errors)

### DIFF
--- a/rust/perspective-python/perspective/__init__.py
+++ b/rust/perspective-python/perspective/__init__.py
@@ -12,9 +12,9 @@
 
 __version__ = "2.10.1"
 __all__ = [
-    "PySyncClient",
+    "Client",
+    "PerspectivePythonError",
     "PerspectiveError",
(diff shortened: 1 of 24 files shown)


texodus:
This has been subsumed by #2659

PR #2664, opened 2024-07-05
[wip] working out cleaner python handlers and examples

timkpaine:
Some experiments with the tornado/fastapi/aiohttp handlers.
Not suitable for merging yet


texodus:
This has been subsumed by #2659

PR #2665, opened 2024-07-05
Put labextension/nbextension files in correct folder

timkpaine:
This does not fix the labextension/nbextension, but its been annoying already building and having these files end up in what is now a brand new folder 😅


texodus:
Thanks for the PR, this is bitrotted - based on a commit which does not have a working widget, it is irrelevant which directory the assets land. 

`PerspectiveWidget` support has been merged in #2658

PR #2666, opened 2024-07-05
Switch to ruff, fix ruff errors

timkpaine:
Fixes the python listing scripts, looks like we decided to move to ruff so this updates the JS scripts to use ruff and then fixes a bunch of ruff errors, almost all of which were little things in the tests like reusing method names, unnecessary imports, things we've commented out, etc.

### DIFF
--- a/package.json
+++ b/package.json
@@ -118,7 +118,9 @@
         "prepush": "npm run lint",
         "prepare": "husky install",
         "lint": "node tools/perspective-scripts/lint.mjs",
+        "lint_python": "node tools/perspective-scripts/lint_python.mjs",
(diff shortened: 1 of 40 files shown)


texodus:
Thanks for the PR but this is not how we want lint asserted. Closing due to bitrot, lack of CI coverage, dispatch to removed build scripts (`lint_python` and `fix_python` - 3.x has removed _all_ of these custom per-language build frontends in favor of a centralized build command structure).

Please also try to separate rules changes from rules applications in the future. I will not prevent a merge for lacking this, but neglecting to do this virtually guarantees the change cannot be rebased and will bitrot.

PR #2667, opened 2024-07-05
Let callback in python client be any callable

timkpaine:
Tests updated now that lambdas are spurious. We should probably throw a check `if hasattr("__call__")` somewhere in the mix rather than fail on first client callback attempt.

### DIFF
--- a/rust/perspective-python/src/client/client_sync.rs
+++ b/rust/perspective-python/src/client/client_sync.rs
@@ -40,7 +40,7 @@ pub struct PySyncClient(PyClient);
 #[pymethods]
 impl PySyncClient {
     #[new]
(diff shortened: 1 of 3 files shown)


rust/perspective-python/src/client/client_sync.rs
```diff
@@ -40,7 +40,7 @@ pub struct PySyncClient(PyClient);
 #[pymethods]
 impl PySyncClient {
     #[new]
-    pub fn new(callback: Py<PyFunction>) -> PyResult<Self> {
+    pub fn new(callback: Py<PyAny>) -> PyResult<Self> {
         let client = PyClient::new(callback);
```
timkpaine, line 44: TODO check if hasattr callback "__call__" and throw exception if not

texodus:
This is already done in #2661

PR #2674, opened 2024-07-15
Cleanup top level import, remove unused threadpool setter

timkpaine:


rust/perspective-python/perspective/__init__.py
```diff
@@ -15,32 +15,20 @@
     "PerspectiveError",
     "PerspectiveWidget",
     "PerspectiveViewer",
-    "PerspectiveTornadoHandler",
```
timkpaine, line 18: do not require tornado, aand most of these are not defined here, and server/client are missing

texodus:
Thanks for the PR, but this is not the right solution - `set_threadpool_size()` needs to be implemented for the internal Arrow threadpool as per 2.10.1

PR #2679, opened 2024-07-16
Add Pyodide support and split Rust/C++ builds for Python wheel

timbess:
This PR does the following:
* Adds Pyodide support by splitting Rust & C++ and linking them within Python (linking Rust and C++ directly seemed to lead to weird errors in indirect function call offsets at runtime).
* Unifies the Python and JS binding to use the same translation unit (binding_api.cpp).
* Refactors the binding API to be in terms of packed C structs over the manual malloc + pointer arithmetic.


cpp/perspective/CMakeLists.txt
```diff
@@ -131,7 +131,7 @@ elseif(PSP_PYTHON_BUILD)
     if(NOT DEFINED PSP_PYTHON_VERSION)
         set(PSP_PYTHON_VERSION 3.10)
     endif()
-    if($ENV{PYODIDE})
+    if(PSP_WASM_BUILD)
         set(PSP_PYODIDE 1)
```
timkpaine, line 135: `PSP_PYODIDE` should be an `option` not a random variable, as in https://github.com/finos/perspective/pull/2677

cpp/perspective/CMakeLists.txt
```diff
@@ -679,6 +688,7 @@ elseif(PSP_CPP_BUILD OR PSP_PYTHON_BUILD)
             target_compile_options(psp PRIVATE -fvisibility=hidden)
             # target_compile_options(psppy PRIVATE -fvisibility=hidden)
         else()
+            target_compile_options(psp PRIVATE -fvisibility=hidden)
```
timkpaine, line 691: why is this flag set for `MANYLINUX`? can you identify what cases this `else()` targets?

cpp/perspective/CMakeLists.txt
```diff
@@ -644,25 +653,25 @@ elseif(PSP_CPP_BUILD OR PSP_PYTHON_BUILD)
...
         # #######################
         # Python extra targets #
         # #######################
-        add_library(psp STATIC ${PYTHON_SOURCE_FILES})
+        if(PSP_WASM_BUILD)
+            set(CMAKE_EXECUTABLE_SUFFIX ".wasm")
+            set(CMAKE_EXE_LINKER_FLAGS "${PSP_WASM_LINKER_FLAGS} --pre-js \"${PSP_CPP_SRC}/env.js\" ")
+            add_library(psp ${PYTHON_SOURCE_FILES})
+            add_executable(psppy ${PSP_CPP_SRC}/src/cpp/binding_api.cpp)
+            target_compile_definitions(psppy PRIVATE PSP_ENABLE_PYTHON=1 PSP_ENABLE_WASM=1)
```
timkpaine, line 665: duplicate lines here

cpp/perspective/CMakeLists.txt
```diff
@@ -131,7 +131,7 @@ elseif(PSP_PYTHON_BUILD)
     if(NOT DEFINED PSP_PYTHON_VERSION)
```
timkpaine, line 131: This file seems to be getting messier and messier, and this PR seems to exacerbate that by making the control flow even more complicated

timkpaine:
Ok I think https://github.com/finos/perspective/pull/2677 will be done soon, I really think if we are making any CMake changes we should do on top of #2677, the file is already clearly in a poor state (e.g. multiple sets of clobbering blocks like [this](https://github.com/finos/perspective/blob/fae909221e09a88ca347ff8bb935f3d8fa90692e/cpp/perspective/CMakeLists.txt#L58) / [this](https://github.com/finos/perspective/blob/fae909221e09a88ca347ff8bb935f3d8fa90692e/cpp/perspective/CMakeLists.txt#L571)) which has already caused issues, e.g. the most recent re-enabling of windows.

cpp/perspective/CMakeLists.txt
timbess: I'm not sure it makes sense to have pyodide references at all anymore. We're really just building an emscripten dynamic library that can be dynamically loaded by any compatible emscripten compiled Python implementation. I think we should just delete references to pyodide all together.

cpp/perspective/CMakeLists.txt
timbess: This was for emscripten, I can add another `elseif` for manylinux explicitly. Good catch 👍

cpp/perspective/CMakeLists.txt
timbess: Agreed this file could use a makeover. I'm not sure where `PSP_WASM_BUILD` and `PSP_PYTHON_BUILD` are being conflated though. This is my understanding, correct me if I'm wrong:
* `PSP_WASM_BUILD` only is a JS target.
* `PSP_PYTHON_BUILD` only is a Python native dylib.
* `PSP_WASM_BUILD` and `PSP_PYTHON_BUILD` together means we're building an emscripten SIDE_MODULE loadable by any compatible Python implementation.

If there's somewhere this is being handled incorrectly, we definitely need to fix it. I'll fix the `MANYLINUX` thing when I'm at my computer again, thanks for catching that 👍

cpp/perspective/CMakeLists.txt
timkpaine: sounds good to me, in #2677 i separate out emscripten in the top platform block, we could just rely on that

cpp/perspective/CMakeLists.txt
timkpaine: depending on the level of overlap, we could just have 3 completely separate paths. I dont think theres much overlap between python and pyodide, since we’re not going to support building from pyodide on e.g. windows. Emscripten and pyodide of some overlapping flags, but if theres enough separation we can literally just do:

```raw
if(PSP_WASM_BUILD)
set(PSP_PYTHON_BUILD OFF)
set(PSP_PYODIDE_BUILD OFF)
…
elseif(PSP_PYODIDE_BUILD)
…
elseif(PSP_PYTHON_BUILD)
…
endif()
```

texodus:
Fixed in #2707

PR #2690, opened 2024-07-24
fix: Doctype in html export

aszenz:
Firefox throws a warning 

> Bogus doctype.Quirky doctype. Expected “<!DOCTYPE html>”.


texodus (approved):

Normally, I'd ask for a test for something like this, especially because this is a [noted prerequisite for _official_ Firefox support](https://github.com/finos/perspective/issues/2693), but this is obviously invalid HTML and this is the only occurring instance in the project.

PR #2700, opened 2024-08-02
Stringify Arrow lists

sinistersnare:
Today, Perspective fails to load when given an arrow with a `ListArray`. This PR provides minimal support for this data type by stringifying the list so it can be rendered as a string. Only a few list-element data-types are currently supported by this PR, but more can be added by this scaffolding.


texodus (approved):

PR #2705, opened 2024-08-06
Fix Windows JupyterLab packaging

texodus:
Fixes `perspective-python` prebuilt JupyterLab extension packaging for Windows, where some Node.js' `fs` module functions do not behave as they do on Linux/OSX.

### DIFF
--- a/tools/perspective-scripts/repack_wheel.mjs
+++ b/tools/perspective-scripts/repack_wheel.mjs
@@ -12,21 +12,21 @@
 
 import { execSync } from "node:child_process";
 import * as fs from "node:fs";
-import * as path from "node:path";
 import pkg from "../../package.json" assert { type: "json" };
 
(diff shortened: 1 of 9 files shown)


tools/perspective-scripts/repack_wheel.mjs
```diff
@@ -12,21 +12,21 @@
...
 
+const dest = `${pkg_name}/perspective_python-${version}.data`;
+const src = `rust/perspective-python/perspective_python-${version}.data`;
+fs.cpSync(src, dest, {
+    recursive: true,
+});
+
 execSync(`wheel pack ${pkg_name}`);
+
+// what kind of customers
```
timkpaine, line 32: ?

texodus: _what kind of customers?!_

PR #2713, opened 2024-08-16
Better Timestamp Parsing

sinistersnare:
Hello!

This PR improves the parsing of timestamp strings in Perspective! It also improves our Timezone parsing logic to better support strange timezones (like Nepal, Myanmar, or Mars). This PR also adds a few tests, so I hope that you have a good day and can sleep more restfully with this in mind.

Thanks!


texodus (approved):

Outstanding test coverage!

PR #2736, opened 2024-09-03
Added new decimal decoration for chart tooltips 

### LINKED ISSUE #1006: More readable styling for chart tooltips
### Description of Problem:

When looking at tooltips on charts, the decimal point between floats are unclear and hard to read at a glance:

Screen Shot 2020-04-06 at 6 35 53 PM

### Potential Solutions:

Tooltips should be styled with a clear differentiation/highlight on the number, or the bold styling should be removed as it seems to (at least for me) hinder quick at-a-glance readability.

Av1sF:
This change resolves #1006 by adding simple padding and bold decoration to presented decimal numbers. 
It is a non-breaking feature, that allows easy visibility of a decimal place at first glance that works with any theme or colour. 

This is with decoration: 
Screenshot 2024-09-03 170207
This is without: 
undecoratedScreenshot 2024-09-03 170246

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,16 @@
+# [v3.0.2](https://github.com/finos/perspective/releases/tag/v3.0.2)
+
+_30 August 2024_ ([Full changelog](https://github.com/finos/perspective/compare/v3.0.1...v3.0.2))
+
+Fixes
+
+- Fix link error in emscripten wheel [#2726](https://github.com/finos/perspective/pull/2726)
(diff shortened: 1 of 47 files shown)


texodus:
Closing:

1) Non-substantive & subjective. While there is an issue for this, the issue is also non-substantive and outdated (reported from before Perspective used monospace fonts by default). We are not looking for contributions that change trivial elements like font sizes, colors and padding.
2) PR commit history is broken (rewritten) and unmerge-able anyway.

PR #2739, opened 2024-09-06
Emscripten wheel fixes

tomjakubowski:
- Fix Emscripten wheel build to better support use of Emscripten -fexceptions
- Add pyodide virtual package, selectable from `pnpm run setup`, which builds perspective-python with Pyodide environment flags set
- Add pyodide-pytest smoke tests for the Emscripten wheel, which run pytests in pyodide in Chrome using playwright, locally and in CI
- Add convenient vscode snippets for copyright notice in `#` and `//` flavors
- Add build helper functions to scripts for various important build paths, directories, and configs
- Replace old-style `import` asserts for reading `package.json` with new helpers, which improves forwards Node compatibility


tools/perspective-scripts/workspace.mjs
```diff
@@ -0,0 +1,53 @@
...
+    return workspaceRoot;
+}
+export function getRustTargetDir() {
+    return rustTargetDir;
+}
+export function getRustWheelsDir() {
+    return rustWheelsDir;
+}
+export function getEmscriptenWheelPath() {
+    const pspVersion = getWorkspacePackageJson().version;
```
tomjakubowski, line 41: I'm not sure, is the workspace package.json an OK place to get the version of the emscripten wheel? Or should this read from rust/perspective-python/Cargo.toml? If they're meant to be kept in sync, it shouldn't matter

tools/perspective-scripts/workspace.mjs
```diff
@@ -0,0 +1,53 @@
...
+    const wheeljunk = "cp39-abi3-emscripten_3_1_58_wasm32";
```
tomjakubowski, line 42: not sure if there's a good way to compute this (from the pyproject.toml? or Cargo.toml?)

tools/perspective-scripts/install_pyodide.mjs
```diff
@@ -0,0 +1,57 @@
...
+        console.log(`Removing ${tarball}...`);
+        fs.rmSync(tarball);
+        // assert presence of a known file
+        assert(fs.existsSync(pyodideLock), `${pyodideLock} not found`);
+        console.log(`Extracted to ${downloadDir}`);
+        fs.writeFileSync(buildStamp, ""); // prevent re-download/extract
+    }
+}
+
+downloadPyodide();
```
tomjakubowski, line 57: was looking for an equivalent to `if __name__ == "__main__"` but the closest I could find appears to be undergoing active bikeshedding https://github.com/nodejs/node/issues/49440

`require.main` doesn't work in Node esm

tools/perspective-scripts/repack_wheel.mjs
```diff
@@ -12,7 +12,11 @@
 
 import { execSync } from "node:child_process";
 import * as fs from "node:fs";
-import pkg from "../../package.json" assert { type: "json" };
+
+const __dirname = url.fileURLToPath(new URL(".", import.meta.url)).slice(0, -1);
+const pkg = JSON.parse(
```
tomjakubowski, line 17: need to update these to use the new getWorkspacePackageJson() function

tools/perspective-scripts/workspace.mjs
tomjakubowski: huh, interesting. maybe should do like this, and read it from perspective-python's package.json https://github.com/finos/perspective/pull/2739/files#diff-fbc295126e0b0b6fdc48b0a7f73d15b69d7cbb25140f65f0453200c1b341f809L28-R32

texodus: We update all `package.json` files at once, so the "local" one is the right place to fethc it generally, except the scripts in `perspective-scripts` don't live in a module, so the workspace `package.json` is the only option.

tools/perspective-scripts/workspace.mjs
texodus: IMO we get this by listing any `.whl` target with "emscripten" in the triple name - and if there are two (or more), error with a message instructing the dev to run `pnpm run clean`.

tools/perspective-scripts/install_pyodide.mjs
texodus: There's no easy way to do this, but every script in this folder should be considered a "main" script, that's what the folder is for. There are a handful of exceptions already e.g. `sh.mjs`.

texodus (commented):

This is a lot of commits for a cleanup PR but I will let it slide this time because the commits are descriptive and additive! (e.g. no "WIP reverting previous commit" or the like).

texodus (approved):

rust/perspective-python/package.json
```diff
@@ -12,9 +12,10 @@
     "scripts": {
         "build": "node build.mjs",
         "clean": "rimraf perspective.data && rimraf dist && rimraf build",
-        "test": "python -m pytest perspective/tests"
+        "test": "node test.mjs"
     },
     "devDependencies": {
+        "@finos/perspective-scripts": "workspace:*",
```
texodus, line 18: Posthumous review update -

Packages in the `tools/` directory are private to the source repository and they can't be included as dependencies in the published packages, even if they are just dev dependencies, because these tools ultimate are intended to depend on Perspective also and you end up with circular imports.

tomjakubowski: Makes sense. I didn't know that we don't publish `@finos/perspective-scripts` to NPM

PR #2747, opened 2024-09-12
<TAB> key press inserts '\t' character in code editor as indentation

### LINKED ISSUE #2703: Difficult to indent code in exprtk editor because <TAB> key blurs the input
### Steps to Reproduce:

1. Create a new computed column
2. Type a multi-line expression in the editor
3. Attempt to indent it with `<TAB>`

### Expected Result:

Indentation is added.

### Actual Result:

`<TAB>` acts like in an ordinary web form, where it blurs focus from the editor and focuses the next available form field (in this case, the "RESET" button)

Screenshot 2024-08-05 at 12 57 48 PM

While that behavior makes sense in many form contexts, I think given the form field's use case as code editor, it would be better to make <TAB> apply indentation in the editor - whether by inserting literal <TAB> characters or soft-tabs.

### Environment:
Chrome 127, macOS

onesimus-wiafe:
Hi all,
This PR fixes #2703.
I updated the function `on_keydown` to ignore the default behavior of the <TAB> key.

### DIFF
--- a/rust/perspective-viewer/src/rust/components/form/code_editor.rs
+++ b/rust/perspective-viewer/src/rust/components/form/code_editor.rs
@@ -74,6 +74,25 @@ fn on_keydown(event: KeyboardEvent, deps: &(UseStateSetter<u32>, Callback<()>)) 
     if event.shift_key() && event.key_code() == 13 {
         event.prevent_default();
         deps.1.emit(())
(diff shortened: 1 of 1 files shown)


texodus:
This PR does not compile and could really benefit from an integration test!

Issue #2575, opened 2024-03-20
Tooltip in x/y line chart shows `-` and `NaN`


mhkeller:
## Bug Report

When creating a multi-series line chart, the values display correctly in the chart but y value shows up as `NaN` in a tooltip and x value is repeated and shows up as `-`.

### Steps to Reproduce:

1. Load this dummy data of student exams in different subjects and periods: https://gist.github.com/mhkeller/8831f579f814af2a7415e043457d4e56
2. Select the X/Y line visualization
3. Group by `exam_month`, split by `exam_subject`, set the x axis to `exam_month` (any) and the y axis to `exam_score` (avg)

### Expected Result:

I'd expect to see the proper x and y-value in the tooltip.

### Actual Result:

I see `NaN` in the tooltip for `exam_score` and `exam_month` shows up twice: once as the proper value and once as `-`.

### Environment:

2.9.0 reproduced at https://perspective.finos.org/block/?example=file

### Additional Context:
![Screenshot 2024-03-20 at 13 38 34](https://github.com/finos/perspective/assets/498744/fc9c0ef8-d854-433c-aa89-f36f875c3e3b)

texodus:
Thanks for the report - I'm pretty sure this is the same behavior described in #2254, let me take a look and I will update

mhkeller:
Just wanted to check in on this and offer to take a look if you had some pointers on what code might be affecting this. I'm doing a data training next month and wanted to show this as an example.

texodus:
@mhkeller I appreciate the enthusiasm but finding the issue is 90% of the work.

The `master` branch of this repo has already been updated for the 3.0 release and is _not_ backwards compatible, so any fix for this will need to wait at least as long as this release, but I will make sure this fix is in by then.

mhkeller:
Understood – and if I'm able to offer anything more than enthusiasm, I'll add a comment here. I'll start by getting a dev environment set up and see if there's any insight I can provide.

wakek:
> @mhkeller I appreciate the enthusiasm but finding the issue is 90% of the work.
> 
> The `master` branch of this repo has already been updated for the 3.0 release and is _not_ backwards compatible, so any fix for this will need to wait at least as long as this release, but I will make sure this fix is in by then.

Hi @texodus,

I am able to reproduce this bug in version 3.0.3. I note that it has been added to the 3.0 milestone, however, I don’t see any in-progress draft PRs regarding this issue, so I’d like to contribute a fix for it.

[...]

Issue #2254, opened 2023-06-12
Y Bar: split-by value labels are broken when multiple aggregate values are plotted


tomjakubowski:
### Steps to Reproduce:

I made a small example which is easy to cause the bug in. **Skip to step 4** using this export [mls_club_transfer_profit.html.zip](https://github.com/finos/perspective/files/11717423/mls_club_transfer_profit.html.zip), or go here while the site is still up: https://mls-stats.netlify.app/mls_club_transfer_profit.html

1. Create a Y Bar chart
2. "Group by" some column
3. Choose a couple aggregate columns to plot on the bar chart
4. Hovering over a bar shows the correct aggregate values ✅<br><img width="1502" alt="Screen Shot 2023-06-11 at 9 11 04 PM" src="https://github.com/finos/perspective/assets/667901/fcf5d18f-ef97-48b1-a416-d1c5c3f6626b">
5. "Split by" another column (in this example, "name" or "season" colum
6. Hovering over a bar shows NaN aggregate values ❌ <img width="1502" alt="Screen Shot 2023-06-11 at 9 13 51 PM" src="https://github.com/finos/perspective/assets/667901/c1c1c5da-8f4c-481f-9a22-3ddefaea5075">

### Expected Result:

When plotting only one of the Spend/Earning columns, hovering shows the value correctly. I would expect to see the same values when plotting both columns

<img width="1502" alt="Screen Shot 2023-06-11 at 9 17 05 PM" src="https://github.com/finos/perspective/assets/667901/53608e89-6568-4514-ae49-49b7bdce5a75">

<img width="1502" alt="Screen Shot 2023-06-11 at 9 17 16 PM" src="https://github.com/finos/perspective/assets/667901/90605ba2-3631-44a2-8bdf-bf413359c577">

### Actual Result:

NaN

### Environment:

perspective-python v2.2.1

nahoom-belete:
Hi. I noticed that when I select a different view(e.g. Y Area) then switch the view back to Y Bar after following steps 5 & 6, the expected output is displayed. Maybe this can help narrow down what the issue might be.

texodus:
Fixed (originally) in #2753 - but also this component was removed and rebuilt, and subsequently re-implemented correctly in the `viewer-charts` package.

PR #2753, opened 2024-09-16
Missing tooltip data

wakek:
This PR addresses the bug reported in #2575 and #2254 where tooltip data labels were missing their corresponding values when multiple data values are configured. The issue stemmed from an invalid property used to fetch tooltip data from the `data.row` JSON, which was missing a required “prefix”. This has now been corrected.


wakek:
Hi @texodus, could you take a look this PR to let me know if anything is amiss, thanks.

sinistersnare:
Thanks for submitting this! Although this succeeds for simple cases, it does not work for more complex examples. Luckily the code you wrote is on the right path, it only needs to be generalized further.

Here is a config that shows that your fix is not enough. Using the example data from #2575 and the file example, providing more than 1 split overwhelms your fix. Instead of hardcoding one level of split, you need to allow for any number of splits. 

```
{
  "version": "3.1.0",
  "plugin": "X/Y Line",
  "plugin_config": {},
  "columns_config": {},
  "settings": true,
  "theme": "Pro Light",
  "title": null,
  "group_by": [
    "exam_month"
  ],
  "split_by": [
    "exam_subject",
    "teacher"
  ],
  "columns": [
    "exam_month",
    "exam_score",
    null
  ],
  "filter": [],
  "sort": [],
  "expressions": {},
  "aggregates": {
    "exam_score": "avg",
    "exam_month": "any"
  }
}
```

To apply the config you can use the debug panel provided in the top right of the sidebar.

Also, We generally require a unit test for any fix, regardless of complexity. If you could write a unit test that ensures that this functionality works, it would be amazing!

Again thanks for your work in fixing this issue, I hope we can get this merged soon!

texodus (changes requested):
See @sinistersnare's feedback pls

wakek:
@sinistersnare Thanks for the feedback and suggestions. I'll work on it further.

texodus (approved):

PR #2768, opened 2024-09-26
Draft: only require boost headers, not boost_system

tomjakubowski:


tomjakubowski:
Will address this in an upcoming PR

Issue #2769, opened 2024-09-26
Not able to find `PerspectiveWidget, Table, Aggregate, Sort, Plugin`


adambouras:
### Steps to Reproduce:
 1. created a poetry project with following specs:
```
[tool.poetry.dependencies]
python = "^3.12"
perspective = "^1.0.3"
perspective-python = "^3.1.0"
pandas = "^2.2.3"
jupyterlab = "^4.2.5"
ipykernel = "^6.29.5"
bokeh = "^3.5.2"
```

### Expected Result:
```from perspective import PerspectiveWidget, Table``` should work properly in jupyter lab

### Actual Result:
see screenshot

### Environment:
```python3.12`

### Additional Context:
![image](https://github.com/user-attachments/assets/7ae3bb33-9db4-47f6-8a00-59671c72c68a)

Please advise!

Thank you

timkpaine:
The widget is moved to `perspective.widget` and the other types have been removed.

PR #2773, opened 2024-09-28
Correct Python docs

dsedivec:
I tried to update the Python docs for the changed location of imports circa 3.1.0.  The Python enums are gone (#2769).

There are several references to e.g. `perspective.Table` in prose.  I suspect those are meant to refer to the JS end of things, since `perspective.table` doesn't exist in Python, so I left them as-is.

I've really never used any of these APIs, so double checking my corrections would be wise. :)

### DIFF
--- a/rust/perspective-python/docs/lib.md
+++ b/rust/perspective-python/docs/lib.md
@@ -165,7 +165,7 @@ treat the index as a primary key:
 
 ```python
 data.set_index("datetime")
-table = perspective.Table(data, index="index")
+table = perspective.table(data, index="index")
 ```
 
 #### Time Zone Handling
@@ -281,7 +281,8 @@ of clients and large datasets.
 _*server.py*_
 
 ```python
-from perspective import Server, PerspectiveTornadoHandler
+from perspective import Server
+from perspective.handlers.tornado import PerspectiveTornadoHandler
 
 # Create an instance of Server, and host a Table
 SERVER = Server()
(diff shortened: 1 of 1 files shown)


timkpaine:
These are already fixed in-source

dsedivec:
@timkpaine I'm sorry, I don't understand, what does that mean?  Does that mean the Python documentation is auto-generated from somewhere?  I searched the source code and couldn't find anywhere the docs could be auto-generated from.

Does it mean that the Python bindings are fixed, perhaps in some branch other than `master`?

texodus:
I think what @timkpaine means is, we're already working on rewriting the docs in this module in [this un-PR-ed branch](https://github.com/finos/perspective/tree/update-docs). I appreciate the fixes, I'll make sure they are incorporated in this branch.

As per the [`CONTRIBUTING.md`](https://github.com/finos/perspective/blob/update-docs/CONTRIBUTING.md), we don't merge PRs without a DCO, nor PRs from first-time contributors without an associated Issues validated by a maintainer.

dsedivec:
@texodus Thanks for the explanation!  BTW, I might suggest putting a link to that `CONTRIBUTING.md` from the [Developer Guide](https://perspective.finos.org/docs/development.html) that's linked near the top of the `README.md`.  That was actually the first place I checked.  Probably my bad for not thinking to check for a separate `CONTRIBUTING.md`.

Thanks for this project!  It's a nice way to browse large tables of data in a Jupyter notebook.

Issue #2580, opened 2024-03-26
Field type icon in `Where` input isn't accurate


mhkeller:
### Steps to Reproduce:
 1. Load this dummy data of student exams: https://gist.github.com/mhkeller/8831f579f814af2a7415e043457d4e56
 2. Add the `exam_score` field to the `Where` filter

### Expected Result:
The icon next to `exam_score` is a `#` symbol because it's numeric

### Actual Result:
The field type symbol is an `A`, signifying a text field.

### Environment:
2.10.0 loaded from the website: https://perspective.finos.org/block/?example=file

### Additional Context:
The filter still works properly

![Screenshot 2024-03-25 at 22 22 11](https://github.com/finos/perspective/assets/498744/89c2303f-5119-488b-a8c4-8eb94b403362)

tarekquao:
But in the columns section, it shows "#" meaning it is a numeric column. only changes to A when moved to "split by", "order by", "group by".
The same effect is seen for any other column (including exam date)

![image](https://github.com/user-attachments/assets/fc597919-83e8-401c-84b8-f086c80263f0)

texodus:
Fixed in #2780

PR #2780, opened 2024-10-03
Column type icon matches the datatype type of the field in column selectors (where, group by, sortby, filter)

tarekquao:
This PR fixes the issue of icons of columns defaulting to string icon instead of the right data type in issue #2580

### DIFF
--- a/rust/perspective-viewer/src/rust/components/column_selector/filter_column.rs
+++ b/rust/perspective-viewer/src/rust/components/column_selector/filter_column.rs
@@ -468,7 +468,7 @@ impl Component for FilterColumn {
                 <LocalStyle href={css!("filter-item")} />
                 <div class="pivot-column-border">
                     // <TypeIcon ty={ColumnType::String} />
-                    <TypeIcon ty={final_col_type}/>
+                    <TypeIcon ty={final_col_type} />
                     <span class="column_name">{ filter.column().to_owned() }</span>
                     <FilterOpSelector
                         class="filterop-selector"
--- a/rust/perspective-viewer/src/rust/components/column_selector/filter_column.rs
+++ b/rust/perspective-viewer/src/rust/components/column_selector/filter_column.rs
@@ -456,6 +456,8 @@ impl Component for FilterColumn {
         }
         .unwrap_or_default();
 
+        let final_col_type = col_type.expect("Unknown column");
+
         html! {
             <div
                 class="pivot-column-draggable"
@@ -465,7 +467,8 @@ impl Component for FilterColumn {
             >
                 <LocalStyle href={css!("filter-item")} />
                 <div class="pivot-column-border">
-                    <TypeIcon ty={ColumnType::String} />
+                    // <TypeIcon ty={ColumnType::String} />
+                    <TypeIcon ty={final_col_type}/>
                     <span class="column_name">{ filter.column().to_owned() }</span>
                     <FilterOpSelector
                         class="filterop-selector"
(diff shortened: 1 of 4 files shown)


rust/perspective-viewer/src/rust/components/column_selector/filter_column.rs
```diff
@@ -456,6 +456,8 @@ impl Component for FilterColumn {
         }
         .unwrap_or_default();
 
+        let final_col_type = col_type.expect("Unknown column");
```
texodus, line 459: There is no point in using `expect()` in Perspective WASM code because the error message does not propagate to the JS console. Generally we want to safely handle when this is missing because `panic!` leads to a frozen/broken web app that has to be refreshed.

rust/perspective-viewer/src/rust/components/column_selector/filter_column.rs
```diff
@@ -465,7 +467,8 @@ impl Component for FilterColumn {
             >
                 <LocalStyle href={css!("filter-item")} />
                 <div class="pivot-column-border">
-                    <TypeIcon ty={ColumnType::String} />
+                    // <TypeIcon ty={ColumnType::String} />
```
texodus, line 470: ☠️

rust/perspective-viewer/src/rust/components/column_selector/pivot_column.rs
```diff
@@ -10,18 +10,20 @@
 // ┃ of the [Apache License 2.0](https://www.apache.org/licenses/LICENSE-2.0). ┃
 // ┗━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┛
 
-use perspective_client::ColumnType;
+// use perspective_client::ColumnType;
```
texodus, line 13: ☠️

rust/perspective-viewer/src/rust/components/column_selector/pivot_column.rs
```diff
@@ -69,7 +78,8 @@ impl Component for PivotColumn {
                 ondragend={dragend}
             >
                 <div class="pivot-column-border">
-                    <TypeIcon ty={ColumnType::String} />
+                    <TypeIcon ty={col_type} />
+                    // <TypeIcon ty={ColumnType::String} />
```
texodus, line 82: ☠️

rust/perspective-viewer/src/rust/components/column_selector/sort_column.rs
```diff
@@ -11,7 +11,7 @@
 // ┗━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┛
 
 use perspective_client::config::*;
-use perspective_client::ColumnType;
+// use perspective_client::ColumnType;
```
texodus, line 14: ☠️

texodus (approved):
Thanks for the PR! A few notes to keep in mind for future PRs:

* We normally ask for a unit test with all PRs, but this is relatively minor so I'll let it slide. For the record though, testing this should be quite easy - just set up a viewer with a config and check the DOM output ([many of our `perspective-viewer` tests do this already e.g.](https://github.com/finos/perspective/blob/master/rust/perspective-viewer/test/js/expressions.spec.js#L87)).
* Watch for ☠️ dead code ☠️ please, there's no point in leaving commented-out code when we have the project history on GitHub.
* Generally prefer _not_ to use `.expect()` in Perspective WASM code, as it leads to bad errors when an expectation is missed (sometimes generations after your PR is merged).

PR #2798, opened 2024-10-22
arrow: Fix cross-compiling errors on macos

tomjakubowski:
This fixes the issue with missing LZ4 symbols in the arm64 wheel first introduced in v3.1.1.

The crux of the issue is that Arrow was building lz4 (and zstd) in a separate cmake build, and it does not specially pass on the `CMAKE_OSX_ARCHITECTURES` flag that was set when perspective-server's build script invokes cmake.

Arrow's build does forward on the value of CMAKE_TOOLCHAIN_FILE to its external dependencies, if it's set, so this commit creates two minimal toolchain files and uses them in the Mac build to set `CMAKE_OSX_ARCHITECTURES` (and a couple other variables) correctly when cross-compiling.

a good follow-up would be to see if we can enable Python tests on arm64 using a macos-14 runner, which is available for public repos: <https://docs.github.com/en/actions/using-github-hosted-runners/using-github-hosted-runners/about-github-hosted-runners#standard-github-hosted-runners-for-public-repositories>

### Pull Request Checklist

  - I am planning to follow up with a CI job which tests the arm64 Python wheel

### DIFF
--- a/cmake/toolchains/darwin-arm64.cmake
+++ b/cmake/toolchains/darwin-arm64.cmake
@@ -1,0 +1,24 @@
+#  ┏━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┓
+#  ┃ ██████ ██████ ██████       █      █      █      █      █ █▄  ▀███ █       ┃
+#  ┃ ▄▄▄▄▄█ █▄▄▄▄▄ ▄▄▄▄▄█  ▀▀▀▀▀█▀▀▀▀▀ █ ▀▀▀▀▀█ ████████▌▐███ ███▄  ▀█ █ ▀▀▀▀▀ ┃
+#  ┃ █▀▀▀▀▀ █▀▀▀▀▀ █▀██▀▀ ▄▄▄▄▄ █ ▄▄▄▄▄█ ▄▄▄▄▄█ ████████▌▐███ █████▄   █ ▄▄▄▄▄ ┃
+#  ┃ █      ██████ █  ▀█▄       █ ██████      █      ███▌▐███ ███████▄ █       ┃
+#  ┣━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┫
+#  ┃ Copyright (c) 2017, the Perspective Authors.                              ┃
+#  ┃ ╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌ ┃
+#  ┃ This file is part of the Perspective library, distributed under the terms ┃
+#  ┃ of the [Apache License 2.0](https://www.apache.org/licenses/LICENSE-2.0). ┃
+#  ┗━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┛
+
+# Minimal toolchain file to ensure that Arrow's bundled dependency builds use
+# the correct architecture for Mac builds.
+
+# (Arrow does not as of Oct 2024 pass on CMAKE_OSX_ARCHITECTURES to its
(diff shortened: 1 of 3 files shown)


rust/perspective-server/build/psp.rs
```diff
@@ -26,10 +26,24 @@ pub fn cmake_build() -> Result<Option<PathBuf>, std::io::Error> {
...
-        dst.define("CMAKE_OSX_ARCHITECTURES", "x86_64");
-    } else if std::env::var("PSP_ARCH").as_deref() == Ok("arm64") {
-        dst.define("CMAKE_OSX_ARCHITECTURES", "arm64");
+
+    if cfg!(target_os = "macos") {
+        // Set CMAKE_OSX_ARCHITECTURES et al. for Mac builds.  Arrow does not forward on
+        // CMAKE_OSX_ARCHITECTURES and but it does forward on a
+        // CMAKE_TOOLCHAIN_FILE. In Conda builds, the environment sets
+        // `CMAKE_ARGS` up with various toolchain arguments. This block may need
+        // to be patched out or adjusted for Conda.
```
tomjakubowski, line 35: by this I mean this block may need some `if !conda_build` style guards. I am not sure which wins out between a value set in `CMAKE_TOOLCHAIN_FILE` vs. a value set later at the command line

cmake/toolchains/darwin-arm64.cmake
```diff
@@ -0,0 +1,24 @@
...
+#  ┃ of the [Apache License 2.0](https://www.apache.org/licenses/LICENSE-2.0). ┃
+#  ┗━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┛
+
+# Minimal toolchain file to ensure that Arrow's bundled dependency builds use
+# the correct architecture for Mac builds.
+
+# (Arrow does not as of Oct 2024 pass on CMAKE_OSX_ARCHITECTURES to its
+# dependencies if it is set, but does pass on a toolchain file)
+
+set(CMAKE_OSX_ARCHITECTURES "arm64" CACHE STRING "Target architecture for Mac builds")
```
tomjakubowski, line 19: don't think this needs the `CACHE` etc. If it's confusing I'm happy to remove it

tomjakubowski:
If it turns out we can't easily get arm64 tests running on macos-14, then a cheap though maybe overly specific regression test would be to run something like `! nm -u perspective.abi3.so | grep -i lz4`

rust/perspective-server/build/psp.rs
tomjakubowski: funny enough, the conda 3.1.1 macos cross-compiling builds were apparently unaffected -- their import tests succeeded. https://github.com/conda-forge/perspective-feedstock/runs/31853180964

I'm not exactly sure why, but conda-build passes its toolchain arguments to us (in CMAKE_ARGS) with compilers/tools with arm64 in the name, `-DCMAKE_C_COMPILER=arm64-…-cc -DCMAKE_LIBTOOL=arm64-…-cc`, which I presume is a compiler toolchain which targets arm64 without needing any other flags, and doesn't set `CMAKE_OSX_ARCHITECTURE` itself. And Arrow does forward those specific flags on to its external dependencies

rust/perspective-server/build/psp.rs
```diff
@@ -26,10 +26,24 @@ pub fn cmake_build() -> Result<Option<PathBuf>, std::io::Error> {
...
+        let toolchain_file = match std::env::var("PSP_ARCH").as_deref() {
+            Ok("x86_64") => "./cmake/toolchains/darwin-x86_64.cmake",
+            Ok("aarch64") => "./cmake/toolchains/darwin-arm64.cmake",
+            arch @ Ok(_) | arch @ Err(_) => {
+                panic!("Unknown PSP_ARCH value: {:?}", arch)
```
tomjakubowski, line 40: this might be the wrong behavior if PSP_ARCH isn't set (which I think is the case for local dev). I guess it should just not set a toolchain file when PSP_ARCH is unset? (but perhaps still error + bail out when PSP_ARCH is an unknown value)

tomjakubowski:
Shoot, CI doesn't build a wheel so we can't test it. I'll open another PR which 

I did create a one-off repo with one of the free arm64 macos-14 runners. These are apparently very new, like released in the last couple weeks. Here's a run from the `v3.1.1` release which demonstrates the issue https://github.com/tomjakubowski/perspective-arm64-macos-wheel-test/actions/runs/11452924272/job/31864559075

tomjakubowski:
OK, grabbed an artifact from !2799. Did a test run with it an arm64 macos-14 runner here, it passes the pytest suite https://github.com/tomjakubowski/perspective-arm64-macos-wheel-test/actions/runs/11453537555/job/31866177154

texodus (approved):

PR #2807, opened 2024-10-28
Escape esbuild plugin's symbol replacement.

timbess:
Minification sometimes adds `$` to variable names causing it to be interpreted as regex rather than literally.


texodus (approved):

PR #2811, opened 2024-10-28
Generate sdist by hand

tomjakubowski:
This updates the perspective-python build to create an sdist tarball by
hand, not going through `maturin sdist`. This bypasses some issues we
have seen with Maturin mispackaging the perspective Cargo workspace in
the sdist.

Maturin is still used as the `build-backend` to build a wheel from the
sdist.  CI still uses `maturin build` to build wheels.  Local
development builds still use `maturin develop`.

The sdist now includes the .data directory so that Maturin will place it
in the wheel correctly.  Building an sdist with PSP_BUILD_SDIST=1 will
now error if it appears the jupyterlab plugin was not built into the
.data directory.

Most of the work is in generating a PKG-INFO file.  Its output matches
what's in the 3.1.2 sdist on pypi, minus a correction to fix the
`Home-page` field miscapitalized by Maturin and some unimportant
whitespace changes in `Require-Dist`.
 
### Pull Request Checklist

        Discussions, this PR applies to. (N/A)

### DIFF
--- a/rust/perspective-python/Cargo.toml
+++ b/rust/perspective-python/Cargo.toml
@@ -25,8 +25,12 @@ include = [
     "bench/**/*.py",
     "build.rs",
     "./Cargo.toml",
+    "README.md",
     "./package.json",
     "perspective/**/*.py",
+    "perspective/extension",
+    "perspective/templates",
+    "perspective/tests/**/*.arrow",
     "./pyproject.toml",
     "src/**/*.rs",
     "docs/**/*",
@@ -52,9 +56,12 @@ pyo3-build-config = "0.20.2"
 python-config-rs = "0.1.2"
 
 [dependencies]
+# NOTE: when building from the git repo, these perspective-* dependencies are
+# overridden with path dependencies in .cargo/config.toml.  This is done to
+# support the sdist, which doesn't include these packages.
+perspective-client = { version = "3.1.2" }
+perspective-server = { version = "3.1.2" }
 async-lock = "2.5.0"
-perspective-client = { version = "3.1.2", path = "../perspective-client" }
-perspective-server = { version = "3.1.2", path = "../perspective-server" }
 pollster = "0.3.0"
 extend = "1.1.2"
 futures = "0.3.28"
--- a/rust/perspective-python/build.mjs
+++ b/rust/perspective-python/build.mjs
@@ -14,6 +14,9 @@ import * as fs from "node:fs";
 import sh from "../../tools/perspective-scripts/sh.mjs";
 import * as url from "url";
 import * as TOML from "@iarna/toml";
+import * as tar from "tar";
+import * as glob from "glob";
+import * as path from "path";
 
 const __dirname = url.fileURLToPath(new URL(".", import.meta.url)).slice(0, -1);
 const pkg = JSON.parse(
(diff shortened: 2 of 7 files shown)


tomjakubowski:
test_js failure looks familiar, I think it's a known flaky test

In the [sdist artifact](https://github.com/finos/perspective/actions/runs/11562521495/artifacts/2114692332), looks like the labextension is packaged as expected 

And the [wheel artifact](https://github.com/finos/perspective/actions/runs/11562521495/artifacts/2114692261) still has the one copy of the labextension, in the .data directory, as expected

rust/perspective-python/Cargo.toml
```diff
@@ -52,9 +52,12 @@ pyo3-build-config = "0.20.2"
 python-config-rs = "0.1.2"
 
 [dependencies]
+# NOTE: when building from the git repo, these perspective-* dependencies are
+# overridden with path dependencies in .cargo/config.toml.  This is done to
+# support the sdist, which doesn't include these packages.
```
tomjakubowski, line 57: there may be pitfalls in doing it this way I'm not aware of, but it seemed less messy than patching the Cargo.toml file in the tarball.

tomjakubowski:
Did a manual run in my `oneoffs` repo to verify that the sdist installs on macOS/aarch64: https://github.com/tomjakubowski/perspective-oneoff-tests/actions/runs/11584975281

I think we probably want at least the option to test in CI that (1) installs from the sdist (2) runs the pytest test suite on the installed package (3) verifies that the labextension is listed in `jupyter labextension list`. Maybe a manual-only (workflow_dispatch) workflow, or one which runs only on release tags? I'll plug away at something like that.

tomjakubowski:
Another CI job with a diff of the sdist from this branch compared to the 3.1.2 sdist on pypi: https://github.com/tomjakubowski/perspective-oneoff-tests/actions/runs/11585843746/job/32255425110

One thing that stood out is that the sdist has lost the `release` profile flags from the workspace root's Cargo.toml. May want to add those back

rust/perspective-python/build.mjs
```diff
@@ -111,22 +114,96 @@ if (build_wheel) {
...
+    );
+    if (!fs.existsSync(testfile)) {
+        throw new Error(
+            "labextension is not present in data directory, please build `perspective-jupyterlab`"
+        );
+    }
+    const readme_md = fs.readFileSync("./README.md");
+    const pkg_info = generatePkgInfo(pyproject, cargo, readme_md);
+    fs.writeFileSync("./PKG-INFO", pkg_info);
+    const include_paths = Array.from(cargo["package"]["include"]).concat([
```
tomjakubowski, line 139: this is excluding some files from the `perspective` Python folder that are otherwise added by Maturin, like test arrows

https://github.com/PyO3/maturin/blob/main/src/source_distribution.rs#L624-L646

will port that

tomjakubowski: decided against porting, instead I just expanded the set of files included with the crate in `Cargo.toml`. that way is more explicit and it's less code, to boot

rust/perspective-python/Cargo.toml
tomjakubowski: Since the new test_python_sdist job in CI runs in a git checkout, the path directives in .cargo/config.toml will apply when that job builds the perspective-python crate (and wheel) from the sdist. That means in test_python_sdist, changes in the git repo to perspective-client and perspective-server which haven't yet been published to crates.io would be tested too. 

End users building from that same sdist artifact, if it's from between releases, would get a different result, building against the most recent release of those crates on crates.io. Workarounds for them would be to set up a `.cargo/config.toml` (maybe one in ~? I don't know if pip's build isolation would defeat that) which uses paths pointing into a local perspective git clone. This is an unfortunate downside of bundling only perspective-python's source in the sdist

tomjakubowski: Oh, uh, I stand corrected. The sdist build in CI is downloading from crates.io https://github.com/tomjakubowski/perspective-oneoff-tests/actions/runs/11604564693/job/32313519748#step:8:187

I think that's because of pip's build isolation -- the build is happening in some other working directory, so Cargo doesn't find .cargo/config.toml

will want to adjust the CI job. I think, like I mentioned setting it in `$HOME/.cargo/config.toml` will work

tomjakubowski:
Superceed by #2817

tomjakubowski:
The CI job is still a WIP. Since we are purposely . May want to revisit that.

Issue #2693, opened 2024-07-24
Support Firefox


texodus:
This PR is a tracking issue for adding official Perspective support for [Mozilla Firefox](https://www.mozilla.org/en-US/firefox/)

Perspective (as of this writing) runs on Firefox already, but does not _officially_ support it (Chrome only), principally because we do not currently run our integration test suite against this browser on our Github Actions CI. Without the test suite, there is no mechanism to assert that any fixes we _do_ add for Firefox continue to work (or are not mistakenly overwritten in the very next PR), which makes it hard to justify accepting whack-a-mole PRs which attempt to address these issues.

As a starting point, we'll need to run our `playwright` test suite against Firefox on CI and see what breaks - from there, we should be able to put together a proper roadmap of the work left to do to get every test feature working.

A (not yet exhaustive) list of reports of Firefox issues follows. We'll want to add tests for in the course of this task, as I've closed them to minimize "Issues" noise. If you are reading this and about to report a Firefox issue, please post as a reply in this thread instead - I will likely close a re-direct the Issue here otherwise!

#2357
#2115
#1983

Issue #1983, opened 2022-10-03
treemap svg elements missing position and size making them invisible on firefox


GREEB:
## Bug Report

#1327 is maybe the same bug but its closed.

### Steps to Reproduce:
Go to https://perspective.finos.org/ and toggle example to treemap
![image](https://user-images.githubusercontent.com/1221769/193642164-1091168b-62d8-4be8-bb97-51bc90b10470.png)

### Expected Result:
Open same page with `Version 104.0.5112.102 (Official Build, ungoogled-chromium) (64-bit)`
![image](https://user-images.githubusercontent.com/1221769/193642325-3ce401e7-9153-4587-b036-ab607a66a9d0.png)

### Environment:
`Firefox 105.0.1 (64-bit)`
`Windows 10`

### Additional Context:
it seems that `x: y: width: height:` do not get set in Firefox

So in Chromium the rect style is something like:
 ```css
{    
    x: 0px;
    y: 89.4314;
    width: 126.966px;
    height: 56.4545px;
    fill: rgb(71, 120, 194);
}
```

In Firefox when svg are "invisible": 
 ```css
{
   x: 0px;
   y: 0px;
   fill: rgb(36, 38, 40);
}
```

GREEB:
another example using @latest for sure https://bl.ocks.org/texodus/raw/6b4dcebf65db4ebe4fe53a6de5ea0b48/?raw=true
here it seems the `rect` does not have `x: y:`
### Steps to Reproduce:
- Go to https://bl.ocks.org/texodus/raw/6b4dcebf65db4ebe4fe53a6de5ea0b48/?raw=true
- Set `Ratings vs Sales` view to "Treemap"
![image](https://user-images.githubusercontent.com/1221769/193649341-495ba4aa-865e-4f20-8d60-827916d162d8.png)

### Expected
![image](https://user-images.githubusercontent.com/1221769/193649442-2c9c3663-f233-4b9c-834d-99dd056b6d37.png)

Environment:

`Firefox 105.0.1 (64-bit)`
`Windows 10`

waynexia:
Cannot render treemap in the example page too. On Firefox `111.0.1 (64-bit)`

![image](https://user-images.githubusercontent.com/15380403/230290155-1b6d9c17-6121-4118-bd8d-5470b800b69a.png)

![image](https://user-images.githubusercontent.com/15380403/230290191-1a5b02c4-1c73-4cb0-a99f-6a19a5794128.png)

[...]

PR #2813, opened 2024-10-29
fix treemap rendering issue in firefox

onesimus-wiafe:
This pull request fixes one of the Firefox issues in #2693, specifically issue #1983
I used string interpolation for the style on the rect element (for x and y) to introduce layout consistency by adding px to the `d.x0` and `d.y0` values. I realized that without this, Firefox ignored some of the values, hence the issue.

### Pull Request Checklist

-   [ ] Include any relevant Documentation changes related to this change.


texodus (approved):
Thanks for the PR!

Perspective _does not currently support FireFox_ (See #2818) and its generally our policy not to merge fixes for unsupported platforms because they are difficult for correctness. However in this case, this fix is correct and trivial.

PR #2817, opened 2024-10-31
Fix `perspective-python` sdist and add sdist tests to CI

tomjakubowski:
This updates the perspective-python build to create an sdist tarball by
hand, not going through `maturin sdist`. This bypasses some issues we
have seen with Maturin mispackaging the perspective Cargo workspace in
the sdist.

Maturin is still used as the `build-backend` to build a wheel from the
sdist.  CI still uses `maturin build` to build wheels.  Local
development builds still use `maturin develop`.

The sdist now includes the .data directory so that Maturin will place it
in the wheel correctly.  Building an sdist with PSP_BUILD_SDIST=1 will
now error if it appears the jupyterlab plugin was not built into the
.data directory.

Most of the work is in generating a PKG-INFO file.  Its output matches
what's in the 3.1.2 sdist on pypi, minus a correction to fix the
`Home-page` field miscapitalized by Maturin and some unimportant
whitespace changes in `Require-Dist`.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -436,7 +436,7 @@ jobs:
     # /` |     |    /|
     # `--'     '   `-'
     #
-    build_and_test_juptyerlab:
+    build_and_test_jupyterlab:
         needs: [build_js, build_python]
         runs-on: ${{ matrix.os }}
         strategy:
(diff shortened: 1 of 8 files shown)


tomjakubowski:
This is #2811 plus another commit which adds a CI job to test the sdist. See my commentary in the commit message and on #2811 for some caveats

.github/workflows/build.yaml
```diff
@@ -765,7 +828,7 @@ jobs:
     publish:
         needs:
             [
-                build_and_test_juptyerlab,
+                build_and_test_jupyterlab,
```
texodus, line 831: 🤯

texodus (approved):

PR #2830, opened 2024-11-06
Export Table and View with aliasing to discourage instantiation

timkpaine:
https://github.com/finos/perspective/pull/2754 remove the top level `Table` and `View` as they are not constructible. However, libraries like `pydantic` and `csp` like to have types for type validation, so here we reintroduce them but alias them to types to discourage direct instantiation.

### Pull Request Checklist

-   [x] Description which clearly states what problems the PR solves.
-   [ ] Description contains a link to the Github Issue, and any relevent
        Discussions, this PR applies to.
-   [ ] Include new tests that fail without this PR but passes with it.
-   [ ] Include any relevent Documentation changes related to this change.
-   [x] Verify all commits have been _signed_ in accordance with the DCO policy.
-   [ ] Reviewed PR commit history to remove unnecessary changes.
-   [ ] Make sure your PR passes _build_, _test_ and _lint_ steps _completely_.

PR #2831, opened 2024-11-06
Add back Table and View exports, make constructors throw like pyarrow.Table

timkpaine:
followup to https://github.com/finos/perspective/pull/2830

reexpose these types for `pydantic`, `csp`, any other type checking libraries, and make their constructors throw helpful error messages:

Screenshot 2024-11-06 at 18 45 51


texodus (approved):

PR #2841, opened 2024-11-08
Import importlib metadata submodule

timkpaine:
This PR adds an explicit import for `importlib.metadata` to the `perspective.widget` module, resolving (?) a reported import exception on Conda (?) environments. While this change is technically correct, the root cause of this module not being cached on these envs remains undetermined. I've included no tests, documentation or description, because  I'm a loner rebel who doesn't play by society's rules!

- Description added retroactively by @texodus. Dictated but not read.


texodus:
FYI actual import path in my local non-conda environment which caused this to work was to [`pygments`](https://github.com/pygments/pygments/blob/master/pygments/plugin.py#L35).

texodus (approved):

PR #2843, opened 2024-11-09
sdist licensing follow ups

tomjakubowski:
This adds LICENSE.md from perspective to the sdist, and uses `cargo-bundle-licenses` to bundle perspective-python's dependency licenses into LICENSE_THIRDPARTY_cargo.yml, which is also added to the sdist.

When Maturin builds a wheel, it searches the source for files named `LICENSE*` (among others) and includes them in dist-info. C++ license packaging can use this same framework: we'll generate a file in the perspective-python source tree which aggregates licenses from arrow and other bundled C++ dependencies.

Adds a basic test of the license packaging to the sdist test. Note that this doesn't test the wheels we distribute, which takes a different build path in CI.

### DIFF
--- a/rust/perspective-python/build.mjs
+++ b/rust/perspective-python/build.mjs
@@ -134,10 +134,12 @@ if (build_sdist) {
         );
     }
     const readme_md = fs.readFileSync("./README.md");
+    fs.copyFileSync("../../LICENSE.md", "./LICENSE.md");
     const pkg_info = generatePkgInfo(pyproject, cargo, readme_md);
     fs.writeFileSync("./PKG-INFO", pkg_info);
     const include_paths = Array.from(cargo["package"]["include"]).concat([
         data_dir,
+        "./LICENSE.md",
         "./PKG-INFO",
     ]);
     const files = glob.globSync(include_paths);
--- a/rust/perspective-python/build.mjs
+++ b/rust/perspective-python/build.mjs
@@ -37,6 +37,7 @@ const version = pkg.version.replace(/-(rc|alpha|beta)\.\d+/, (x) =>
 );
 
 fs.mkdirSync(`./perspective_python-${version}.data`, { recursive: true });
+fs.copyFileSync("../../LICENSE.md", "./LICENSE.md");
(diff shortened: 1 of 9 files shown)


rust/perspective-python/build.mjs
```diff
@@ -37,6 +37,7 @@ const version = pkg.version.replace(/-(rc|alpha|beta)\.\d+/, (x) =>
 );
 
 fs.mkdirSync(`./perspective_python-${version}.data`, { recursive: true });
+fs.copyFileSync("../../LICENSE.md", "./LICENSE.md");
```
tomjakubowski, line 40: it turns out `maturin develop` needs this file around now that it's specified in `pyproject.toml` as `{license { file = "LICENSE.md" }}`, so I moved it out of the sdist branch so it always runs.

rust/perspective-server/build/psp.rs
```diff
@@ -84,15 +84,15 @@ pub fn cmake_build() -> Result<Option<PathBuf>, std::io::Error> {
...
+            "cargo:warning=Setting CMAKE_ARGS from environment {:?}",
             cmake_args
         );
         for arg in Shlex::new(&cmake_args) {
             dst.configure_arg(arg);
         }
     }
 
-    println!("cargo:warning=MESSAGE Building cmake {}", profile);
+    println!("cargo:warning=Building cmake {}", profile);
```
tomjakubowski, line 95: MESSAGE is not needed in these warnings, I think we misunderstood the docs. See this CI run: https://github.com/tomjakubowski/perspective-oneoff-tests/actions/runs/11752078774/job/32743108312

tomjakubowski:
Not ready yet, need to update CI so that the metadata is copied for perspective-python (and then pulled down for the sdist build). I'll find an hour to do that this weekend

Also, a little surprising that `node-tar` silently ignores missing files 😬. Can make them error by doing this https://github.com/isaacs/node-tar?tab=readme-ov-file#warnings-and-errors

texodus (approved):

PR #2849, opened 2024-11-15
Support input/output to/from polars via python (not native rust)

timkpaine:
This is a naive rust implementation of conversion to/from polars. I describe it as “naive” as it incurs an “extra” copy of polars-wrapped arrow table to binary buffer to perspective. 

A test suite is included, copied from pandas but omitting some pandas-specific functionality, and skipping some stuff that might warrant a further look (polars converts to several arrow types that we don’t currently support). 

Some initial TODOs:
- [x] tests for `LazyFrame`
- [ ] tests for outbound
- [x] docs updates

### DIFF
--- a/rust/perspective-python/docs/client/to_pandas.md
+++ b/rust/perspective-python/docs/client/to_pandas.md
@@ -1,0 +1,1 @@
+Serialize the data to a `pandas.DataFrame`.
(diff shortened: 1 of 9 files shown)


texodus (approved):

Issue #2566, opened 2024-03-12
[Chart Plugin] Multi-tier grouping by date displays epoch time in axis label


brianblakely:
### Steps to Reproduce:
 1. Go to https://perspective.finos.org/blocks/editable/index.html
 2. Switch to `X Bar` plugin
 3. Group by: `Ship Date` > `Product Name`

### Expected Result:
Date labels rendered on Y axis should match formatting one would see in Datagrid plugin

### Actual Result:
Dates appear to be rendered in epoch time

![image](https://github.com/finos/perspective/assets/401422/182b7cc5-1497-43e7-9a51-2223d3c78717)

### Environment:
Perspective 2.9.0
Chrome 122

PR #2857, opened 2024-11-19
Show appropriate date when date is part of grouped data on chart.

onesimus-wiafe:
This PR addresses issue #2566. When the data is grouped and a date is part of the grouped data, the date values are then represented in epoch format. I couldn't figure out the exact source of the problem, and this solution addresses the problem. I included a function in the filter.ts file since all the plugins affected by this issue import the functions in filter.ts. I refine the date type values before the functions in filter.ts run. The tests included confirm this solution too.

### Pull Request Checklist

### DIFF
--- a/packages/perspective-viewer-d3fc/src/ts/legend/filter.ts
+++ b/packages/perspective-viewer-d3fc/src/ts/legend/filter.ts
@@ -17,7 +17,7 @@ function refineDateData(
     settings: Settings,
     data: any[] | undefined = undefined
 ) {
-    let dataToRefine = data !== undefined ? data : settings.data;
+    let dataToRefine = data ?? settings.data;
 
     settings.crossValues.forEach((value, index) => {
         if (
--- a/packages/perspective-viewer-d3fc/src/ts/legend/filter.ts
+++ b/packages/perspective-viewer-d3fc/src/ts/legend/filter.ts
@@ -18,7 +18,7 @@ function refineDateData(settings: Settings) {
     const { crossValues } = settings;
 
     crossValues.forEach(({ type }, index) => {
-        console.log("Type:", type)
+        console.log("Type:", type);
         const formatType =
             type === "date" || type === "datetime"
                 ? (value: any) => new Date(value).toLocaleString()
--- a/packages/perspective-viewer-d3fc/src/ts/legend/filter.ts
+++ b/packages/perspective-viewer-d3fc/src/ts/legend/filter.ts
@@ -22,14 +22,15 @@ function refineDateData(
 
     crossValues.forEach(({ type }, index) => {
         const formatType =
-            (type === "date" || type === "timestamp" || type === "datetime")
-             ? (value: any) => new Date(value).toLocaleDateString()
-            : ((type === "time") ? (value: any) => new Date(value).toLocaleTimeString()
-            : null);
+            type === "date" || type === "timestamp" || type === "datetime"
+                ? (value: any) => new Date(value).toLocaleDateString()
+                : type === "time"
+                ? (value: any) => new Date(value).toLocaleTimeString()
+                : null;
 
         if (formatType) {
             dataToRefine.forEach((row: { __ROW_PATH__: any[] }) => {
-                row.__ROW_PATH__[index] = formatType(row.__ROW_PATH__[index])
+                row.__ROW_PATH__[index] = formatType(row.__ROW_PATH__[index]);
             });
         }
     });
--- a/packages/perspective-viewer-d3fc/src/ts/legend/filter.ts
+++ b/packages/perspective-viewer-d3fc/src/ts/legend/filter.ts
@@ -13,11 +13,41 @@
 import { groupFromKey } from "../series/seriesKey";
 import { DataRowsWithKey, Settings } from "../types";
 
+function refineDateData(
+    settings: Settings,
+    data: any[] | undefined = undefined
+) {
+    let dataToRefine = data !== undefined ? data : settings.data;
+
+    settings.crossValues.forEach((value, index) => {
+        if (
+            value.type === "date" ||
+            value.type === "timestamp" ||
+            value.type === "datetime"
+        ) {
+            dataToRefine.forEach((row: { __ROW_PATH__: any[] }) => {
+                row.__ROW_PATH__[index] = new Date(
+                    row.__ROW_PATH__[index]
+                ).toLocaleDateString();
+            });
+        }
+        if (value.type === "time") {
+            dataToRefine.forEach((row: { __ROW_PATH__: any[] }) => {
+                row.__ROW_PATH__[index] = new Date(
+                    row.__ROW_PATH__[index]
+                ).toLocaleTimeString();
+            });
+        }
+    });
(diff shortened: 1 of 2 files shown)


packages/perspective-viewer-d3fc/src/ts/legend/filter.ts
```diff
@@ -13,11 +13,41 @@
 import { groupFromKey } from "../series/seriesKey";
 import { DataRowsWithKey, Settings } from "../types";
 
+function refineDateData(
+    settings: Settings,
+    data: any[] | undefined = undefined
+) {
+    let dataToRefine = data !== undefined ? data : settings.data;
```
aszenz, line 20:
```suggestion
    let dataToRefine = data ?? settings.data;
```

packages/perspective-viewer-d3fc/src/ts/legend/filter.ts
```diff
@@ -13,11 +13,36 @@
 import { groupFromKey } from "../series/seriesKey";
 import { DataRowsWithKey, Settings } from "../types";
 
+function refineDateData(
+    settings: Settings,
+    data: any[] | undefined = undefined
```
texodus, line 18: This is a strange API for this function - especially because you only call it with `settings.data` as the 2nd argument.

packages/perspective-viewer-d3fc/test/js/axisLabel.spec.ts
```diff
@@ -0,0 +1,335 @@
...
+     */
+    let isNumericOnly = /^\d+$/.test(dateValues[index]);
+
+    expect(isNumericOnly).toEqual(false);
+}
+
+test.describe("Axis Values With Grouped Data With A Date Field In The Group", () => {
+    test("X Bar y-axis label with grouped data", async ({ page }) => {
+        await page.click('div[data-plugin="X Bar"]');
+        await page.waitForSelector("perspective-viewer");
```
texodus, line 66: This doesn't do anything if you've already clicked the plugin button - you are awaiting a selector you already interacted with on the previous line.

packages/perspective-viewer-d3fc/test/js/axisLabel.spec.ts
```diff
@@ -0,0 +1,335 @@
...
+        const dateValues = await page.evaluate(async () => {
+            let viewer = document.querySelector("perspective-viewer");
+
+            if (!viewer) return [];
+
+            const plugin_element = viewer.querySelector(
+                `perspective-viewer-d3fc-xbar`
+            );
+
+            if (!plugin_element) return [];
```
texodus, line 77: These early returns only fail because the `confirmDateData` function fails for empty arrays - why not just throw/return an error here so we get a real message for the test failure?

packages/perspective-viewer-d3fc/test/js/axisLabel.spec.ts
```diff
@@ -0,0 +1,335 @@
...
+
+            const shadowRoot = plugin_element.shadowRoot;
+            const elements = shadowRoot.querySelectorAll(
+                "div d3fc-group d3fc-svg.y-axis.left-axis svg, g.group"
+            );
```
texodus, line 82: Why does this select `div d3fc-group d3fc-svg.y-axis.left-axis svg` only to drop it on line 85? I think this was intended to be `div d3fc-group d3fc-svg.y-axis.left-axis svg g.group` (without a comma), but because you only take the "last" group the extra selects does not affect the test outcome. It does make this code hard to read though. You could also just combine this entire section into the selector `div d3fc-group d3fc-svg.y-axis.left-axis svg g.group:last-child g.tick text`.

packages/perspective-viewer-d3fc/src/ts/legend/filter.ts
```diff
@@ -13,11 +13,36 @@
...
+) {
+    let dataToRefine = data ?? settings.data;
+    const { crossValues } = settings;
+
+    crossValues.forEach(({ type }, index) => {
+        const formatType =
+            type === "date" || type === "timestamp" || type === "datetime"
+                ? (value: any) => new Date(value).toLocaleDateString()
```
texodus, line 26: We should really use the formatting logic from `perspective-viewer-datagrid` here ... but in the interim, this only works for `date` Perspective types. `datetime` types should use `toLocaleString()` for now please.

texodus (changes requested):
Thanks for the PR!

I suspect some parts of this PR are AI generated. We do not yet have an official project policy on this, but I'm not eager to review such code. If you're going to use these tools, please _rigorously_ check for relevance, brevity and sanity before submitting - technically _correct_ code is not enough. If this is not AI generated code - please do this anyway.

Regardless of the tooling, I would like some cleanup of this code before we merge it (see review comments).

packages/perspective-viewer-d3fc/test/js/axisLabel.spec.ts
```diff
@@ -0,0 +1,335 @@
...
+    await page.click(".plugin-select-item");
+});
+
+function confirmDateData(dateValues: any[]) {
+    // expect the first to be a date.
+    expect(!isNaN(Date.parse(dateValues[0]))).toEqual(true);
+
+    // expect any random index of axisLabels to also be a date.
+    let len = dateValues.length;
+    const index = Math.floor(Math.random() * len);
```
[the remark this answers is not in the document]
onesimus-wiafe: I thought it would be wasteful to check through the entire array and if the first, last, and a random index of the array contain dates, it was highly likely that the array contains only date types. `Date.parse()` confirms they are dates but not the way to verify that they were not in epoch form. For 58, I realized that only epoch formats would be dates after completing the regex and it felt unnecessary to continue with the steps I outlined. Even if it was not 10 or 13 digits, it was still not the form required. I could add the remaining conditions though. @texodus

packages/perspective-viewer-d3fc/src/ts/legend/filter.ts
onesimus-wiafe: I first wrote the implementation in the `filterData` function and when I realized `filterDataByGroup` would also need it, I made it a function on its own. Because of how filterData was defined, I decided to define the new function similarly so that I don't lose data, but I just noticed I still used `settings.data` in the new function which makes the whole endeavor redundant.

packages/perspective-viewer-d3fc/test/js/axisLabel.spec.ts
onesimus-wiafe: `await page.waitForSelector("perspective-viewer")` was added to obtain the reloaded page. When a test is run without this call, the data array returned to be verified is empty.

packages/perspective-viewer-d3fc/test/js/axisLabel.spec.ts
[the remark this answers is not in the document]
onesimus-wiafe: I will split the current solution into 2, one that confirms that the entire array is of date values, and another that confirms that they are not in epoch form.

texodus (approved):
Looks good!

PR #2873, opened 2024-12-09
Support ndjson

texodus:
* Adds [NDJSON](https://github.com/ndjson/ndjson-spec) support to the `perspective.table()` constructor, via the `format` option.
* Adds `view.to_ndjson()` serialization method.

```javascript
const table = await perspective.table(`{"x": 1}\n{"x": 2}`, {format: "ndjson"});
const view = await table.view();
const ndjson = await view.to_ndjson();
```


timkpaine:
just noting that, annoyingly, `ndjson` and `jsonl` are both widely used names for the same thing, as it has not been standardized.

Issue #2368, opened 2023-09-19
When in grouped state, last group is shown as if it was already unfolded


superrnovae:
### Steps to Reproduce:
 1. Populate the table with some data
 2. Add 2 columns to group by
 3. Double click on the total
 4. Notice last group having a minus sign before it

### Expected Result:
Last group has a plus sign before it, and you have to click once to unfold it

### Actual Result:
Last group has a minus sign, and you have to click twice to unfold it, as for every other subgroup if there are multiple.

### Environment:
Version 2.5.1

### Additional Context:
Screenshot from the example page:
![image](https://github.com/finos/perspective/assets/57127875/256b6aa2-1f8b-406d-8118-0005ecb9cc9a)

ozgurtumer:
last update on this - 

I monkey patched the file 

packages/perspective-viewer-datagrid/src/js/style_handlers/table_cell/row_header.js

 on my local environment with something equivalent to  this (not exactly as the js is compacted)

    const is_collapse_or_undefined =
        next &&
        next.row_header &&
        typeof next.row_header[metadata.row_header_x + 1] !== "undefined";

    const is_collapse = is_collapse_or_undefined ? true : false;

and it seems to fix the issue - I would appreciate one of the maintainers kindly to apply something equivalent to this - many thanks!

PR #2895, opened 2025-01-15
Fix last row label state after collapse and expand

tarekquao:
This fixes the issue found in #2368 where in a grouped state (within Datagrid), the last grouped row header seemed as if it had been already unfolded showing the (-) sign instead of the (+) sign.
Now the last grouped row rightly shows the (+) sign.

### Pull Request Checklist

-   [ ] Description which clearly states what problems the PR solves.
-   [ ] Verify all commits have been _signed_ in accordance with the DCO policy.
-   [ ] Reviewed PR commit history to remove unnecessary changes.

### DIFF
--- a/packages/perspective-viewer-datagrid/src/js/style_handlers/table_cell/row_header.js
+++ b/packages/perspective-viewer-datagrid/src/js/style_handlers/table_cell/row_header.js
@@ -20,7 +20,11 @@ export function cell_style_row_header(regularTable, td, metadata) {
         dx: 0,
         dy: metadata.y - metadata.y0 + 1,
     });
+
+    const is_last_item = !next || next.row_header === undefined;
+
     const is_collapse =
+        !is_last_item &&
         next &&
         next.row_header &&
         typeof next.row_header[metadata.row_header_x + 1] !== "undefined";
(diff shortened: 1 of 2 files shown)


packages/perspective-viewer-datagrid/test/js/row_header.spec.js
```diff
@@ -0,0 +1,46 @@
...
+                plugin: "Datagrid",
+                group_by: ["Region", "Category"],
+                columns: ["Sales", "Quantity", "Discount", "Profit"],
+            });
+        });
+
+        const totalHeader = await page.locator('th:has-text("TOTAL")').first();
+
+        await totalHeader.click();
+        await totalHeader.click();
```
texodus, line 39: You can use [`View::set_depth`](https://docs.rs/perspective-client/3.3.3/perspective_client/struct.View.html#method.set_depth) for this.

packages/perspective-viewer-datagrid/src/js/style_handlers/table_cell/row_header.js
```diff
@@ -20,7 +20,11 @@ export function cell_style_row_header(regularTable, td, metadata) {
         dx: 0,
         dy: metadata.y - metadata.y0 + 1,
     });
+
+    const is_last_item = !next || next.row_header === undefined;
```
texodus, line 24: This only checks whether the row is the last row in the _viewport_, not the dataset. it just makes the last row visible on screen always appear folded.

https://github.com/user-attachments/assets/6da2eedb-a782-4ff6-a916-53fa35a56e4b

texodus:
Thanks for the PR, but I don't think this approach is going to work. I can see two potential approaches:
1) Capture an n + 1 sized window in the `dataListener`, then use this row to determine the collapse state. We'd need to be careful to make sure the extra row doesn't get rendered or we'll end up with the same problem.
2) Add an explcit `View` API to fetch this data, which is beyond the scope of this comment.

PR #2901, opened 2025-01-23
Support for pyodide 0.27.1, including pyarrow + pandas

tomjakubowski:
- Switch to using global `CMAKE_CXX_FLAGS` to ensure every target we
  need to link, including all Arrow components, is built with Emscripten
  exceptions support.  The build already does the same thing for wasm
  exceptions so it's at least more consistent now.
- Add two regression tests to the pytest-pyodide suite for pyarrow and
  pandas examples, which were previously crashing due to inconsistent
  Emscripten exceptions support between the object files

Also includes:
- small devex fix for the setup script when using Pyodide
- splits CI job for Pyodide wheel into build/test, uploads the artifact even when test suite fails

### Pull Request Checklist

        Discussions, this PR applies to. (N/A)

### DIFF
--- a/.github/actions/install-deps/action.yaml
+++ b/.github/actions/install-deps/action.yaml
@@ -197,21 +197,6 @@ runs:
               fi
           if: ${{ runner.os == 'Linux' && inputs.cpp == 'true' && inputs.javascript == 'false' }}
 
-        - name: Install Python Pyodide dependencies
-          shell: bash
-          run: python -m pip install -r rust/perspective-python/requirements-pyodide.txt
-          if: ${{ inputs.pyodide == 'true' }}
-
-        - name: Install Pyodide distribution
-          shell: bash
-          run: pnpm run install_pyodide
-          if: ${{ inputs.pyodide == 'true' }}
-
-        - name: Install Python Playwright browsers
-          shell: bash
-          run: python -m playwright install
-          if: ${{ inputs.pyodide == 'true' }}
-
         # - name: Install CCache
         #   shell: bash
         #   run: sudo apt install -y ccache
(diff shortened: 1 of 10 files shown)


tools/perspective-scripts/setup.mjs
```diff
@@ -168,7 +168,14 @@ async function focus_package() {
             message: "Focus NPM package(s)?",
             default: () => {
                 if (CONFIG["PACKAGE"]) {
-                    return CONFIG["PACKAGE"].split(",");
+                    const packages = CONFIG["PACKAGE"].split(",");
```
tomjakubowski, line 171: to test this, run`pnpm -w run setup`, select perspective-pyodide. then run setup again. perspective-pyodide should be still selected in the menu.

previously the menu selected perspective-python on the second go - which made it easy to accidentally switch back to cpython

tomjakubowski:
I think that's a spurious test failure. I can't restart Actions runs in this repo, but I did do a manual dispatch on my fork: https://github.com/tomjakubowski/perspective/actions/runs/13041122245

.github/actions/install-deps/action.yaml
```diff
@@ -197,21 +197,6 @@ runs:
...
-        - name: Install Pyodide distribution
-          shell: bash
-          run: pnpm run install_pyodide
-          if: ${{ inputs.pyodide == 'true' }}
-
-        - name: Install Python Playwright browsers
-          shell: bash
-          run: python -m playwright install
-          if: ${{ inputs.pyodide == 'true' }}
-
```
texodus, line 214: Why did we move these?

cpp/perspective/CMakeLists.txt
```diff
@@ -338,6 +338,15 @@ if(PSP_PYODIDE)
     string(APPEND CMAKE_CXX_FLAGS "${RELOCATABLE_FLAGS}")
 endif()
 
+if(PSP_PYODIDE)
```
texodus, line 341: I propose we set this for everything.

texodus (approved):

PR #2902, opened 2025-01-23
Add connection error/retry API and UI support

texodus:
This PR adds a reconnect API to Perspective, as well as a UX to invoke a reconnect to `<perspective-viewer>`.


greg-atomic:
Websockets can only handle auth during handshake. And auth tokens typically have an expiration time for security reasons. If a websocket is open for long enough then the token expires. The currently implemented reconnect function doesn't allow for updating the auth details in the new handshake, which happens at reconnect: it sends the old auth data that was used at the initial handshake. So the new reconnect functionality can't be used for this case: instead, one has to explicitly terminate and clean up the websocket connection and establish a brand new one to handle broken connections properly.

I'd suggest to change the reconnect function to enable for an updated URL to reconnect to as a parameter.

texodus:
We would need to see your implementation, you have quite a lot more going on here well beyond the scope of this (merged, 3mo old) PR. `Client` doesn't tick on its own, it is continuing to throw errors because you are still calling methods on it in your application somewhere after calling `Client::terminate` (perhaps on a derived object) - and regardless, creating a new `Client` instance on disconnect (and enough of them to _literally OOM the runtime_ as they are only a few bytes ...) is _not_ the right way to do whatever it is you are trying to do (e.g. I do not understand what you expect to be the result of using the `Client` after explicitly terminating it).

Either way, this PR is not the appropriate venue - please reach out to us at [Prospective](mailto:eric@prospective.co), or you can open an Issue or Discussion if you have a _self-contained, isolated reproduction_ of a specific issue or behavior you think is broken.

PR #2909, opened 2025-01-31
Fix overallocations

timbess:
This fixes a couple of places where we're overallocating.
1. Reserve vocabs while loading arrows into perspective to prevent overgrowth.
2. Free data tables and protobuf messages before processing the table to minimize the concurrent copies of the data in memory.


texodus (approved):

We should consider adding heap benchmarks to our benchmark suite.

PR #2913, opened 2025-02-03
Heap Profiling Tooling

timbess:
Conditionally compile in heap instrumentation to find allocation hotspots.


texodus (approved):
IMO`PSP_HEAP_INSTRUMENTS` should be decoupled from `-g0`. This is helpful but surprising behavior, and it would be the only build flag that supersedes build type for symbols. It sounds like we may benefit from more build levels than `debug` and `release`, maybe we should add support for these explicitly.

texodus:

PR #2918, opened 2025-02-07
JS ProxySession

tomjakubowski:
Add support in perspective-js for creating proxy sessions from a
connected client.


tomjakubowski:
#2929

PR #2920, opened 2025-02-10
Add a stroke-width configuration to lineSeries d3fc charts

sinistersnare:
Hello, I was working on a custom plugin based on perspective-viewer-d3fc and wanted to add a custom stroke width to a lineSeries. This change keeps the previous default value if not provided, so it is completely backwards compatible. 

This PR also includes some fixes to the build scripts. First, it exports typescript declarations during debug builds, and second it puts HTML into the `perspective-viewer-d3fc/dist` directory for the ESM builds, as the CDN builds had the HTML inlined.

### DIFF
--- a/cpp/perspective/CMakeLists.txt
+++ b/cpp/perspective/CMakeLists.txt
@@ -220,6 +220,7 @@ if(PSP_WASM_BUILD)
             -g3 \
             -Wcast-align \
             -Wover-aligned \
+            --emit-tsd=perspective-server.d.ts \
             ")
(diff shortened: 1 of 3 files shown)


tomjakubowski:
Nice catch on the missing .d.ts in debug builds!

texodus:
Thanks for the PR!

Discussed offline. I think this is a good idea generally, adding support for extending the `d3fc` charts in user code, but public APIs needs docs and tests. I've cherry-picked the missing declaration file into #2934

PR #2929, opened 2025-02-21
Python AsyncClient and JS ProxySession

tomjakubowski:
### Pull Request Checklist

* Add support in perspective-js for creating proxy sessions from a
  connected client.
* Support for running maturin commands with uv
  * set `PSP_UV=1` in environment
  * use `uv venv` to create venv, then activate it
  * `uv pip install -r **/requirements.txt` to install requirements
* Update pyo3 from 0.21 to 0.23
  * this was to support use of the `pyo3-async-runtimes` crate
* Build `abi3` .so when building with `maturin develop`
  * **Caveat**: after this change, old non-abi3 .so's still lingering in
    the rust/perspective-python directory may be loaded preferentially
    by Python when using an editable install.  It is best to delete
    those old .so's.
* Repurpose PyClient as AsyncClient and export bindings to Python
* Add async client pytests for both cpython and pyodide
  * Uses `pytest-asyncio` plugin
* Add python and pytest vscode settings
  * Supports running cpython pytest suite from vscode's Test Explorer
* Return an error when `ProxySession::handle_request()` fails to decode
  protobuf, instead of unwrap + panic
 
### DIFF
--- a/rust/perspective-python/src/client/client_async.rs
+++ b/rust/perspective-python/src/client/client_async.rs
@@ -17,14 +17,14 @@ use std::sync::Arc;
 use async_lock::RwLock;
 use futures::FutureExt;
 use perspective_client::{
-    assert_table_api, assert_view_api, clone, Client, OnUpdateMode, OnUpdateOptions, Table,
-    TableData, TableInitOptions, TableReadFormat, UpdateData, UpdateOptions, View,
-    ViewOnUpdateResp, ViewWindow,
+    assert_table_api, assert_view_api, Client, OnUpdateMode, OnUpdateOptions, Table, TableData,
+    TableInitOptions, TableReadFormat, UpdateData, UpdateOptions, View, ViewOnUpdateResp,
+    ViewWindow,
 };
 use pyo3::exceptions::PyValueError;
 use pyo3::prelude::*;
 use pyo3::types::{PyAny, PyBytes, PyDict, PyString};
-use pythonize::depythonize_bound;
+use pythonize::depythonize;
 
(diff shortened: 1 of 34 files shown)


tomjakubowski:
I was seeing some new error spew in the pyodide tests and I think there's a bug in pytest-pyodide, filed here https://github.com/pyodide/pytest-pyodide/issues/148

rust/perspective-python/src/client/client_async.rs
```diff
@@ -33,60 +33,75 @@ use super::update_data::UpdateDataExt;
...
             move |msg| {
-                clone!(handle_request);
+                let handle_request = handle_request.clone();
                 Box::pin(async move {
-                    Python::with_gil(move |py| {
-                        handle_request.call1(py, (PyBytes::new_bound(py, msg),))
-                    })?;
+                    if let Some(fut) = Python::with_gil(move |py| -> PyResult<_> {
+                        let ret = handle_request.call1(py, (PyBytes::new(py, msg),))?;
+                        if isawaitable(ret.bind(py)).unwrap_or(false) {
```
tomjakubowski, line 66: Crucial change here: if calling the Python callback returns a coroutine (i.e. if the callback is a coroutine function), then convert the coro into a Rust future and then await it (line 72)

tomjakubowski:
Back to a draft, need to make a change to JSProxySession

texodus (approved):

The inline and post-hoc comments are super helpful for code review btw 👍 

I tested the emscripten wheel on Jupyterlite with a async client which just sleeps a bit on every msg and everything appears to works as expected.

PR #2943, opened 2025-03-08
Fix segfault in Python webservers

texodus:
This PR fixes a regression in `3.4.0` which causes a panic during the `View.on_update` notification of a non-synchronous `Client`. I've updated the multi threading test to assert single and concurrent thread updates (the old tests dispatched `on_update` notifications to an already-closed event loop that drops errors ...), and I've added a local-loopback websocket client test which replicates the bug in `3.4.0` and passes in this PR.


timkpaine:
Should fix #2945 as well

PR #2952, opened 2025-03-12
Add timeouts to pytest, fix jupyterlab error

tomjakubowski:
We had a pytest which deadlocked in the CI sdist tests. Exception spew from errors which
caused the deadlock was also somehow being swallowed and didn't make it to the log
output in Github Actions.

Add a timeout with pytest-timeout when running the non-pyodide pytest suite, including
for the CI sdist tests. When a test deadlocks, the timeout will fire and pytest-timeout then
prints the deadlocked test's output.

This is a run on a test tag, which shows the error output which was previously being lost in CI:
https://github.com/tomjakubowski/perspective/actions/runs/13820587424/job/38668499105

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -776,7 +776,7 @@ jobs:
               run: python -c 'import perspective'
 
             - name: Run pytests
-              run: pytest -v --pyargs 'perspective.tests'
+              run: pytest -v --pyargs 'perspective.tests' --timeout=300
 
     # ,-,---.             .                 .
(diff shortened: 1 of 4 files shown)


tomjakubowski:
the real test is over here, on my personal fork, where I've created a test tag: https://github.com/tomjakubowski/perspective/actions/runs/13820587424

tomjakubowski:
The [job's log](https://github.com/tomjakubowski/perspective/actions/runs/13820587424/job/38668499105) now includes the spew from the error which caused the deadlock, which is great. But for some reason the job continued to run for several minutes even after pytest appeared to have finished, until I canceled it. Not sure what's going on there.

texodus:

Issue #2949, opened 2025-03-11
Off-by-one-month date filter in `perspective-select` and `perspective-click` events


nburlett:
## Bug Report

Date filters in the `perspective-select` and `perspective-click` events use a zero-indexed month, but the date filter input uses a one-indexed month. The date filter returned by `perspective-config-update` is correctly one-indexed. This issue makes the filtered data in a workspace incorrect. It also makes perspective crash if a January 1 date gets converted to `2025-00-01` and then passed to a second view.

### Steps to Reproduce:

1. Start at https://perspective.finos.org/blocks/superstore/index.html

2. Change the left-hand datagrid to filter on `"Order Date",  "<",  "2025-01-01"`. Here's the json to paste into the Debug JSON field for simplicity: 

```json
{
  "version": "3.4.0",
  "plugin": "Datagrid",
  "plugin_config": {
    "columns": {},
    "edit_mode": "READ_ONLY",
    "scroll_lock": false
  },
  "columns_config": {},
  "settings": true,
  "theme": "Pro Light",
  "title": "Sales Report 2",
  "group_by": [
    "Region",
    "State"
  ],
  "split_by": [
    "Category",
    "Sub-Category"
  ],
  "columns": [
    "Sales",
    "Profit"
  ],
  "filter": [
    [
      "Order Date",
      "<",
      "2025-01-01"
    ]
  ],
  "sort": [],
  "expressions": {},
  "aggregates": {}
}
```

2. Right-click on the heading and select "Create global filter"

3. Open the javascript console

4. Click on any row in the create global filter.

5. Notice the following error in the javascript console:

```
perspective-viewer.js:3232 Uncaught (in promise) Error: Abort(): Invalid date format
    at G.t.wbg.__wbindgen_error_new (perspective-viewer.js:3232:21)
    at 0093b1e2:0x10afc4
    at 0093b1e2:0x179439
    at 0093b1e2:0x521b1
    at 0093b1e2:0x59134
    at 0093b1e2:0x34315
    at 0093b1e2:0x16094f
    at 0093b1e2:0x91ca6
    at 0093b1e2:0xa0d0d
    at 0093b1e2:0xb774d
```

### Expected Result:

The date filter should be one-indexed and the event emitted should be the correct date for passing back into a filter. 

### Actual Result:

Crash due to invalid date if the filtering involves January. Incorrect filtering on the sub-table (off by one month) happens otherwise.

### Environment:

[...]

PR #2957, opened 2025-03-17
Fix date filters in perspective-{select,click} events

nburlett:
Fixes issue #2949 where date filters in the `perspective-select` and `perspective-click` events use a zero-indexed month, but the date filter input uses a one-indexed month. The date filter returned by `perspective-config-update` is correctly one-indexed. This issue makes the filtered data in a workspace incorrect. It also makes perspective crash if a January 1 date gets converted to `2025-00-01` and then passed to a second view.

Fix this by simply adding 1 to the month when generating the config in `ProtoServer::_handle_request`.

### Pull Request Checklist

-   [ ] Description which clearly states what problems the PR solves.
-   [ ] Verify all commits have been _signed_ in accordance with the DCO policy.
-   [ ] Reviewed PR commit history to remove unnecessary changes.


nburlett:
I'm happy to write a regression test, but I'm new to playwright and I'm not sure where to place it. I do not see any existing tests for `perspective-select` or `perspective-click` to use as a starting place.

nburlett:
I figured out how to write a regression test and added it in 0f29a2417b6cc6ae62f2fbe1af9b578197734892. The test fails without 0b2e5dbd026eae9b520c9eab4f586b7426668ad4 and passes with it.

texodus (approved):

The excellent regression test is also much appreciated!

PR #2958, opened 2025-03-19
Invalidate viewer session before loading table

tomjakubowski:
This fixes a bug where a failed `viewer.load()` call would poison the viewer and prevent it from loading any more tables.


texodus (approved):

PR #2964, opened 2025-03-28
Update `limit` behavior to use incrementing `index` (fix `last_by_index` aggregate)

timbess:
This PR decouples primary keys from the limit functionality of perspective tables, allowing us to handle aggregates like "last by index" a bit better when users have a limit and want to rely on insert order to window their data.

Along the way I also fixed some memory safety issues where we had called `new` to allocate some memory and free'd it with `free` rather than `delete`. This wasn't breaking anything at the moment, but may have been relying on undefined behavior accidentally.

The new benchmarks also triggered a bug in our polling implementation where if a table was dirtied and deleted before the `poll` call, it would segfault the poll call because we hadn't removed it from the dirty map in the server, so I fixed that as well.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -212,6 +212,10 @@ jobs:
                       is-release: false
 
         steps:
+            - name: Free up disk space
+              run: |
+                  rm -rf /__t/*
+
             - name: Checkout
               uses: actions/checkout@v4
 
--- a/cpp/perspective/build.js
+++ b/cpp/perspective/build.js
@@ -42,7 +42,7 @@ try {
     execSync(`mkdirp ${cwd}`, { stdio });
     process.env.CLICOLOR_FORCE = 1;
     execSync(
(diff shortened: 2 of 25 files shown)


texodus (approved):

You forgot to mentions his PR also fixes CI (much appreciated!) ... which has been failing due to the GHA CI machines running out of HD space during the Python build.

Test coverage looks great and covers [`last_by_index` aggregate behavior specifically](https://github.com/finos/perspective/pull/2964/files). 

[Benchmarks for feature change to `limit` behavior wrt `Table.update()`](https://github.com/finos/perspective/pull/2964/files#diff-6b2d7bcb8dc584ac694327ef8856161f951e03322f64348b199d1b7afbf7324aR232) are great too. This change does facially increase the allocation size with the new `psp_old_key` column (I think - there is a lot changed including the flatten behavior so I may be wrong here), but its great to see there are no observable impact within the bounds of our CPU time benchmarking. It is also a good reminder IMO we should implement heap benchmarks!

PR #2974, opened 2025-04-07
AsyncTable support in widget

tomjakubowski:
AsyncTable support for PerspectiveWidget

When a `PerspectiveWidget()` is constructed with an `AsyncTable` data
argument, its interface becomes async and methods which operate on the
table return coroutines.

- Add async version of ProxySession.handle_request() with unit test
- Reënable perspective-client unit tests
- New jupyter playwright tests for widget + AsyncTable

### DIFF
--- a/packages/perspective-jupyterlab/README.md
+++ b/packages/perspective-jupyterlab/README.md
@@ -1,6 +1,7 @@
 # Perspective JupyterLab Extension
 
-This extension allows in-lining perspective based charts in jupyterlab notebooks.
+This extension allows in-lining perspective based charts in jupyterlab
+notebooks.
 
 [Examples](https://github.com/finos/perspective/tree/master/examples/jupyter-notebooks)
 
--- a/packages/perspective-jupyterlab/src/js/view.js
+++ b/packages/perspective-jupyterlab/src/js/view.js
@@ -180,7 +180,6 @@ export class PerspectiveView extends DOMWidgetView {
     }
 
     get psp_client_id() {
-        // XXX(tom): why can't we just use the GUID already used to uniquely identify the view in comms ?
         return this.#psp_client_id;
     }
 
--- a/packages/perspective-jupyterlab/test/jupyter/widget.spec.mts
+++ b/packages/perspective-jupyterlab/test/jupyter/widget.spec.mts
@@ -110,6 +110,65 @@ describe_jupyter(
             }
         );
 
+        test_jupyter(
+            "Loads AsyncTable",
+            [
+                `
+server = perspective.Server()
+sync_client = server.new_local_client()
+sync_client.table({"Income": [5,4,3,2,1], "Expense": [4,3,2,1,1], "Profit": [1,1,1,1,0]}, name="Microstore")
+proxy_sess = perspective.ProxySession(sync_client, lambda msg: asyncio.create_task(async_client.handle_response(msg)))
+
+async_client = perspective.AsyncClient(proxy_sess.handle_request_async)
+async_table = await async_client.open_table("Microstore")`,
+                "w = perspective.widget.PerspectiveWidget(async_table)",
+                "w",
+            ],
+            async ({ page }) => {
+                await default_body(page);
+                const num_columns = await page
+                    .locator("regular-table thead tr")
+                    .first()
+                    .evaluate((tr) => tr.childElementCount);
+
+                expect(num_columns).toEqual(3);
+                await expect(
+                    page.locator("regular-table tbody tr")
+                ).toHaveCount(5);
+            }
+        );
+
+        test_jupyter(
+            "Loads updates to AsyncTable",
+            [
+                [
+                    `
+server = perspective.Server()
+sync_client = server.new_local_client()
+sync_table = sync_client.table(arrow_data)
+proxy_sess = perspective.ProxySession(sync_client, lambda msg: asyncio.create_task(async_client.handle_response(msg)))
+
+async_client = perspective.AsyncClient(proxy_sess.handle_request_async)
+async_table = await async_client.open_table(sync_table.get_name())`,
+                    "w = perspective.widget.PerspectiveWidget(async_table, columns=['f64', 'str', 'datetime'])",
+                ].join("\n"),
+                "w",
+                "sync_table.update(arrow_data)",
+            ],
+            async ({ page }) => {
+                await default_body(page);
+                const num_columns = await page
+                    .locator("regular-table thead tr")
+                    .first()
+                    .evaluate((tr) => tr.childElementCount);
+
+                expect(num_columns).toEqual(3);
+
+                await expect(
+                    page.locator("regular-table tbody tr")
+                ).toHaveCount(10);
+            }
+        );
         // Restore
 
         test_jupyter(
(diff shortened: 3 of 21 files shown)


rust/perspective-python/perspective/widget/viewer/viewer.py
```diff
@@ -165,33 +165,53 @@ def load(self, data, **options):
         sort, etc.) is cleared to prevent applying settings on columns that
         don't exist.
 
+        When a ``perspective.AsyncTable`` is loaded, the widget's interface
+        becomes async. Methods which operate on the underlying Perspective
+        view, inclusive of the ``load()`` call itself, return coroutine values
+        which must be awaited.
```
tomjakubowski, line 171: This presto-change-o magic API that switches between sync/async might be a little too weird. I don't know much about Python typing but it seems like it could be hard to write good types for this. Maybe mypy supports something like type state?

packages/perspective-jupyterlab/README.md
```diff
@@ -18,3 +19,19 @@ jupyter labextension install @finos/perspective-jupyterlab
 ```bash
 pip install perspective-python
 ```
+
+## Build
```
tomjakubowski, line 23: this is not exactly suited to be distributed with the package README, can move this to a DEVELOPMENT.md or similar

packages/perspective-jupyterlab/src/js/view.js
```diff
@@ -104,8 +104,8 @@ export class PerspectiveView extends DOMWidgetView {
                     }
                 });
 
-            this.luminoWidget.load(table);
-            this._restore_from_model();
+            await this.luminoWidget.load(table);
```
tomjakubowski, line 107: I don't think this change was necessary, I'll undo it and test it

rust/perspective-client/Cargo.toml
```diff
@@ -76,3 +76,6 @@ features = ["prost-derive", "std"]
 [dependencies.ts-rs]
 version = "10.0.0"
 features = ["serde-json-impl", "no-serde-warnings"]
+
+[dev-dependencies]
+tokio = { version = "~1", features = ["full"] }
```
tomjakubowski, line 81: oops, left over from some tests I wrote (which were just to help me understand better how ProxySession works)

rust/perspective-python/perspective/widget/__init__.py
```diff
@@ -10,11 +10,13 @@
 #  ┃ of the [Apache License 2.0](https://www.apache.org/licenses/LICENSE-2.0). ┃
 #  ┗━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┛
 
+import asyncio
```
tomjakubowski, line 13: can delete (import is scoped below at `asyncio.create_task()` call)

rust/perspective-python/perspective/widget/__init__.py
```diff
@@ -201,18 +210,19 @@ def handle_message(self, widget, content, buffers):
...
+            self._sessions[client_id] = self.new_proxy_session(send_response)
         elif content["type"] == "binary_msg":
             [binary_msg] = buffers
             client_id = content["client_id"]
             session = self._sessions[client_id]
-            logging.debug("view {} message {}", client_id, len(binary_msg))
             if session is not None:
-                session.handle_request(binary_msg)
+                import asyncio
+                asyncio.create_task(session.handle_request_async(binary_msg))
```
tomjakubowski, line 225: I believe it's the case that ipykernel typically runs python code [inside an event loop](https://blog.jupyter.org/ipython-7-0-async-repl-a35ce050f7f7?gi=faf99afc522d), so this should be OK to do whether or not the underlying table is an AsyncTable or (sync) Table.

This would break in a custom Python kernel which doesn't run an asyncio event loop. I could implement a fallback in case one is not running, but I don't know a good way offhand to set up an environment to test that.

tools/perspective-scripts/sh.mjs
```diff
@@ -323,7 +323,7 @@ function parse_bind_args(fragments, ...args) {
     for (let i = 0; i < fragments.length - 1; i++) {
         const arg = args[i];
         const start = terms.length === 0 ? fragments[i] : terms.pop();
-        if (arg === undefined || arg !== arg || arg === false) {
+        if (arg === undefined || Number.isNaN(arg) || arg === false) {
```
tomjakubowski, line 326: Unrelated change, I was poking around in here and was a little bit surprised by seeing `x !== x`. I think `Number.isNan()` does the same thing, although to be fair I haven't seen the Wat talk in a minute.

tools/perspective-scripts/test_js.mjs
```diff
@@ -104,7 +104,9 @@ if (process.env.PACKAGE) {
         process.env.PACKAGE.indexOf("perspective-python") >= 0 &&
         process.env.PACKAGE.indexOf("!perspective-python") === -1
     ) {
-        sh`pnpm run --recursive --filter @finos/perspective-python test`.runSync();
+        // Support `pnpm test -- --my_cool --test_arguments`
```
tomjakubowski, line 107: new feature -- I wanted to be able to thread pytest arguments down through a top-level `pnpm run test`

tools/perspective-scripts/test_js.mjs
```diff
@@ -137,7 +139,7 @@ if (process.env.PACKAGE) {
             target = "--target=aarch64-unknown-linux-gnu";
         }
 
-        sh`cargo test ${flags} ${target} -p perspective`.runSync();
+        sh`cargo test ${flags} ${target} -p perspective -p perspective-client`.runSync();
```
tomjakubowski, line 142: There are some unit tests in `perspective-client`'s clone module which I think were not running before. I re-enabled them here.

tools/perspective-test/playwright.config.ts
```diff
@@ -226,7 +226,7 @@ export default defineConfig({
     snapshotPathTemplate:
         "dist/snapshots/{projectName}/{testFilePath}/{arg}{ext}",
     webServer: {
-        command: "node --loader ts-node/esm src/js/start_test_server.ts",
+        command: "tsx src/js/start_test_server.ts",
```
tomjakubowski, line 229: When I was adjusting the playwright config, `node --loader` was driving me a little crazy with obscure errors. tsx was already in the dependency list and it worked a lot better for me.

packages/perspective-jupyterlab/test/jupyter/widget.spec.mts
```diff
@@ -110,6 +110,65 @@ describe_jupyter(
...
+
+                expect(num_columns).toEqual(3);
+                await expect(
+                    page.locator("regular-table tbody tr")
+                ).toHaveCount(5);
+            }
+        );
+
+        test_jupyter(
+            "Loads updates to AsyncTable",
```
tomjakubowski, line 142: it would be nice to add test coverage for the async `.replace()`, `.delete()` calls too

texodus: Also `.remove()`

texodus (approved):

PR #2987, opened 2025-04-29
Nightly artifacts

timbess:
This modifies the CI to build production releases on the `master` branch whenever a PR is merged. Previously, the same _dev_ CI workflow was run on `master` branch merges, and only release tags built the full production workflow, which includes MacOS & Windows python wheels and benchmarks among other tweaks.

As a result, we'll now have apples-to-apples intermediate builds between releases, which will halpe keep regressions out of released perspective wheels/crates/packages.


texodus (approved):

PR #3007, opened 2025-06-13
Add a failing test for updating a table with a headerless CSV.

sinistersnare:
This PR adds a test that crashes the test suite. It does not provide a fix, so the PR is marked as a draft for now.

Currently Perspective aborts in the C++ by updating a table with a CSV that has no header.

```py
tbl = Table({ "stringcol": "string" })
tbl.update("rowval  \n another row val")
```

Results in `Fatal Python error: Aborted`

This bug also occurs in Rust, see the example in the below comment

### DIFF
--- a/rust/perspective-python/perspective/tests/table/test_exception.py
+++ b/rust/perspective-python/perspective/tests/table/test_exception.py
@@ -63,3 +63,13 @@ class TestException(object):
             tbl.view(group_by=["b"])
 
         assert str(ex.value) == "Abort(): Invalid column 'b' found in View group_by.\n"
+
+    def test_exception_from_csv_update_with_no_header(self):
+        tbl = Table({ "stringcol": "string" })
+        data = "rowval  \n another row val"
+        errored = False
+        try:
+            tbl.update(data)
+        except:
+            errored = True
+        assert errored


sinistersnare:
Here is the Rust example:

```rs
use std::error::Error;

use perspective::client::{ColumnType, TableData, UpdateData};
use perspective::server::Server;

#[tokio::main(flavor = "multi_thread")]
async fn main() -> Result<(), Box<dyn Error>> {
    let server = Server::default();
    let client = server.new_local_client();
    let schema = TableData::Schema(vec![("stringcol".into(), ColumnType::String)]);
    let table = client.table(schema, Default::default()).await?;
    // Note: No header row!
    let update = UpdateData::Csv("A\nB\n".to_string());
    table.update(update, Default::default()).await?;
    let csv = table.view(None).await?.to_csv(Default::default()).await?;
    println!("CSV:: {csv:?}");
    client.close().await;
    Ok(())
}
```

```sh
$ cargo run
   Compiling example_repo v1.0.0 (/Path/To/Example)
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 1.53s
     Running `target/debug/example_repo`
libc++abi: terminating due to uncaught exception of type std::invalid_argument: stoll: no conversion
[1]    19940 abort      cargo run
```

PR #3044, opened 2025-07-29
PerspectiveWorkspace React component

sinistersnare:
Hello! This PR adds onto our `perspective-react` package an API for the `<perspective-workspace>` custom element in the form of a `<PerspectiveWorkspace />` React component. You can use it as such:

```jsx
import * as Workspace from "@finos/perspective-workspace";
const tables = {a: client.table("a,b,c\n1,2,3\n4,5,6", { name: "a" })};
const base = { sizes: [1], detail: { main: null }, viewers: {});
const id = Workspace.genId(base);
const layout = Workspace.addViewer(base, { table: "a", title: "View A!"}, id);
<PerspectiveWorkspace tables={tables} layout={layout} />
```

You can see that this PR also adds some free functions (namely `addViewer` and `genId`) to assist in the functional style of React.

There are also callbacks that can be passed for the `on-layout-update` and `on-new-view` events.

The `PerspectiveWorkspaceConfig` interface was updated to better reflect the reality of the current workspace structure.

During the course of this, I encountered a bug of an internal engine message not being swallowed by the engine. That bug is tracked in #3039.


packages/perspective-workspace/src/ts/workspace/workspace.ts
```diff
@@ -112,6 +236,10 @@ export class PerspectiveWorkspace extends SplitPanel {
     }
 
     init_indicator() {
+        const exists = document.querySelector("body > perspective-indicator");
+        if (exists) {
+            return exists as HTMLElement;
+        }
```
texodus, line 242: This smells like a previous `<perspective-workspace>` that was disconnected leaked this element, not the responsibility of this method to correct.

packages/perspective-workspace/src/ts/workspace/workspace.ts
```diff
@@ -32,21 +32,145 @@ const DEFAULT_WORKSPACE_SIZE = [1, 3];
...
+          currentIndex: number;
+          widgets: string[];
+      };
 
 export interface ViewerConfigUpdateExt extends ViewerConfigUpdate {
     table: string;
 }
 
-export interface PerspectiveWorkspaceConfig<T> {
+// TODO: I have made some changes to this interface because it seemed to not reflect reality.
```
texodus, line 52: Lumino's `DockPanel.saveLayout()` does not return string identifiers for `widgets` field, ergo the types for e.g. `workspace.save()` is now not `PerspectiveWorkspaceConfig` (it's an anonymous struct type). Add the explicit returns types to `save()` will show this error.

packages/perspective-workspace/src/ts/workspace/workspace.ts
```diff
@@ -1054,4 +1188,29 @@ export class PerspectiveWorkspace extends SplitPanel {
...
+            "perspective-viewer"
+        ) as NodeListOf<HTMLPerspectiveViewerElement>;
+        // eject all viewers that have a table not in the new set of tables.
+        for (const viewer of viewers) {
+            if (!this._last_updated_state) break;
+            const t = this._last_updated_state.viewers[viewer.slot]?.table;
+            if (!newTables[t]) {
+                await viewer.eject();
+            } else {
+                await viewer.load(newTables[t]);
```
texodus, line 1207: Please batch these and await them all at the end using `Promise.all()`, so the user does not have to watch a slide show of these elements updating.

PR #3048, opened 2025-08-28
Fix Workspace.removeTable()

tomjakubowski:
A recent change to the workspace/viewer error handling resulted in the workspace
removeTable() method always failing.  This fixes the bug and adds a smoke/regression
test.

Also includes a Typescript fix for types in the custom elements registry.

### DIFF
--- a/packages/perspective-workspace/src/ts/utils/observable_map.ts
+++ b/packages/perspective-workspace/src/ts/utils/observable_map.ts
@@ -25,12 +25,8 @@ export class ObservableMap<K, V> extends Map<K, V> {
     }
 
     delete(name: K) {
-        const result = this._delete_listener?.(name);
-        if (result) {
-            return super.delete(name);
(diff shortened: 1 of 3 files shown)


packages/perspective-workspace/src/ts/utils/observable_map.ts
```diff
@@ -25,12 +25,8 @@ export class ObservableMap<K, V> extends Map<K, V> {
     }
 
     delete(name: K) {
-        const result = this._delete_listener?.(name);
```
tomjakubowski, line 28: The delete listener registered by the workspace always returns `undefined` now, so this ended up always taking the `else` branch below, which skips removing the key from the map.  It looks like the delete listener is not meant anymore to prevent deletes, so I removed the conditional entirely.

texodus (approved):

PR #3057, opened 2025-10-06
Provide perspective token in JupyterLab to ensure proper ordering of other extensions

### LINKED ISSUE #3056: Perspective JupyterLab plugin cannot be used with other plugins that use perspective
Due to the initialization of custom components:

Image

This causes the jupyter widget initialization to fail, which results in it not being displayable:

Image

This appears to be unavoidable as an import of `@finos/perspective-viewer` is now required, so it is not an option to not import perspective:

```javascript
import perspective from "@finos/perspective";
// triggers component registration
import perspective_viewer from "@finos/perspective-viewer";
import SERVER_WASM from "@finos/perspective/dist/wasm/perspective-server.wasm";
import CLIENT_WASM from "@finos/perspective-viewer/dist/wasm/perspective-viewer.wasm";

// Required, otherwise perspective.table will not work
await Promise.all([perspective.init_server(fetch(SERVER_WASM)), perspective_viewer.init_client(fetch(CLIENT_WASM))]);
```

To get my extension working, I was able to make the following change in the minified assets, which shows the problematic parts:

Image

problem 1: https://github.com/finos/perspective/blob/a074bed2908964fbbb09f301d4a1b7157581c3c0/rust/perspective-viewer/src/rust/utils/custom_element.rs#L70
problem 2: https://github.com/finos/perspective/blob/a074bed2908964fbbb09f301d4a1b7157581c3c0/packages/perspective-viewer-datagrid/src/js/index.js#L38
problem 3:
https://github.com/finos/perspective/blob/a074bed2908964fbbb09f301d4a1b7157581c3c0/packages/perspective-viewer-d3fc/src/ts/plugin/plugin.ts#L88

timkpaine:
JupyterLab has a [mechanism for having plugins depend on other plugins](https://jupyterlab.readthedocs.io/en/stable/extension/extension_dev.html#plugins-interacting-with-each-other).

Since perspective registers custom elements (see #3056), this change lets other Jupyter extensions "depend" on `@finos/perspective-jupyterlab` by doing:

```javascript
import {IPerspective} from "@finos/perspective-jupyterlab";

export const MyCoolPlugin = {
  activate,
  id: "my-cool-plugin",
  requires: [IPerspective],
  autoStart: true,
};
```

This then ensures that `@finos/perspective-jupyterlab` gets activated before `MyCoolPlugin`, ensuring that custom webcomponents and wasm loading occurs. Downstream plugins like `MyCoolPlugin` can then just do:

```javascript
import perspective from "@finos/perspective";
```

without any additional initialization. 

Fixes #3056 

I'm open to adding tests but its just a constant, not sure what the best mechanism for doing so is.
~~We can also add developer docs if we want, happy to write them.~~ done

### Pull Request Checklist

### DIFF
--- a/packages/perspective-jupyterlab/package.json
+++ b/packages/perspective-jupyterlab/package.json
@@ -55,6 +55,10 @@
                 "bundled": true,
                 "singleton": true
             },
+            "@finos/perspective-jupyterlab": {
+                "bundled": true,
+                "singleton": true
+            },
             "@finos/perspective-viewer": {
                 "bundled": true,
                 "singleton": true
@@ -64,6 +68,10 @@
                 "singleton": true
             },
             "@finos/perspective-viewer-datagrid": {
+                "bundled": true,
+                "singleton": true
+            },
+            "@finos/perspective-viewer-openlayers": {
                 "bundled": true,
                 "singleton": true
             }
--- a/packages/perspective-jupyterlab/src/js/plugin.js
+++ b/packages/perspective-jupyterlab/src/js/plugin.js
@@ -11,12 +11,18 @@
 // ┗━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┛
 
 import { IJupyterWidgetRegistry } from "@jupyter-widgets/base";
-import { Token } from "@lumino/coreutils";
 import { PerspectiveModel } from "./model";
+import {
+    IPerspective,
+    IPerspectiveJupyterlab,
+    IPerspectiveViewer,
+    IPerspectiveViewerD3fc,
+    IPerspectiveViewerDatagrid,
+    IPerspectiveViewerOpenlayers,
+} from "./tokens";
 import { PerspectiveView } from "./view";
 import { PERSPECTIVE_VERSION } from "./version";
 const EXTENSION_ID = "@finos/perspective-jupyterlab";
-export const IPerspective = new Token(EXTENSION_ID);
 
 /**
  * PerspectiveJupyterPlugin Defines the Jupyterlab plugin, and registers `PerspectiveModel` and `PerspectiveView`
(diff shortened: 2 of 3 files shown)


texodus:
> Fixes https://github.com/finos/perspective/issues/3056

This is not a solution to this problem - you are abusing an empty service definition so that you can stack side effects (sans tests). The issue here is shared _dependencies_, not shared plugin state/services, for which Jupyter prescribes [Webpack Module Federation](https://github.com/jupyterlab/jupyterlab/tree/main/examples/federated) as a mechanism for bundling shareable JS modules between plugins. This is the _only_ solution which allows:

* Two (or N) plugins which use `@finos/perspective` and/or `@finos/perspective-viewer` to work when installed concurrently, _and_
* Two (or N) plugins to _also_ work when installed separately.

I'm not against accepting a patchwork solution to cross a real functionality gap where we're actually blocked upstream, but we haven't even tried to fix this correctly before cramming in another hack. Show me that module federation does not work first (and then add tests, presuming there is a second step at all).

timkpaine:
> > We can suggest info as well for the plugins to use the dependencies:
> 
> That's module federation, like I said above (and also not in this PR). If you can get it to work, I will merge it, but just adding those annotations to the `@finos/perspective-jupyterlab` webpack config does not (I've tested)

This is done in the consuming plugin, not in the producing plugin though: https://jupyterlab.readthedocs.io/en/stable/extension/extension_dev.html#requiring-a-service

I don't think the thing you're looking for exists in the way you expect it to exist.

texodus:
> Will see if there is a different approach available, though there aren't any examples of it in the ecosystem that I can find.

[This is the example (again)](https://github.com/jupyterlab/jupyterlab/tree/main/examples/federated). Implement this and add a multi-plugin test and re-open please.

timkpaine:
> [This is the example (again)](https://github.com/jupyterlab/jupyterlab/tree/main/examples/federated). Implement this and add a multi-plugin test and re-open please.

I am doing what that example does.

PR #3058, opened 2025-10-07
Upgrade to `emsdk` 4.0.9

texodus:
This PR is not ready to merge.

In order to build for `pyodide`, we must use the same `emscripten` ABI version as they do. This has recently been upgraded to `4.0.9` (from `3.1.58`), but the pyodide version which supports this ABI (`0.28.x`) deprecates `pyarrow` and `polars` (or rather - decouples their compatibility from the pyodide release itself). Ergo, while this PR can compile, we cannot test it yet.


texodus:
Subsumed by #3118

PR #3060, opened 2025-10-07
Provide perspective token in JupyterLab to ensure proper ordering of other extensions

### LINKED ISSUE #3056: Perspective JupyterLab plugin cannot be used with other plugins that use perspective

timkpaine:
JupyterLab has a [mechanism for having plugins depend on other plugins](https://jupyterlab.readthedocs.io/en/stable/extension/extension_dev.html#plugins-interacting-with-each-other).

Since perspective registers custom elements (see #3056), this change lets other Jupyter extensions "depend" on `@finos/perspective-jupyterlab` by doing:

```javascript
import {IPerspective} from "@finos/perspective-jupyterlab";

export const MyCoolPlugin = {
  activate,
  id: "my-cool-plugin",
  requires: [IPerspective],
  autoStart: true,
};
```

This then ensures that `@finos/perspective-jupyterlab` gets activated before `MyCoolPlugin`, ensuring that custom webcomponents and wasm loading occurs. Downstream plugins like `MyCoolPlugin` can then just do:

```javascript
import perspective from "@finos/perspective";
```

without any additional initialization. 

Fixes #3056 

I'm open to adding tests but its just a constant, not sure what the best mechanism for doing so is.
~~We can also add developer docs if we want, happy to write them.~~ done

### Pull Request Checklist

### DIFF
--- a/docs/md/how_to/python/jupyterlab.md
+++ b/docs/md/how_to/python/jupyterlab.md
@@ -59,3 +59,27 @@ Perspective also exposes a JS-only `mimerender-extension`. This lets you view
 this by right clicking one of these files and `Open With->CSVPerspective` (or
 `JSONPerspective` or `ArrowPerspective`). Perspective will also install itself
 as the default handler for opening `.arrow` files. -->
+
+## Depending on Perspective in your own JupyterLab Widget
+
+Perspective provides a [token for integration with JupyterLab's federated dependency model](https://jupyterlab.readthedocs.io/en/stable/extension/extension_dev.html#plugins-interacting-with-each-other).
+This allows your plugin to simply depend on Perspective for initialization.
+
+In your plugin code:
+
+```javascript
+import {IPerspective} from "@finos/perspective-jupyterlab";
+
+export const MyCoolPlugin = {
+  activate,
+  id: "my-cool-plugin",
+  requires: [IPerspective],
+  autoStart: true,
+};
+
+// to use perspective, simply import. No initialization required
+import perspective from "@finos/perspective";
+```
+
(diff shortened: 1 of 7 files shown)


timkpaine:
was supposed to open on fork

PR #3084, opened 2025-11-17
Support a wider array of integers

sinistersnare:
Currently arrow files may be sent to Perspective that use a variety of integer types that are not supported internally by Perspective. This PR allows these types by properly handling them on the server-side.

Look at the provided test to generate the appropriate arrow that fails on `master` but passes on this branch.


timbess (approved):
Look good!

Issue #3091, opened 2025-12-03
Help to provide linux aarch64 wheels


Oblynx:
### Description of Problem:

Like #2221, I want to provide linux aarch64 wheels.

Since that issue was opened, github now provides standard arm64 runners: `ubuntu-22.04-arm`.
Would you like a PR to add them to your CI?

PR #3092, opened 2025-12-03
Build arm64/aarch64 Linux wheels

tomjakubowski:
Add aarch64 Linux wheel to the build, and include it in the release artifacts.

Combined effort with @Oblynx from #3093.

Also:

- Add a simple DCO check to the pre-push hook. This is not quite 100% foolproof, as it only checks HEAD, but can fix in a follow up.
- Fix an issue in the build script where we were both building protoc from source and also downloading binaries to be installed.
- Add some `df -h` invocations to the Python job to monitor disk free space.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -220,10 +220,18 @@ jobs:
                       arch: x86_64
 
         steps:
+            - name: Run df -h
+              if: ${{ runner.os == 'Linux' }}
+              run: df -h
+
             - name: Free up disk space
               if: ${{ runner.os == 'Linux' }}
               run: |
                   rm -rf /__t/*
+
+            - name: Run df -h
+              if: ${{ runner.os == 'Linux' }}
+              run: df -h
 
             - name: Checkout
               uses: actions/checkout@v4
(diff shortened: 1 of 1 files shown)


tomjakubowski:
Public note to self: I think maybe the action we use to install cmake is installing an x86-64 build of cmake on the arm runner. Will pick up on this thread next.

Screenshot 2025-12-04 at 5 23 04 PM

Oblynx:
Hey @tomjakubowski thanks for working on this! I don't know if you were influenced by my issue #3091 ?

I ran into the same problems as you and propose solutions in #3093.

texodus (approved):

I have run a [full build on this branch](https://github.com/perspective-dev/perspective/actions/runs/20088683078) and it passed ... on the second try, failing the first due to disk space on the x64 python build step. I am a bit concerned we are so close to the disk space limit for these machines, this failure is going to be an ongoing problem for CI builds. Flattening out the `protoc` duplication is a great step!

Oblynx:
@texodus can we release the new wheels? Thanks for merging!

PR #3116, opened 2026-02-04
Better TypeScript types for `viewer-datagrid`

texodus:
This PR updates `viewer-datagrid` to use new TypeScript types exporte din `regular-table@0.8.1`, and fixes a flapping test.


packages/viewer-datagrid/src/ts/model/create.ts
```diff
@@ -222,7 +215,7 @@ export async function createModel(
         }),
         _series_color_map: new Map<string, string>(),
         _series_color_seed: new Map<string, number>(),
-        get_psp_type,
+        // get_psp_type,
```
sinistersnare: dead code

packages/viewer-datagrid/src/ts/style_handlers/table_cell/datetime.ts
```diff
@@ -38,6 +34,7 @@ export function cell_style_datetime(
 
     const [hex, r, g, b] = colorRecord;
 
+    // @ts-ignore
```
sinistersnare: This pattern is used elsewhere, maybe we shoul dkeep the `WithFlags` bit

texodus: TypeScript can't extend union types.

PR #3160, opened 2026-04-15
Add WebGL chart plugin with scatter, line, and treemap

timbess:
New viewer-webgl package that renders perspective views on the GPU. Three chart types (scatter, line, treemap), each with custom GLSL shaders, chunked Arrow streaming, and a shared buffer pool for GPU memory reuse. Handles axis/legend/tick layout, gridline rendering, and zoom/pan interaction. Hooks into the viewer's streaming draw lifecycle. Adds the webgl-example app and registers the package in the workspace and setup script.

Issue #2938, opened 2025-03-01
Mixed-type charts to display multiple series types on the same chart


GuntuAshok:
### Description of Problem:

Currently there is no option to have mixed charts like the bar chart for primary Y-axis and a line chart for secondary Y-axis.

### Potential Solutions:

I found the support of mixed chart type in d3fc examples here

https://d3fc.io/examples/series-canvas-multi/

https://github.com/d3fc/d3fc/tree/master/examples/series-canvas-multi/

Mixed chart option really helps to compare 2 different entities on same chart with distinction.

texodus:
All `viewer-charts` types support mixed glyphs (#3178). There are much more complex scenarios to consider here e.g. overlapping columns from separate Views or Tables, but this issue doesn't really capture those, so closing.

PR #3166, opened 2026-04-23
Rename `@perspective-dev/viewer-charts`, add missing chart types

texodus:
This PR introduces `@perspective-dev/viewer-charts` (ported from #3160, `@perspective-dev/viewer-webgl`), and deletes (but does not yet deprecate) `@perspective-dev/viewer-d3fc`. This is a breaking change, upgrading to `v4.5.0` will require using `@perspective-dev/charts` (this package has a compatible superset of its predecessor's API, though it has a new name and internals).

In addition, the new `@perspective-dev/viewer-charts` packages has been upgraded to mostly feature parity with the package it replaces. This includes:

- Adds nine new chart types in addition to the original three: `X/Y Scatter`, `X/Y Line`, `Y Bar`, `X Bar`, `Y Line`, `Y Scatter`, `Y Area`, `Treemap`, `Sunburst`, `Candlestick`, `OHLC`, `Heatmap`.
- Significant refactor internally - drops streaming API in favor of the new `with_typed_arrays`. While streaming is better for interactive performance of one large dataset, it is worse for throttling streaming udpates. I've left the batching structure of WebGL rendering in, but the Viewer API which supported streaming has been removed entirely.
- Visual regression tests using the `swiftShader` backend for `playwright`. We have tried this approach before (with HighCharts) and it was miserable, but this seems to works so far out-of-the-box without any nonsense.

Not featured in the old `@perspective-dev/viewer-d3fc`, the new `@perspective-dev/viewer-charts` has some improvements:

- Per-column `chart_type` ∈ `{Bar, Line, Scatter, Area}` and optional `stack` override. Flattened into `columns_config`, now enabled for Y-type charts; defaults (`Bar` + no stack override). #2938 #2390 #2331
- **All** charts types now support locked/unlocked zooming (even categorical axes) as well as auto-fit (though this is not fully enabled in the UI yet, only via `plugin-config`).
- **All** chart types are now super fast. So fast, the Render Warning thresholds have all been increased to 1,000,000 rows.


MPizzotti:
very dumb question on this: can we include apache-echarts / custom vis libraries are still supported with the current implementation ?

PR #3180, opened 2026-05-25
Add `gmv` aggregate

texodus:
This PR adds a new aggregate `gmv` ("gross market value"), which for any node in a Perspective `group_by` / `split_by` tree:

* Leaf (full row depth): plain signed `sum`.
* Parent: Σ over immediate row children of `|sum(child_subtree)|`.

The parent uses each child's raw signed sum, not the child's already-rolled-up `gmv` value. This makes parent elements `≠ Σ|child_gmv|`.

This specialized aggregation will be noticeably slower than regular `sum` due to recursively recalculating `gstate` for each depth, so use it wisely.


nmichaud:
GMV = sum(abs(MV)). It's never negative. Rolling it up can be tricky depending on how you want to treat positions on the same names across multiple subgroups (ie net them or treat as separate).

texodus:
@nmichaud This implementation nets by the _last_ `group_by` column, for all ancestor groups/splits, which is why it is slow (it is re-grouping for every ancestor level). E.g., the net column in the [test case](https://github.com/perspective-dev/perspective/pull/3180/changes#diff-d7bd733bb6f1acb1e23280d422769b7bdc089c68185798de2b7a38a56976c9a7R427) is `"symbol"`.

The last column's aggregate is `"sum"` instead of `"sum abs"`, because this implementation is from [a spec](https://github.com/perspective-dev/perspective/pull/3180/changes#diff-d7bd733bb6f1acb1e23280d422769b7bdc089c68185798de2b7a38a56976c9a7R427). You can create an `expressions` column using the `abs()` function to calculate this:

```
{
    "expressions": {
        "abs(MV)": "abs(\"MV\")"
    },
    "aggregates": {
        "abs(MV)": "gmv"
    }
}
```

... or y'all can change the spec (@timkpaine).

nmichaud:
Ok that makes sense. I wouldn't call the agg GMV, at least without calling out that you need to use it on an abs(x) column.

PR #3183, opened 2026-06-02
Fix `mallinfo` use

abalabin-bamfunds:
fix the use of `mallinfo` function:

- on Linux use `mallinfo2` as `mallinfo` is deprecated and buggy as per https://www.man7.org/linux/man-pages/man3/mallinfo.3.html - the deprecation warning is gone from the build after the change
- in wasm use mallinfo as there is no mallinfo2 implementation available; however work around the bug outlined on the same page above to do with the returned datastructure declaring fields as signed while in the implementation they are not hence producing 'wrap around zero' bugs - the sysinfo call can now report use of more than 2GB of memory by wasm (this is probably not worth building a unit test around though)

### DIFF
--- a/rust/perspective-server/cpp/perspective/src/cpp/server.cpp
+++ b/rust/perspective-server/cpp/perspective/src/cpp/server.cpp
@@ -40,6 +40,7 @@
 
 #if defined(PSP_ENABLE_WASM) || defined(__linux__)
 #include <malloc.h>
+#include <dlfcn.h>
 #endif
 
 #if !defined(WIN32) && !defined(PSP_ENABLE_WASM)
@@ -820,7 +821,7 @@ ProtoServer::handle_request(
 ) {
     const auto start = std::chrono::high_resolution_clock::now();
     proto::Request req_env;
-    req_env.ParseFromString(data);
+    req_env.ParseFromArray(data.data(), data.size());
     std::vector<ProtoServerResp<std::string>> serialized_responses;
     std::vector<proto::Response> responses;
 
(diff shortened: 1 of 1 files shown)


texodus (approved):
Thanks for the PR! ~Looks good!~

I should have waited for CI to finish before reviewing this!

texodus (changes requested):
This change fails to build in CI on GA's Ubuntu 22.04 x86_64 and aarch64 targets, likely due to the [Manylinux docker environment](https://github.com/pypa/manylinux) these builds run in.

I will merge this if it can be gated to the correct glibc to pass CI, but this means it will not end up in generic wheel builds currently anyway, only when building Perspective from source.

tomjakubowski:
To support both mallinfo and mallinfo2 in a single binary, maybe you could use dlsym() to call mallinfo2 when it's available, and otherwise fall back to mallinfo. It would require inlining the definition of the mallinfo2 struct since the header won't define it for glibc < 2.33.

abalabin-bamfunds:
I guess there is no appetite for migrating to `manylinux_2_34`? I must admit I do not know what repercussions of that would be and quick googling suggests there are.

I'll have a look at loading mallinfo2 dynamically.

texodus:
Adding support for `manylinux_2_34` isn't a problem, but removing support for `manylinux_2_28` means old linux systems cannot upgrade Perspective.

`manylinux_2_28` can be deprecated iff @timkpaine says it is ok.

timkpaine:
nope

texodus (approved):
Looks good! [full build](https://github.com/perspective-dev/perspective/actions/runs/29556618769)

PR #3188, opened 2026-06-19
Fix memory safety issues

texodus:
I asked Claude to find memory safety bugs in Perspective and write a PR, while I played Balatro on my phone. Here's its own summary of what it found:

## Fixes

- **Uneven column lengths in columnar table create/update** (`table.cpp` —
  `Table::from_cols`, `Table::update_cols`): the table was sized from a single
  column's length while every column was filled to its own length. Now the
  table is sized to the longest column and shorter columns are null-padded, so
  all writes stay in bounds.

- **NDJSON row under-allocation** (`table.cpp` — `Table::from_ndjson`): capacity
  was estimated from the newline count but one row was written per parsed
  object. The table is now grown per row (amortized O(1) via geometric capacity
  growth), so concatenated objects without newline separators can no longer
  overrun the buffer.

- **Arrow row-count truncation** (`arrow_loader.cpp` — `ArrowLoader::row_count`):
  Arrow's 64-bit row count was silently truncated to 32 bits, under-sizing the
  destination table relative to the data written during fill. Oversized/negative
  counts are now rejected instead of truncated.

- **Arrow `time32` element width** (`arrow_loader.cpp` — `copy_array`): `time32`
  values are 32-bit and map to a 4-byte column, but the loader copied 8 bytes
  per element, over-reading the source and over-writing the destination. Now
  copies 4 bytes per element.

- **`first`/`last` aggregate with a missing sort dependency** (`sparse_tree.cpp`
  — `first_last_helper`): the helper assumed the aggregate spec always carried
  both a value and a sort dependency and indexed the dependency list
  unconditionally. A view whose sort column falls outside the visible set could
  produce a spec without the sort dependency, causing an out-of-bounds read. Now
  guarded (covers `first`, `last`, and `last − first`).

- **`count` aggregate over a string column reads at the wrong stride**
  (`aggregate.cpp` — `t_aggregate::init`, `AGGTYPE_COUNT` / `DTYPE_STR`): a
  string column's backing store holds 4-byte `t_uindex` vocabulary indices, but
  the `count` aggregate was instantiated as `t_aggimpl_count<std::uint64_t, …>`,
  so `build_aggregate` → `t_column::fill` read the input column at an 8-byte
  (`uint64`) stride — double the real element width — and ran off the end of the
  buffer. The read is pure waste (`t_aggimpl_count::reduce` discards the values;
  the count is filled in later from `nstrands`), but it still tripped a storage
  bounds check. The raw type is now `t_uindex`, matching the storage width. This
  was a latent over-read for years, masked by `push_back`'s ~20% capacity
  headroom; tighter column sizing in a later optimization removed the slack and
  exposed it. It is reachable from every pivoted (`group_by`/`split_by`) view
  with a string column, since the default per-column aggregate for a string is
  `count` and that aggregate is built during view construction — so the abort
  fired before the view's actual query (`get_min_max`, `to_columns`, etc.) ever
  ran. Found by updating and enabling `PSP_STORAGE_VERIFY` in debug builds.

- **Residency eviction data race** (`residency.cpp`, `residency.h`): the
  shared pending-eviction vectors were cleared outside the manager's mutex, so
  concurrent request-thread eviction passes could double-free. All mutations now
  occur under the lock, and a dedicated mutex serializes each eviction cycle.

- **Unvalidated Arrow input** (`arrow_loader.cpp` — `ArrowLoader::initialize`):
  Arrow's IPC reader does not validate buffer contents, yet the fill paths index
  value/offset/dictionary buffers directly. The loaded (remotely supplied) table
  is now fully validated (`ValidateFull()`) before those buffers are trusted, so
  a malformed payload — bad offsets, out-of-range dictionary indices, inconsistent
  chunk lengths — is rejected instead of read out of bounds. This is the systemic
  defense behind the `time32` and row-count fixes above. Note: validation is
  O(data) per ingested payload; `Validate()` plus targeted bounds checks would be
  a cheaper alternative if ingest throughput matters.

- **Out-of-bounds access in expression vector functions** (`computed_function.cpp`
  — `diff3`, `norm3`, `cross_product3`, `dot_product3`): these operate on
  3-element vectors and indexed `v[0..2]` unconditionally, but their exprtk
  parameter sequences (`"VVV"`/`"VV"`/`"V"`) enforce vector *type*, not *length*.
  A user expression can declare shorter vectors (e.g. `var v[2] := {1,2};
  norm3(v)`), causing an out-of-bounds read — and for `diff3`/`cross_product3`,
  an out-of-bounds *write* to the (short) output vector. Each function now
  clears its result for vectors shorter than 3 before indexing; this surfaces
  as an invalid expression (rejected at view creation) rather than an
  out-of-bounds access. Vectors of length 3 are unaffected.


rust/perspective-js/test/js/constructors/arrow.spec.ts
```diff
@@ -162,5 +162,44 @@ test.describe("Arrow", function () {
...
+                "ffffffff78000000140000000000000000000a000e000c000b0004000a000000" +
+                "140000000000000304000a0018000c00080004000a0000002c00000010000000" +
+                "0100000000000000000000000100000001000000000000000100000000000000" +
+                "0000000000000000000000000000000000000000000000000000000000000000" +
+                "ffffffff00000000";
+            const bytes = new Uint8Array(
+                hex.match(/.{2}/g)!.map((b) => parseInt(b, 16)),
+            );
+            await expect(perspective.table(bytes)).rejects.toThrow(
+                /row count exceeds maximum supported size/,
```
texodus, line 187: This is brittle and irrelevent, we don't cate about the message.

rust/perspective-js/test/js/constructors/arrow.spec.ts
```diff
@@ -162,5 +162,44 @@ test.describe("Arrow", function () {
...
+                "140000000000000304000a0018000c00080004000a0000002c00000010000000" +
+                "ffffffff000000000000000001000000ffffffff00000000ffffffff00000000" +
+                "0000000000000000000000000000000000000000000000000000000000000000" +
+                "ffffffff78000000140000000000000000000a000e000c000b0004000a000000" +
+                "140000000000000304000a0018000c00080004000a0000002c00000010000000" +
+                "0100000000000000000000000100000001000000000000000100000000000000" +
+                "0000000000000000000000000000000000000000000000000000000000000000" +
+                "ffffffff00000000";
+            const bytes = new Uint8Array(
+                hex.match(/.{2}/g)!.map((b) => parseInt(b, 16)),
```
texodus: Overly sensitive

rust/perspective-js/test/js/constructors/arrow.spec.ts
```diff
@@ -162,5 +162,44 @@ test.describe("Arrow", function () {
...
+                "0000000004000400040000000100000078000000000000000000000000000000" +
+                "ffffffff78000000140000000000000000000a000e000c000b0004000a000000" +
+                "140000000000000304000a0018000c00080004000a0000002c00000010000000" +
+                "ffffffff000000000000000001000000ffffffff00000000ffffffff00000000" +
+                "0000000000000000000000000000000000000000000000000000000000000000" +
+                "ffffffff78000000140000000000000000000a000e000c000b0004000a000000" +
+                "140000000000000304000a0018000c00080004000a0000002c00000010000000" +
+                "0100000000000000000000000100000001000000000000000100000000000000" +
+                "0000000000000000000000000000000000000000000000000000000000000000" +
+                "ffffffff00000000";
```
texodus: This is going to be a PITA to update or understand or debug, also the other tests already store raw arrows for test in the repo.

rust/perspective-js/test/js/aggregates.spec.js
```diff
@@ -164,6 +164,52 @@ const std = (nums) => {
...
+                aggregates: { v: "first" },
+                sort: [["v", "desc"]],
+            });
+
+            const result = await view.to_json();
+            const paths = result.map((r) => r.__ROW_PATH__);
+            expect(paths.length).toEqual(3);
+            expect(paths).toContainEqual([]);
+            expect(paths).toContainEqual(["x"]);
+            expect(paths).toContainEqual(["y"]);
```
texodus, line 185: These data structures are small, for readability its worth just asserting the entire structure instead of multiple playwright assertion calls in a row.

rust/perspective-server/cpp/perspective/src/cpp/arrow_loader.cpp
```diff
@@ -211,6 +212,11 @@ ArrowLoader::initialize(const std::uint8_t* ptr, const uint32_t length) {
         load_stream(ptr, length, m_table);
     }
 
+    const auto validation = m_table->ValidateFull();
+    if (!validation.ok()) {
+        PSP_COMPLAIN_AND_ABORT(validation.ToString());
+    }
```
texodus, line 218: I checked that this has a minimal (nonexistent) performance impact.

rust/perspective-server/cpp/perspective/src/cpp/arrow_loader.cpp
```diff
@@ -972,7 +978,16 @@ ArrowLoader::fill_column(
 
 std::uint32_t
 ArrowLoader::row_count() const {
-    return m_table->num_rows();
+    const std::int64_t n = m_table->num_rows();
+    if (n < 0
+        || static_cast<std::uint64_t>(n)
+            > std::numeric_limits<std::uint32_t>::max()) {
+        PSP_COMPLAIN_AND_ABORT(
```
texodus, line 985: This is not something Perspective itself will check internally, e.g. a subsequent `update` call will still cause the same issue this verbose error messages "fixes".

rust/perspective-server/cpp/perspective/src/cpp/computed_function.cpp
```diff
@@ -1885,6 +1885,11 @@ diff3::operator()(t_parameter_list parameters) {
     t_vector_view v2(parameters[1]);
     t_vector_view out(parameters[2]);
 
+    if (v1.size() < 3 || v2.size() < 3 || out.size() < 3) {
+        rval.m_status = STATUS_CLEAR;
+        return rval;
+    }
```
texodus, line 1891: This is an actual good catch but we can validate this better by just checking the exact cases and erroring else?

rust/perspective-server/cpp/perspective/src/cpp/sparse_tree.cpp
```diff
@@ -1238,6 +1238,11 @@ t_stree::update_agg_table(
             case AGGTYPE_WEIGHTED_MEAN: {
                 auto pkeys = get_pkeys(nidx);
 
+                if (spec.get_dependencies().size() < 2) {
```
texodus, line 1241: I don't think empty categories can be constructed with the current config but this fix satisfies the analysis.

PR #3193, opened 2026-07-09
fix(client): preserve underscores in SQL column aliases

### LINKED ISSUE #3187: GenericSQLVirtualServerModel mangles column names with underscores
### Steps to Reproduce:

`package.json`:
```json
{
  "name": "perspective-underscore-alias-repro",
  "version": "1.0.0",
  "private": true,
  "type": "module",
  "description": "repro",
  "scripts": {
    "repro": "node repro.mjs"
  },
  "dependencies": {
    "@perspective-dev/client": "4.5.1"
  }
}
```

`repro.mjs`:
```javascript
import { GenericSQLVirtualServerModel } from '@perspective-dev/client/node';

const model = new GenericSQLVirtualServerModel();

const sql = model.tableMakeView('memory.t', 'v', {
  sort: [],
  group_by: [],
  split_by: [],
  columns: ['account_number'],
  filter: [],
  expressions: {},
  aggregates: {},
});

console.log(sql);
// CREATE TABLE v AS (SELECT "account_number" as "account-number" FROM memory.t ORDER BY rowid)
```

To run:
```bash
npm install
npm run repo
```

### Expected Result:

Column names with underscores (eg `account_name`) are passed through unchanged.

### Actual Result:

Underscores are replaced with hyphens, so `account_name` becomes `account-name`.

In addition to not matching the underlying data, this leads to issues using the datatable, because the table "data" uses the mangled names, but the column picker uses the correct (ie underscore) names.  Additionally, sorting breaks with error messages like "Uncaught (in promise) Error: Unknown column "account-name" in field `sort`" because the names don't match the actual columns:

Image

LarryHu0217:
## Description

Preserves underscores in column aliases generated by `GenericSQLVirtualServerModel`. The SQL builder previously replaced `_` with `-`, so selecting `account_number` produced an alias named `account-number` and changed the schema exposed to clients.

The fix keeps the original identifier while continuing to escape embedded double quotes. Regression tests cover both flat and aggregated view queries.

Fixes #3187.

## Validation

- `cargo test --features generate-proto,omit_metadata generic_sql_model --no-default-features` (25 passing)
- `rustfmt --check` on the changed Rust files
- `git diff --check origin/master...HEAD`

## Tooling assistance

Codex assisted with implementation and local validation. I reviewed the resulting diff and test output.

### DIFF
--- a/rust/perspective-client/src/rust/virtual_server/generic_sql_model/table_make_view.rs
+++ b/rust/perspective-client/src/rust/virtual_server/generic_sql_model/table_make_view.rs
@@ -380,7 +380,7 @@ impl<'a> ViewQueryContext<'a> {
         if self.needs_aggregation() {
             for col in self.config.columns.iter().flatten() {
                 let agg = self.get_aggregate(col);
-                let escaped = col.replace('"', "\"\"").replace("_", "-");
+                let escaped = col.replace('"', "\"\"");
                 clauses.push(format!(
                     "{}({}) as \"{}\"",
                     agg,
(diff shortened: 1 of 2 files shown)


texodus:
I've update #3187 to clarify this is an explicit design decision, not a bug.

PR #3194, opened 2026-07-09
fix(server): serialize dates at UTC midnight

### LINKED ISSUE #3106: Date column renders one day early in server mode due to local‑midnight serialization
This bug report has been written with the help of GTP-5.2

### Summary
When using `perspective-python` in **server mode** (Python server + `perspective-viewer`), `DATE` values can render **one day earlier** in the datagrid for non-UTC timezones (e.g. `Europe/Amsterdam`). The value returned by `to_columns()` appears to represent **local midnight**, but the datagrid formats it as **UTC**, shifting the calendar day backward.

This does **not** reproduce in client mode (pure JS). CSV export appears correct, likely because Arrow preserves date semantics.

### Steps to Reproduce:

```
Using the uv package manager:
TZ=Europe/Amsterdam uv run --with perspective-python python - <<'PY'
from datetime import datetime, timezone
import perspective as psp

client = psp.Server().new_local_client()
Table = client.table

tbl = Table({"d": "date"})
tbl.update({"d": ["2024-01-01"]})

value = tbl.view().to_columns()["d"][0]
utc_date = datetime.fromtimestamp(value / 1000, tz=timezone.utc).date()
local_date = datetime.fromtimestamp(value / 1000).date()


print("value_ms", value)
print("utc_date", utc_date)
print("local_date", local_date)
PY
Installed 1 package in 16ms
value_ms 1704063600000
utc_date 2023-12-31
local_date 2024-01-01
```

### Expected Result:

DATE values should be timezone‑agnostic and display the same day regardless of local TZ.
In the datagrid, 2024-01-01 should render as 2024-01-01, not 2023-12-31.

### Actual Result:

In server mode, DATE values appear one day early in the datagrid for non‑UTC TZs.
The raw to_columns() value corresponds to local‑midnight, which the datagrid formats in UTC, shifting the day backward.

### Environment

- Ubuntu 22.04.5 LTS on Windows 10 x86_
- python version == 3.12.8. perspective-server version == 4.0.1

LarryHu0217:
## Description

Serializes Perspective `date` values as UTC calendar days instead of converting local midnight with `mktime()`. In positive-offset time zones, the previous conversion shifted a date such as `2024-01-01` to the previous UTC day. Formatted date output now uses the stored calendar value directly.

The regression test fixes the process timezone to `Europe/Amsterdam` and verifies both the numeric UTC-midnight value and CSV output.

Fixes #3106.

## Validation

- Changed native C++ translation units compiled successfully
- `test_to_format.py` (91 passed, 17 skipped)
- Ruff check and format check for the changed Python test
- Post-fix runtime verification for JSON and CSV output
- `git diff --check origin/master...HEAD`

## Tooling assistance

Codex assisted with implementation and local validation. I reviewed the resulting diff and test output.

### DIFF
--- a/rust/perspective-python/perspective/tests/table/test_to_format.py
+++ b/rust/perspective-python/perspective/tests/table/test_to_format.py
@@ -11,7 +11,8 @@
 #  ┗━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┛
 
 import os
-from datetime import date, datetime
+import time
+from datetime import date, datetime, timezone
 from io import StringIO
 
 import numpy as np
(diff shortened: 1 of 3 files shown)


texodus:
This change introduces test suite failures due to (now) non-symmetric encoding (`test_table_datetime.py`). Closing in favor of #3197

PR #3214, opened 2026-08-16
fix(datagrid): restore persisted column widths

### LINKED ISSUE #3163: Column widths are not retained and not persisted
### Steps to Reproduce:

Open the superstore example. Resize the first Sales column to be wide (or narrow). Flip to the 'untitled' tab, flip back. The Sales column is back to its auto-fit size.

After columns are manually resized, open debug JSON panel one can see that column sizes are not written into plugin_config/columns.

The only test I found looking at column width persistence is skipped in datagrid.spec.ts. I think maybe refactoring in 4.1 introduced a regression?

### Expected Result:

I expect when user resizes a column manually the size to be retained and persisted.

### Actual Result:

Columns revert to auto-fit sizes.

### Environment:
 
- Win 11
- perspective 4.4
- Edge Browser to https://perspective-dev.github.io/blocks/superstore/index.html

LarryHu0217:
## Description

On the first Datagrid restore, saved column widths could be discarded in two places: the viewer schema filtered out the plugin-owned `columns` key, and the width restore could run before the plugin and `regular-table` had finished their initial draw.

This change:

- declares `columns` as passthrough plugin config so it survives schema filtering
- caches width overrides received before Datagrid initialization
- reapplies saved widths after the initial `regular-table` draw completes
- adds a regression test that verifies both the rendered width override and the subsequently saved config

Fixes #3163.

## Validation

- `PACKAGE=viewer-datagrid pnpm run lint`
- `pnpm --filter @perspective-dev/metadata run build`
- `pnpm --filter @perspective-dev/server run build`
- `pnpm --filter @perspective-dev/client run build`
- `pnpm --filter @perspective-dev/viewer run build`
- `pnpm --filter @perspective-dev/viewer-datagrid run build`
- `PACKAGE=viewer-datagrid pnpm exec playwright test --config=tools/test/playwright.config.ts packages/viewer-datagrid/test/js/column_settings.spec.ts --grep=First` (1 passed)
- `git diff --check`

Documentation is not applicable because this fixes existing persistence behavior and does not add a public API.

## AI assistance

I used OpenAI Codex to inspect the codebase, assist with the implementation and regression test, and run and review the validation steps. I reviewed the final diff and test results myself.

### DIFF
--- a/packages/viewer-datagrid/src/ts/plugin/column_config_schema.ts
+++ b/packages/viewer-datagrid/src/ts/plugin/column_config_schema.ts
@@ -22,6 +22,7 @@ type ControlSpec = Record<string, unknown> & { kind: string };
 
 export interface ColumnConfigSchema {
     fields: ControlSpec[];
+    passthrough_keys?: string[];
 }
 
 /**
--- a/packages/viewer-datagrid/src/ts/custom_elements/datagrid.ts
+++ b/packages/viewer-datagrid/src/ts/custom_elements/datagrid.ts
@@ -165,6 +165,7 @@ export class HTMLPerspectiveViewerDatagridPluginElement
 
         return {
             fields,
+            passthrough_keys: ["columns"],
         };
     }
 
--- a/packages/viewer-datagrid/src/ts/model/column_overrides.ts
+++ b/packages/viewer-datagrid/src/ts/model/column_overrides.ts
@@ -34,12 +34,12 @@ export function restore_column_size_overrides(
     old_sizes: ColumnOverrides,
     cache = false,
 ): void {
-    if (!this._initialized) {
-        return;
-    }
-
     if (cache) {
         this._cached_column_sizes = old_sizes;
+    }
+
+    if (!this._initialized) {
+        return;
     }
 
     const regular_table = this.regular_table as RegularTableWithOverrides;
--- a/packages/viewer-datagrid/src/ts/plugin/draw.ts
+++ b/packages/viewer-datagrid/src/ts/plugin/draw.ts
@@ -62,6 +62,9 @@ export async function draw(
 
     restore_column_size_overrides.call(this, old_sizes);
     await drawPromise;
+    if (Object.keys(old_sizes).length > 0) {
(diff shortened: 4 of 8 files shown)


packages/viewer-datagrid/src/ts/plugin/column_config_schema.ts
```diff
@@ -22,6 +22,7 @@ type ControlSpec = Record<string, unknown> & { kind: string };
 
 export interface ColumnConfigSchema {
     fields: ControlSpec[];
+    passthrough_keys?: string[];
```
texodus, line 25: Column widths are not special - they are a numeric property of the column's style. Column widths need to be defined/advertised in the datagrid's `column_config_schema`, and set in the `ViewerConfig::columns_config` field, not `plugin_config` (as was the case in previous versions that supported persistent column widths).

texodus (changes requested):
Thanks for the PR, but unfortunately this isn't how I want this fixed (not your fault).

First - the reported issue pre-dates `column_config_schema`, which makes column-specific non-query config fields dynamic and plugin-describable. See [`datagrid.ts`](https://github.com/perspective-dev/perspective/blob/master/packages/viewer-datagrid/src/ts/custom_elements/datagrid.ts#L171) implementation of `columns_config_schema`, or the similar version in the charts package which has elaborate column-level config [`plugin.ts`](https://github.com/perspective-dev/perspective/blob/master/packages/viewer-charts/src/ts/plugin/plugin.ts#L456).

Second and unfortunately blocking ATM is that this implementation (nor the original, which is why it was removed) work in the presence of `split_by`, and its not clear to me how to handle this without adding a new abstraction level to configuration down to the column group/row group level, which pollutes the config with data and potentially makes it much larger.

I would take a PR that implements a fix for the first issue (use correct `ColumnSpec` and datagrid publication), **and** disables column width persistence when `split_by` is populated (or it will look weird, or be silently ignored, when restored).

texodus:
I meant to close this issue, not request changes - if you want to submit a fix please open a new PR (without this code in the history).

